PR #28, opened 2015-10-28
Added obm setting check when posting or patching in nodes api

iceiilin:
Only enclosure node supports Panduit OBM settings, so adding check of invalid obm settings when user post or patch this info.
In this implementation, compute node only supports ipmi obm settings, and enclosure node only supports Panduit.

https://hwjiraprd01.corp.emc.com/browse/MAG-222

Related PRs:
https://github.com/RackHD/on-tasks/pull/55
https://github.com/RackHD/on-core/pull/22

### DIFF
--- a/lib/api/nodes.js
+++ b/lib/api/nodes.js
@@ -164,14 +164,31 @@ function nodesRouterFactory (
      *     HTTP/1.1 404 Not Found
      *     {
      *       "error": "Not Found"
-     *      }
+     *     }
+     *     HTTP/1.1 400 Bad Request
+     *     {
+     *       "message": "Service xxx is not supported in node type xxx"
+     *     }
      */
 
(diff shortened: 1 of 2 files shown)


iceiilin:
@benbp

benbp:
Yep, good points. @iceiilin. :+1: 

Also, note for the future, perhaps if we touch this code later on it would be a good idea to refactor some of the logic into methods within the new nodes-api-service file that @anhou has just created as part of https://github.com/RackHD/on-http/pull/34 ?

benbp:
Just waiting on 2nd +1s for the on-core/on-tasks PRs, then we can merge.

anhou:
:+1: 
@iceiilin @benbp note here, I'll create a story to refactor the methods codes in lib/api/nodes.js to lib/services/nodes-api-service.js

PR #31, opened 2015-10-30
Fix ODR-260 and provide solution for RHEL/CentOS 7 network setting failure issue

pengz1:
To fix ODR-260 and provide solution for RHEL/CentOS 7 network setting failure issue in some cases (apply to ODR-319). 
* ODR-260: RHEL/CentOS bootstrap failed to set network on D51 sometimes
* Fix issue that T41 network setting won't take effective until we reboot OS manually after RHEL7/CentOS7 bootstrapping (part of ODR-319).

I tested this change with configuration of (RHEL7/CentOS7+T41), (RHEL7+D51). All network settings are good and it solved RHEL7/CentOS7 network related issues we have seen before.

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -9,7 +9,7 @@ keyboard 'us'
 timezone America/Los_Angeles --isUtc
 firewall --enabled --http --ssh
 selinux --permissive
-bootloader --location=mbr --driveorder=sda --append="crashkernel=auth rhgb"
+bootloader --location=mbr --driveorder=sda --append="crashkernel=auth rhgb net.ifnames=0 biosdevname=0"
 services --enabled=NetworkManager,sshd
 network --device=<%=macaddress%> --noipv6 --activate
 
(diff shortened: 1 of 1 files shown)


stuart-stanley:
well crap. -1 -1 -1 -1 -1 -1 -1
The issue that spawn this wasn't fixed by it in the real deployment. In fact, during deployment testing, the .ifname=0 (going back to trusting the old-style kernel names) showed inconsistent naming during installs: a few times eth0 got mapped to the first mez card, most other times it got mapped to the 2nd mez.

We need to revert this, since it will also mess up regression tests because of the name changes again.
@benbp can you do that directly, or does another PR need to be created with the change backed out?

benbp:
@stuart-stanley a new PR has to be created, but there should be a revert button right above your comment that you can click to auto-generate that PR.

stuart-stanley:
@benbp no such luck. (Though I'm now a collaborator instead of just a sympathizer now ;)

PR #32, opened 2015-10-30
Fix ODR-260 and provide solution for RHEL/CentOS 7 network setting failure issue

pengz1:
To fix ODR-260 and provide solution for RHEL/CentOS 7 network setting failure issue in some cases (apply to ODR-319). 
* ODR-260: RHEL/CentOS bootstrap failed to set network on D51 sometimes
* Fix issue that T41 network setting won't take effective until we reboot OS manually after RHEL7/CentOS7 bootstrapping (part of ODR-319).

I tested this change with configuration of (RHEL7/CentOS7+T41), (RHEL7+D51). All network settings are good and it solved RHEL7/CentOS7 network related issues we have seen before.


benbp:
:+1:

stuart-stanley:
-1
The issue that spawned this wasn't fixed by it in the real deployment. In fact, during deployment testing, the .ifname=0 (going back to trusting the old-style kernel names) showed inconsistent naming during installs: a few times eth0 got mapped to the first mez card, most other times it got mapped to the 2nd mez.

benbp:
Is someone following up on this?

stuart-stanley:
Yes. The immediate solution was to use the GA release of redhat rather than the beta. Longer term we will need to refactor the kickstart to either use the network services OR NetworkManager, but not both the way we are doing it now. This pull can be closed, however.

heckj:
@pengz1 verifying - you good with closing this? Based on stuart's response, I was tempted to just close, but since you put up the PR, I wanted to check with you.

pengz1:
Close this PR.

PR #34, opened 2015-11-03
remove  node update based on enclosure node update

anhou:
Update removing node based on @iceiilin 's PR https://github.com/RackHD/on-tasks/pull/26 
1. when removing a compute node, check the relationship within this node, if it's enclosed by an enclosure node, after remove compute node, also remove the relationship within enclosure node, when all nodes   of an enclosure are removed, remove the enclosure automatically.
2. add nodes-api-service.js to handle nodes api function.
3. add unit tests

### DIFF
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -1,0 +1,132 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = nodeApiServiceFactory;
+di.annotate(nodeApiServiceFactory, new di.Provide('Http.Services.Api.Nodes'));
+di.annotate(nodeApiServiceFactory,
+    new di.Inject(
+        'Protocol.TaskGraphRunner',
+        'Services.Waterline',
+        'Errors',
+        'Logger',
+        '_',
+        'Promise'
+    )
+);
+function nodeApiServiceFactory(
+    taskGraphProtocol,
+    waterline,
+    Errors,
+    Logger,
+    _,
+    Promise
+) {
+    var logger = Logger.initialize(nodeApiServiceFactory);
+
+    function NodeApiService() {
+    }
+
+    /**
+     * Find Enclosure nodes which enclose compute node
+     * @param  {Object}     node
+     * @return {Promise}
+     */
+    NodeApiService.prototype._findEnclNodes = function(node) {
+        // Find the enclosure nodes who enclose this compute node
+        if (!_.has(node, 'relations')) {
+            return Promise.resolve();
+        }
+
+        var relation = _.find(node.relations, { relationType: 'enclosedBy' });
+        if (!relation || !_.has(relation, 'targets') ) {
+            return Promise.resolve();
+        }
+
+        return Promise.map(relation.targets, function (enclNodeId) {
+            return waterline.nodes.needByIdentifier(enclNodeId)
+            .catch(function (err) {
+                logger.error("Error Getting Enclosure Node", { error: err });
+                return;
+            });
+        });
+    };
+
+    /**
+     * Remove the relations between node and enclosure node, if enclosure node
+     * doesn't enclose any nodes, this enclosure node is also removed.
+     * @param  {Object}     node
+     * @param  {Array}      enclNodes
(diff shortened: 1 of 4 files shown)


tldavies:
We are testing Travis CI. Since there are no .travis.yml files it will fail.

lib/services/nodes-api-service.js
```diff
@@ -0,0 +1,96 @@
...
+                        .catch(function (err) {
+                            logger.error("Error Getting Enclosure Node", { error: err });
+                            // If don't find enclosure node, continue to remove this node
+                            return;
+                        });
+                    });
+                }
+            }
+        })
+        .then(function (enclNodes) {
```
benbp: I think it would be better to make the code in these promise blocks their own named functions, that way we can more easily test their functionality individually, and also have a more easily readable promise chain.

lib/services/nodes-api-service.js
```diff
@@ -0,0 +1,96 @@
...
+    taskGraphProtocol,
+    waterline,
+    Errors,
+    Logger,
+    _,
+    Promise
+) {
+    var logger = Logger.initialize(nodeApiServiceFactory);
+
+    function NodeApiService() {
```
benbp: Good, glad we have a node api service. Just as a note it might make sense going forward for us to start moving the rest of logic in the node routes into here (switch/pdu discovery, etc.).

tldavies:
test this please

benbp:
:+1:

with a suggestion about some potential cleanup.

lib/services/nodes-api-service.js
yyscamper: When I work on the bootstrap, I feel it is better if we can move the profile-api-service and template render service to the on-core, then some of the function (like rendering) can be shared in the job implementation (on-tasks).
Same to the nodes-api-service, in future we may have some internal logic to remove the nodes of a enclosure.

Of course, we can keep it here right now and move to on-core if we come across that situation.

anhou: @yyscamper  I don't think it's proper be put into on-core, some of the service would run specific taskgraph, it's not good that it appears in on-core.

anhou:
codes has been refactored and updated

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -0,0 +1,251 @@
...
+        waterline.workitems = {
+            destroy: function() {}
+        };
+        waterline.lookups = {
+            update: function() {}
+        };
+    });
+
+    beforeEach("Http.Services.Api.Nodes beforeEach", function() {
+        this.sandbox = sinon.sandbox.create();
```
benbp: This was an error that I put in a bunch of tests a while back. This statement can actually go in the before block instead of the beforeEach block, since we only need to create the sandbox once.

benbp:
Thanks @anhou! Still :+1:

spec/lib/services/nodes-api-service-spec.js
anhou: Great.  most of cases didn't pay attention on this. let's correct it from now. I've also update it

lib/services/nodes-api-service.js
```diff
@@ -0,0 +1,132 @@
...
+    };
+
+    /**
+     * Remove the relations between node and enclosure node, if enclosure node
+     * doesn't enclose any nodes, this enclosure node is also removed.
+     * @param  {Object}     node
+     * @param  {Array}      enclNodes
+     * @return {Promise}
+     */
+    NodeApiService.prototype._removeRelations = function(node, enclNodes) {
```
yyscamper: not big problem. the better name is "removeEnclosureRelations" because it cannot be used to remove other future relations.

lib/services/nodes-api-service.js
yyscamper: @anhou, the "Protocol.TaskGraphRunner" is just in on-core, so I think it's proper. Image future we implements the auto node removal, I think this node-api-service is exactly what we need for auto node removal.

Anyway, it's OK for me that move it to on-core after we come across that situation.

yyscamper:
still :+1: , though I have some minor suggestion

PR #38, opened 2015-11-05
merge ESXi bootstrap files and use dynamic ESXi modules files

yyscamper:
This PR is to keep the bootstrap feature sync with the 1.0.0 branch.
(1) We will use dynamic module file name in ESXi bootstrap script, these file name is dynamic fetched from external repository, this can ensure ESXi bootstrap (5.5 & 6.0) can be merged.

(2) RHEL kickstart file will not be used any more because now it shares the same one with CentOS, so I remove it.

(3) Rename esx-boot-cfg-hybrid to esx-boot-cfg, the "hybrid" one is to use both http and tftp protocol to transfer ESXi modules but since the external repository feature is completed, it cannot use hybrid protocol any more.

@RackHD/corecommitters @iceiilin @pengz1 @zyoung51 @WangWinson 

Following PRs need be merged together to ensure the bootstrap function works well for master branch:
1. https://github.com/RackHD/on-http/pull/38
2. https://github.com/RackHD/on-taskgraph/pull/22
3. https://github.com/RackHD/on-tasks/pull/62
4. https://github.com/RackHD/onserve/pull/64


yyscamper:
Some PRs across different repo needs be merged together to ensure the bootstrap feature works well, I will provide these information after all PRs are created.

benbp:
:+1:

data/templates/esx-boot-cfg
```diff
@@ -0,0 +1,7 @@
+bootstate=0
+title=Loading ESXi installer
+kernel=<%=tbootFile%>
+kernelopt=runweasel formatwithmbr com1_baud=115200 com1_Port=<%=comportaddress%> tty2Port=<%=comport%> debugLogToSerial=1 logPort=<%=comport%> ks=http://<%=server%>:<%=port%>/api/common/templates/esx-ks
+modules=<%=moduleFiles%>
+build=
```
zyoung51: Is prefix=<%=repo%> required to get the right url when using the moduleFiles/tbootFile parsed out of the repo's boot.cfg?

yyscamper: The moduleFiles/tbootFile has already been added with repo before render (see the PR of on-core, os-repo-tool.js, function parseEsxBootCfgFile). Otherwise, I need double render and rewrite this template file under http data folder. Render function is currently not in on-core, on-tasks changes on-http file is not good architect design, and change on-http template needs restart. To solve all these problem need much work. so I choose the current solution, the simplest one.

yyscamper: This is line that will add the repo address after get the module names from ESXi boot.cfg.
https://github.com/RackHD/on-tasks/pull/62/files#diff-4773d4a746d926b1b592336d255b54ecR72

zyoung51: Thanks for the details Felix.  The code can be removed and replaced with a "prefix=<%=repo%>" in the boot.cfg file.  Using the prefix would let you skip the step where you are forcibly changing the user's repo input to lowercase in the code above.  Changing the case would affect URLs with auth tokens encoded in them.  I'll make a note of that in the other PR.

iceiilin:
+1

data/templates/esx-boot-cfg
yyscamper: Thanks @zyoung51 , your suggestion simplifies our design. I want to make another task to implement your idea but not in this PR, since it involves on-tasks change and need some testing.

zyoung51:
:+1:

PR #39, opened 2015-11-09
Remove sudo and package dependencies from travis file (so we can have…

benbp:
… travis containers)

Our unit tests here shouldn't require sudo or apt updates, so remove them in order to support new travis containerized infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,9 @@
-sudo: required
 language: node_js
 node_js:
   - "0.10"
 services:
  - mongodb
  - rabbitmq
-before_install:
- - sudo apt-get update -qq
- - sudo apt-get install -qq snmp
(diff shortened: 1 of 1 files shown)


heckj:
+1

benbp:
@tldavies @VulpesArtificem

tldavies:
+1

rolandpoulter:
@benbp We may need to put this back for node v4 support. See https://github.com/RackHD/on-core/blob/master/.travis.yml#L19

benbp:
Ah ok, I'll see if I can just disable gs-stats in these unit tests preferably.

rolandpoulter:
Need to change on-core to not install `gs-stats`, otherwise builds will break during install step.

PR #41, opened 2015-11-11
esx-ks reboot delay fix for 5.5 support

king-jam:
Adjusted this value since it wasn't valid and causes the first boot config not to take place unless another reboot is manually applied.

Here is an error if you manually try to run the old command on 5.5

    esxcli system shutdown reboot -d 5 -r "Rebooting after first boot host configuration"
    Invalid data constraint for parameter 'delay'. Expected integer in the range 10-4294967295


zyoung51:
:+1:

benbp:
:+1:

yyscamper:
:+1:

PR #42, opened 2015-11-11
VEY-219: Automate OS installation media setup

zyoung51:
Automate the process of deploying OS media (ISO) from a local/remote(http) to a local mirror and prepare the local mirror to be used as a target for network installation.

@tannoa2 @RackHD/corecommitters

### DIFF
--- a/data/templates/setup_iso.py
+++ b/data/templates/setup_iso.py
@@ -1,0 +1,154 @@
+# setup_iso.py [http://path.to/file.iso | /path/to/file.iso] [/var/destination]
+#
+# Deploy the ISO file contents to the specified destination directory such that
+# destination directory can be used as a target for RackHD OS bootstrap workflows
+import argparse
+import subprocess
+import os
+import errno
+import os.path
+import tempfile
+import atexit
+import sys
+import shutil
+import re
+import urllib
+
+tmpdir = ''
+verbose=1
+
+# Cleanup
+@atexit.register
+def cleanup_tmp():
+    if tmpdir != '':
+        subprocess.check_call(['umount', tmpdir])
+        os.rmdir(tmpdir)
+
+def get_iso_info(fname):
+    # Access ISO to determine OS type from supported list
+    label1=subprocess.Popen(['isoinfo', '-i', fname, '-d'], stdout=subprocess.PIPE)
+    return label1.stdout.read()
+
+def do_setup_repo(osname, osver, src, dest):
+    print 'Installing {0} {1} to {2}/{0}/{1}'.format(osname, osver, dest)
+    dstpath=dest + '/' + osname + '/' + osver
+    if os.path.isdir(dstpath):
+        print 'Found existing directory, bailing...'
+        sys.exit(1)
+
+    if osname is 'ESXi':
+        shutil.copytree(src, dstpath)
+
+    if osname is 'RHEL' or osname is 'Centos':
+        shutil.copytree(src, dstpath)
+
+    if osname is 'LiveCD':
+        initrd='initrd.img'
+        vmlinuz='vmlinuz'
+        if os.path.isfile(src+'/isolinux/initrd0.img'):
+            initrd='initrd0.img'
+            vmlinuz='vmlinuz0'
+        mount1 = subprocess.Popen(['mount'], stdout=subprocess.PIPE)
+        mount2 = subprocess.Popen(['grep', src], stdin=mount1.stdout, stdout=subprocess.PIPE)
+        mount3 = subprocess.Popen(['awk', '{print $1}'], stdin=mount2.stdout, stdout=subprocess.PIPE)
+        isoname = mount3.communicate()[0]
+
+        iso_basename = os.path.basename(isoname).strip()
+        iso_dirname = os.path.dirname(isoname).strip()
+
+        os.makedirs(dstpath)
+        
+        syscall = '(cd "'+iso_dirname+'" && echo "'+iso_basename+'" | cpio -H newc --quiet -L -o )'
+        syscall += ' | gzip -9'
+        syscall += ' | cat '+src+'/isolinux/'+initrd+' - > '+dstpath+'/initrd.img'
+        os.system(syscall)
+
+        shutil.copyfile(src+'/isolinux/'+vmlinuz, dstpath+'/'+vmlinuz)
+
+    os.system('ln -sf ' + dest + "/" + osname + ' /var/renasar/on-http/static/http/')
+    os.system('ln -sf ' + dest + "/" + osname + ' /var/renasar/on-tftp/static/tftp/')
+
+def mount_iso(fname):
+    # Mount the ISO to tmp directory
+    tmpdir=tempfile.mkdtemp()
+    if os.path.isdir(tmpdir) and os.access(tmpdir, os.W_OK):
+        try:
+            subprocess.check_call(['mount', '-o', 'loop', fname, tmpdir])
+        except subprocess.CalledProcessError as e:
+            print 'Failed with error code: {0}'.format(e.returncode)
+            os.rmdir(tmpdir)
+            return ''
+    else:
+        print 'Unable to access ISO image'
+        os.rmdir(tmpdir)
+        return ''
+
+    return tmpdir
+
+def determine_os_ver(srcdir, iso_info):
+    osname = ''
+    osver = ''
+    vid = ''
+    m = re.search('^Volume id\:\s+(.+)$', iso_info, re.MULTILINE)
+    if m:
+        vid = m.group(1)
+
+    if 'Application id: ESXIMAGE' in iso_info:
+        osname='ESXi'
+        if 'ESXI-6.0.0' in vid:
+            osver='6.0'
+        elif 'ESXI-5.5' in vid:
+            osver='5.5'
+    else:
+        list=os.listdir(srcdir)
+        if "RPM-GPG-KEY-redhat-release" in list:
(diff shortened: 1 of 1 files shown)


data/templates/setup_iso.py
```diff
@@ -0,0 +1,150 @@
...
+        mount3 = subprocess.Popen(['awk', '{print $1}'], stdin=mount2.stdout, stdout=subprocess.PIPE)
+        isoname = mount3.communicate()[0]
+
+        iso_basename = os.path.basename(isoname).strip()
+        iso_dirname = os.path.dirname(isoname).strip()
+
+        os.makedirs(dstpath)
+        
+        syscall = '(cd "'+iso_dirname+'" && echo "'+iso_basename+'" | cpio -H newc --quiet -L -o )'
+        syscall += ' | gzip -9'
```
benbp: I wonder if it's possible to set PIPEFAIL or something here, that way if any of these steps error, they don't get swallowed by gzip and produce a corrupted initrd image? I've had some painful debugging experiences in the past related to pipelines like this.

benbp:
:+1:

data/templates/setup_iso.py
```diff
@@ -0,0 +1,150 @@
...
+    if tmpdir != '':
+        subprocess.check_call(['umount', tmpdir])
+        os.rmdir(tmpdir)
+
+def get_iso_info(fname):
+    # Access ISO to determine OS type from supported list
+    label1=subprocess.Popen(['isoinfo', '-i', fname, '-d'], stdout=subprocess.PIPE)
+    return label1.stdout.read()
+
+def do_install_os(osname, osver, src, dest):
```
yyscamper: From your implementation for this function, seems the better name is "setup_repo"

data/templates/setup_iso.py
```diff
@@ -0,0 +1,150 @@
...
+import os
+import errno
+import os.path
+import tempfile
+import atexit
+import sys
+import shutil
+import re
+import urllib
+
```
yyscamper: We need a comment on the top of file to describe how to use this script

yyscamper:
Thanks for your work on this, this surely brings convenience for bootstrap feature user. I think you should pay attention to following item:

For CentOS 7.0, ESXi (Both 5.5 and 6.0),  the official ISO file is not workable, I have described the reason in this functional spec: https://hardware.corp.emc.com/display/gheonrack/Extend+OnRack+BootImage+API+to+support+using+external+OS+package+repository

So the ISO file need be the customized one. 

,

yyscamper:
In Shanghai, we setup a mirror server to simplify the OS repo data copying, this mirror server connects with every stack and our develop server, the mirror server exposes data via Samba, so everyone just need to mount the mirror server to the /var/renasar/on-http/static/http folder, saving the copying time.

Meanwhile, the OS repo doesn't required be in the on-http static folder, this repo can be placed in anywhere but under the ORA's control network (currently is 172.31.128.xxx).

@jfrey had another solution for this, setup a proxy in ORA, so computer node has the ability to download from external network, the corresponding  functional spec is  https://hardware.corp.emc.com/display/gheonrack/Provide+HTTP+Proxy+in+MonoRail+to+Installation+Repos, 
@jfrey assigns this task to me, but I haven't started because now I works on some higher priority thing.

yyscamper:
One more comment: this is just an utility script to help people setup OS repository from ISO file, I don't think it is suitable to put it under on-http repo. If we agree this is a feature for on-http, put it under data/templates is also not suitable and I prefer to implement it with Javascript. Move it to "on-tools" folder?  @benbp

zyoung51:
@yyscamper In the short term, I want the flexibility to be able to either run it from a workflow or login and run it directly (the latter reason is why it does not belong in code).  

My preference would be to move it into on-ansible and use the Ansible workflow/tasks or ansible command line to run these types of automation tasks.

zyoung51:
@yyscamper For ESXi, the case issue does not occur when the ISO is mounted and extracted in Linux.  I'm assuming you used 7zip, or some other tool, with a samba mount to extract the ISO, which is what introduced the issue you reference above.  Please update your wiki link to reflect this.

yyscamper:
@zyoung51 , Yes, I used the 7-zip, it will convert the case? Actually, I tested both samba mount and non-samba mount case, my testing shows only lower case can success. If use upper case, installation seems success, but when reboot to the "success" system, it will fail and show fail to find some module.

jfrey:
I'd like to see the proxy work done soon since it will aid the vagrant demo setup making it easier for people to try out the technology.

On Nov 12, 2015, at 6:53 AM, Felix Yuan <notifications@github.com<mailto:notifications@github.com>> wrote:


@zyoung51<https://github.com/zyoung51> , Yes, I used the 7-zip, it will convert the case? Actually, I tested both samba mount and non-samba mount case, my testing shows only lower case can success. If use upper case, installation seems success, but when reboot to the "success" system, it will fail and show fail to find some module.

zyoung51:
@yyscamper I'm not sure how http proxy is 'another solution for this'.  This story is aimed at: "I have an ISO, how do I use the ISO to set up a location to point a netinstall/pxeboot at".  How would enabling http proxy result in an .iso being pxebootable and that .iso working in the OS install workflow as the repo= parameter?

benbp:
Related to the discussion about moving the logic into javascript (@yyscamper @zyoung51): I thought about this a bunch when reviewing, as one alternative would be to break up these steps into more modular jobs (at the expense of more code/complexity but the benefit of greater failure tolerance/debuggability). In the end though, I think the way a lot of people will use this system is by uploading their own custom scripts in languages they are most comfortable with to data/templates, and just using the workflow engine as a way of coordinating serving these scripts. I think this is a totally reasonable use case and one that will be very common, so I'm not opposed to doing it ourselves. 

As Zach mentioned though, adding this as a set of Ansible tasks is probably the ideal solution in the long term.

yyscamper:
@zyoung51 , if we think about from the original repo format, your solution and the http proxy can be thought as complementary. If we think about from how to let it be easier for setup the repository, both solutions are target for it.

I agree people will do get benefit from your work, my biggest concern is the place where this utility file should be in.

yyscamper:
@zyoung51 , as you suggested, if we try to support set an iso file to repo option, just like "repo=redhat.iso", then we call your script to setup the repository under on-http static folder, that will be a good feature. If so, we can move your script to on-tasks/lib/utils/scripts folder.

benbp:
Yeah, if a script file doesn't actually have any template variables, it probably makes sense to not call it a template.

data/templates/setup_iso.py
benbp: Or at least a `-h` flag handler, if you still intend for this to be called manually (as you say below).

zyoung51:
This is very similar to get_smart.sh, monorail_backup/restore.sh which are not templated, but are in this same path.  I'm following the pre-established pattern here.

@benbp Right, this lets a user curl and run the command or use it with downloadUrl in a linux-command type task

@yyscampler I'm not sure why you would want python code in the javascript on-tasks library?

benbp:
Good point. Perhaps we want to move all four (not relevant to this PR I recognize)?

zyoung51:
@benbp Works for me.  Is on-http/data/scripts the right new home?

yyscamper:
ahh, get-smart is an example, but not exactly. get-smart is excuted in node, node need to dowmload it from http. But your script is executed in ORA locally, doesn't involve http.

I just think it is a function for install-os task itself, so put it in on-tasks is reasonable.


Felix

-----Original Message-----

yyscamper:
I prefer to only move this scrip in this PR, open another PR to move other scripts. We'd better do testing after each script moving, I don't want to over load Zach's work this time


Felix

-----Original Message-----

benbp:
Oh I was misunderstanding, this is run on the server itself! This changes my opinion, clearly I wasn't reading closely enough, sorry about that.

In that case, yeah on-http has typically been our holding area for this type of stuff (case in point monorail_backup.sh), but it's not ideal and has usually been for lack of a better location at the time. I think something like on-ansible, as you say, is probably the best long term place for it. Since on-ansible itself isn't getting migrated, I think we are planning on adding another repo at some point with more deployment-type helpers from on-ansible, perhaps https://github.com/RackHD/on-tools would be a good place, I don't know). @heckj do you have any thoughts here? Co-opt on-tools for deployment helpers? Make something else? Throw it into the RackHD repo proper? @RackHD/corecommitters?

@yyscamper I wouldn't want to put this in on-tasks because I'm not sure whether it's the best idea to manage the setup of the actual server via the workflow engine, because I would assume you'd usually want to do that prior to the server running, or at least independently of it. I think something like Ansible as a driver here is a good solution, but we're also in new territory since we aren't being deployed by the proprietary pieces anymore.

TL;DR I'm fine with throwing this into on-http/data/templates for now as a short-term holding area until we have the proper repository set up.

yyscamper:
@zyoung51 @benbp : OK, I agree to put it here as temporary solution just because we had bad example before Zach's PR (monorail-backup.sh, monorail-restore.sh and so on)  ^^.

Do you think how to follow on this, open an ODR or a new functional spec?

amymullins:
I'll create a tracking story for the long term solution.  If it's been agreed the short term solution is acceptable, can we get this merged in?   This "short term" solution fixes a hole that was observed in the MFG Testing with OS Installation that will be very helpful for the near term milestones.

benbp:
:+1: @amymullins

benbp:
Needs one more

jlongever:
:+1:

PR #43, opened 2015-11-11
VEY-230: Add switch configurationa nd vmnic configuration to esx-ks

zyoung51:
- Defining a 'vmnic#' device name in networkDevices will select which vmnic is
  bound to vSwitch0 during initial install.
- 'esxSwitchName' parameter can be specified in networkDevices objects to
  control which vSwitch the port group is bound to.  This example would put
  the pg-name portgroup on vSwitch1 and put vmk# in the pg-name portgroup with
  the proper IP settings:
        {
        "device": "pg-name",
        "esxSwitchName": "vSwitch1",
        "ipv4": {
            "netmask": "255.255.255.0",
            "ipAddr": "172.31.128.10",
            "gateway": "172.31.128.1"
          }
      }
- 'switchDevices' parameter is added to create vswitches and control what uplinks
  are on the switch.  if 'uplink' is not specified, then the switch is created with
  no uplinks (internal switch)

  This would create vSwitch1 with vmnic1 as an uplink, and create "internal name" with no uplinks:
     "switchDevices" : [{ "switchName": "vSwitch1", "uplinks": [ "vmnic1" ] }, { "switchName": "internal name" }],

- 'postInstallCommands' parameter is added to run additional commands during the post-installation
  step of the workflow.

  For example, to create a specific vmk50 on portgroup vmnic0 using dhcp:
  "postInstallCommands": [ "esxcli network ip interface add -i vmk50 -p vmnic0", "esxcli network ip interface ipv4 set -i vmk50 -t dhcp" ]

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -6,7 +6,30 @@ accepteula
 <% } %>
 install --firstdisk --overwritevmfs
 rootpw <%=rootPlainPassword%>
-network --bootproto=dhcp --device=vmnic0
+
+# Search the networkDevices and set the first vmnic# (if defined) up.
+# If no vmnic# device is specified in the networkDevices, then we fallback
+# to setting vmnic0 up as dhcp
+<% if( undefined !== networkDevices ) { %>
+  <% need_default = networkDevices.every(function(n) { %>
+    <% if( n.device.substring(0,5) === 'vmnic' ) { %>
+      <% if(typeof n.ipv4 !== 'undefined') { %>
+        <% ipopts = '--ip=' + n.ipv4.ipAddr + ' --gateway=' + n.ipv4.gateway + ' --netmask=' + n.ipv4.netmask %>
+        <% if (typeof n.ipv4.vlanId !== 'undefined' ) { %>
+          <% ipopts += ' --vlandid=' + n.ipv4.vlanId[0] %>
+        <% } %>
+        network --bootproto=static --device=<%=n.device%> <%=ipopts%>
+      <% } else { %>
+        network --bootproto=dhcp --device=<%=n.device%>
+      <% } %>
+      <% return false; %>
+    <% } %>
+    <% return true; %>
+  <% }); %>
+  <% if (need_default) { %>
+    network --bootproto=dhcp --device=vmnic0
+  <% } %>
+<% } %>
 reboot
 
 %firstboot --interpreter=busybox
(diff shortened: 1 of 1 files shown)


zyoung51:
@RackHD/corecommitters

benbp:
+1

Side thought: I wonder if at this point it's worth adding unit tests for the rendered results of these templates given the complexity they are reaching (also would help provide examples of what we intend to see the template produce).

zyoung51:
Agreed.  @amymullins What do you think about "VEY-236: Validate template output with unittests"?

data/templates/esx-ks
```diff
@@ -114,9 +149,24 @@ cp /var/log/esxi_install.log "/vmfs/volumes/datastore1/firstboot-esxi_install.lo
...
 <% } %>
 
+<% if( typeof postInstallCommands !== 'undefined' ) { %>
+  <% postInstallCommands.forEach(function(n) { %>
+    <%=n%>
+  <% }); %>
+<% } %>
+
+#signify ORA the installation completed
+/usr/bin/wget --spider http://<%=server%>:<%=port%>/api/common/templates/renasar-ansible.pub
```
anhou: see 'renasar' here, maybe need another cleanup PR to rename it to 'rachhd' @RackHD/corecommitters

anhou:
:+1:  after fix messange typo 'configurationa nd', although we all understand it :)

yyscamper:
@zyoung51 , could you please updated related document to guide people how to use this feature?

Meanwhile, if you add a new option in Kickstart file, could you please also update the install-esx task definition (https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/install-esx.js) to tell users they totally have these options to configure.

Considering the template becomes more and more complex, unit-test is sure what we need, since there will be more and more template files are defined in future, we need to design a common way for how to do unit-test for all these files.

There is another aspect I think we also deserve to discuss: move the complex to template file or code; Code can do some conversion for user's input and will make the template design be more simple;  Code can be more simple and common but the template be more complex.

Finally, a question for @benbp , Previously @jfrey tells me we will move to use Mustache template engine, I don't know what's the progress on this. I propose we need to speed up the switch for Mustache, every change on current template file is to increase the workload after Mustache switch.

jlongever:
:+1:

amymullins:
@yyscamper, the FS has been updated this afternoon by Zack (assuming that is the "related document to guide people how to use this feature" you were requesting to be updated above).

Vey-236 is created and will be refined/prioritized in our next refinement meeting and going forward, all folks making updates to this code should follow suit adding in unit test coverage for the new code.

It is unclear why updates here are needed: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/install-esx.js) - it's just a list of the params and missing parameter definition within the actual file.  Is the expectation to make this the documented and public "how-to" guide?

benbp:
A couple points/responses:

@yyscamper re moving the complexity into templates or code: I agree, it's better to, as much as we can, offload logic into the code rather than into the template itself, but at the same time keeping the complexity low in the actual task definition/available options. If we have the option to convert user input in a template or code, and if all else is equal, I think we should always prefer code.

@yyscamper re mustache: I think given that we are now using features in these templates that are more readily available in ejs than various mustache engines, I'm fine with sticking to ejs. I think John and I had a discussion to this effect as well at one point.

@amymullins @zyoung51 re why updates are needed to the install-exs.js task: our convention thus far has been to expose all available options for a job/template render within task definitions as much as possible. This is for a few reasons: partly because it is much easier to inspect the source of the task definition schema rather than the template file, partly because it allows us to do more complete option validation, and finally because we need a location to enumerate available options in order to eventually build forms dynamically in the UI to allow a user to input options to a task when creating one. From my understanding, this template file includes variables that aren't actually included in any of the task definitions/jobs that render this template?

yyscamper:
@amymullins @zyoung51 :  

- Thanks for updating the FS, this is exact what I think we need.

- For Vey-236
Could you please drive a common solution for how to validation template file? I think this will benefit all existing and future template design. Below is the first idea that come up from my mind (you may have better solution): 

build a common function (assume it is 'validateTemplate(inputFilePath, expectFilePath, options)', it will load the input file and render it with options, then comparing the render output with the expect file content. return success if the file content is exactly same.

So for each template file, we just need build some expected files, all validation works share the same routine.

- Please take an action item to write these new options to install-esx task definition, since it is ESXi only, so we don't need them in install-centos.

amymullins:
@benbp, thanks for the details.  @zyoung51 was going to help make the updates within the install-esx.js file.

@yyscamper , going forward we should consider updating the formal documentation (https://github.com/RackHD/docs) and/or comment within the code itself as opposed to extending the scope of an FS.  I don’t see FSs as the answer to communicate all OS Bootstrap Parameters since FSs are supposed to be short lived threads of functionality that are closed upon completion (with an FS lifetime spanning a couple of sprints at most).  This enhancement was made specifically in the context of ORFS-119.  @zyoung51 has already made the updates to ORFS-95, but in the future we will be sure to document in the public channels of RackHD/docs, within the code or the new TBD FS process.

PR #48, opened 2015-11-17
fixing jshint score

tldavies:


heckj:
+1

benbp:
:+1:

tldavies:
Coveralls is red cause code coverage went down.

lib/services/ws-service.js
```diff
@@ -37,7 +37,8 @@ function webSocketServiceFactory(
             this.error(new WebSocketError('Broadcast failed, no server is available.'));
         }
 
-        this.webSocketServer.clients.forEach(wsClientBroadcast.bind(this));
+        var wsService = this;
```
benbp: @rolandpoulter this change ok, or does `this` get bound differently in the closure? I've caused a lot of subtle scoping bugs by doing things like this, so perhaps we just want to add an `ignore: true` line or something?

rolandpoulter: This is fine for this case. The broadcast method isn't being used yet anyways.

tldavies:
@RackHD/corecommitters

PR #50, opened 2015-11-18
Enable CoreOS Install to Disk

king-jam:


benbp:
Should we make /dev/sda a talk option that defaults to sda?

king-jam:
I originally had that support when I made these adjustments months ago. Trying to remember why I backed it out. Should be trivial to support.

benbp:
Perhaps because iPXE enumerates devices with inconsistent orders?

benbp:
Maybe we should support wwid instead of udev names if someone is configuring it?

king-jam:
Looking now to see if the coreos-install binary will take those as arguments

king-jam:
It just takes any blockdev path that is accessible. The checks are done using lsblk so the user can specify any path for input. /dev/sda will be the best default and users can use catalog details to refine their path more.

benbp:
awesome

benbp:
:+1:

heckj:
+1

benbp:
Relies on https://github.com/RackHD/on-tasks/pull/77

PR #52, opened 2015-11-23
enabling running CoreOS

heckj:
- install not working in vagrant demo, so switching that PXE setup to only boot/run CoreOS while I sort out the specifics

### DIFF
--- a/data/profiles/install-coreos.ipxe
+++ b/data/profiles/install-coreos.ipxe
@@ -2,8 +2,9 @@ echo Starting CoreOS Stable installer
 set base-url http://<%=server%>:<%=port%>/coreos
 # coreos.autologin <-- enable as kernel param to autologin for easier debugging
 # reference: https://coreos.com/os/docs/latest/booting-with-ipxe.html
-kernel ${base-url}/coreos_production_pxe.vmlinuz console=tty0 console=<%=comport%>,115200n8 coreos.autologin cloud-config-url=http://<%=server%>:<%=port%>/api/common/templates/pxe-cloud-config.yml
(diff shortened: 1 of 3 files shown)


heckj:
will be re-enabling, but need to sort out how to mirror CoreOS stable channel information locally for the demo. Installation process is failing because VM in vagrant setup doesn't have access to pull the stock image from the internet

rolandpoulter:
:+1:

heckj:
Walking to the bus, I thought of two paths for making the full "install" demo work 

1. use the -b option with coreos_install and have a mirror or relevant files on the RackHD host, maybe leveraging the /coreos/ directory structure that's already there and just lightly expand it

2. for the Vagrant demo setup, add in IP forwarding and set the RackHD server (which does have access to the internet) as the default gateway with DHCP so that files can be retrieved remotely - will also need to shim in some DNS - something like Google's 8.8.8.8 as a default would likely work.

I think the 1st may be faster, and leverage to work in the future with the proxy mechanisms we'll be adding

heckj:
Updated PR to optimize the install flow to pull from a mirror installed onto RackHD - at least until we can more formally update this to utilize a proxy mechanism from RackHD to the internet to pull files as needed. Related PR exists in https://github.com/RackHD/RackHD/pull/14 that puts relevant files into the
right place on a demo setup.

Added annotations/comments into the profile (ipxe script) for future reference on how that flow works and where to go for options when/if additional debugging it needed.

benbp:
:+1:

DavidjohnBlodgett:
:+1:

PR #53, opened 2015-11-26
Add script to get drive IDs information

pengz1:
Add script in on-http/data/templates folder for collecting drive IDs.

### DIFF
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -207,10 +207,20 @@ function run() {
             wwidData = stdout0;
             exec(cmdVdInfo, options, function (err1, stdout1) {
                 if (err1) {
-                    console.error(err1.toString());
-                    process.exit(1);
+                    var errStr= "ls: cannot access /dev/disk/by-path: No such file or directory";
+                    //"/dev/disk/by-path" doesn't exist means only SATADOM exists, no HDDs
+                    // Ignore this error and use null string for vdData
+                    if (err1.toString().search(errStr) !== -1){
+                        vdData = '';
+                    }
+                    else {
+                        console.error(err1.toString());
+                        process.exit(1);
+                    }
                 }
-                vdData = stdout1;
+                else {
+                    vdData = stdout1;
+                }
                 exec(cmdScsiId, options, function (err2, stdout2) {
                     if (err2) {
                         console.error(err2.toString());
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -1,0 +1,194 @@
+#!/usr/bin/env node
+
+'use strict';
+
+var exec = require('child_process').exec;
+
+var cmdDriveWwid = 'ls -l /dev/disk/by-id';
+var cmdVdInfo = 'ls -l /dev/disk/by-path';
+var cmdScsiId = 'lsscsi';
+var options = {
+    timeout: 10000
+};
+
+/**
+ * Parse the Drive WWID output
+ * @param {String} data
+ * @return
+ */
+function parseDriveWwid(idList) {
+    //idList example
+    //lrwxrwxrwx. 1 root root  9 Nov 18 20:30 ata-SATADOM-SV_3SE_20150522AA9992050085 -> ../../sdb
+    var lines = idList.split('\n').map(function(line) {
+        var split = line.split(/\s+/);
+        return [split[8],split[10]].join('->');
+    });
+    //According to SCSI-3 spec, vendor specified logic unit name string is 60
+    var scsiLines = [], sataLines = [], requiredStrLen = 60;
+    for (var i=0; i<lines.length; i++) {
+        var line = lines[i];
+        if ( line && !(line.match('part'))){
+            var nameIndex = line.lastIndexOf('/'), idIndex = line.lastIndexOf('->');
+            if (line.indexOf('scsi') === 0) {
+                scsiLines.push([line.slice(nameIndex+1), line.slice(0,idIndex)]);
+            }
+            else if (line.indexOf('ata') === 0) {
+                sataLines.push([line.slice(nameIndex+1), line.slice(0,idIndex)]);
+            }
+        }
+    }
+
+    var esxiSata = sataLines.map(function(esxiLine) {
+        var line = esxiLine[1];
+        var headIndex = line.indexOf('-'), snIndex = line.lastIndexOf('_');
+        var headStr = ['t10.', line.slice(0, headIndex).toUpperCase(), '_____'].join(''),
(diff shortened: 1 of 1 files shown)


pengz1:
@RackHD/corecommitters @iceiilin @WangWinson

data/templates/get_driveid.js
```diff
@@ -0,0 +1,194 @@
...
+ * @param {String} data
+ * @return
+ */
+function parseVdInfo(pathList) {
+    var lines = pathList.split('\n').map(function(line) {
+        var split = line.split(/\s+/);
+        return [split[8],split[10]].join('->');
+    });
+    var pciLines = [];
+    for (var i=0; i<lines.length; i++) {
```
anhou: not only ++, but also it's better to leverage lodash,  _.forEach()

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env node
+
+'use strict';
+
+var exec = require('child_process').exec;
+
+var cmdDriveWwid = 'ls -l /dev/disk/by-id';
+var cmdVdInfo = 'ls -l /dev/disk/by-path';
```
yyscamper: I think we should change this variable name, this command is not to fetch the virtual drive information

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+ * Parse the Drive WWID output
+ * @param {String} data
+ * @return
+ */
+function parseDriveWwid(idList) {
+    //idList example
+    //lrwxrwxrwx. 1 root root  9 Nov 18 20:30 ata-SATADOM-SV_3SE_20150522AA9992050085 -> ../../sdb
+    var lines = idList.split('\n').map(function(line) {
+        var split = line.split(/\s+/);
+        return [split[8],split[10]].join('->');
```
yyscamper: Could you please add a comment to show an example for the output? It's hard for uses to know the what split[8] and split[10] means.

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+                if(elem[0] === diskPath){
+                    scsiId = elem[1];
+                }
+            }
+        });
+        driveIds.push({"scsiId": scsiId, "virtualDisk": vd,
+            "esxiWwid": esxiWwid[1], "devName": diskPath});
+    });
+    var k=0;
+    linuxWwids.forEach(function(linuxWwid){
```
yyscamper: The internal function for "forEach" can be changed to function(linuxWwid, k), so we don't need to create another index variable "k".

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+* Parse scsi ID output via lsscsi command
+* @param {String} data
+*/
+function parseScsiInfo(lsscsiList) {
+	var lines = lsscsiList.split('\n');
+	return lines.map(function(line) {
+		if(line){
+		return [line.slice(line.lastIndexOf('/') + 1).replace(' ', ''),
```
yyscamper: Incorrect indent?

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+                pciLines.push([line.slice(nameIndex + 1), line.slice(0,idIndex)]);
+            }
+        }
+    });
+
+    return pciLines.map(function(line) {
+        var scsiId = line[1].slice(line[1].lastIndexOf('-') + 1), vdStr;
+        if (scsiId.split(':').length === 4){
+		    var scsiIdArray = scsiId.split(':');
+            vdStr = ['/c', scsiIdArray[0], '/v', scsiIdArray[2]].join('');
```
yyscamper: incorrect indent?

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+            if (line.indexOf('scsi') === 0) {
+                scsiLines.push([line.slice(nameIndex + 1), line.slice(0, idIndex)]);
+            }
+            else if (line.indexOf('ata') === 0) {
+                sataLines.push([line.slice(nameIndex + 1), line.slice(0, idIndex)]);
+            }
+        }
+    });
+
+    var esxiSata = sataLines.map(function(esxiLine) {
```
yyscamper: Could you please add a "TODO" description here, I know the conversion from Linux WWID format to ESXi WWID format is just bases on our experiment on limited hardware, we cannot ensure this works for all hardwares. So the TODO item is to find out the method that is suitable for all hardware.

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+        return [linuxLine[0], '/dev/disk/by-id/' + linuxLine[1]];
+    });
+
+    return {esxiDriveIds: esxiSata.concat(esxiScsi), linuxDriveIds: linuxParsed};
+}
+
+
+/**
+ * Parse Virtual Drive output via by-path output
+ * @param {String} data
```
yyscamper: The param doesn't match with the function argument name.

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+var cmdDriveWwid = 'ls -l /dev/disk/by-id';
+var cmdVdInfo = 'ls -l /dev/disk/by-path';
+var cmdScsiId = 'lsscsi';
+var options = {
+    timeout: 10000
+};
+
+/**
+ * Parse the Drive WWID output
+ * @param {String} data
```
yyscamper: The param doesn't match with the function argument name.

data/templates/get_driveid.js
```diff
@@ -0,0 +1,196 @@
...
+                process.exit(1);
+            }
+            vdData = stdout1;
+            exec(cmdScsiId, options, function(err2, stdout2) {
+                if (err2) {
+                    console.error(err2.toString());
+                    process.exit(1);
+                }
+                scsiData = stdout2;
+                if (buildDriveMap(wwidData, vdData, scsiData)) {
```
benbp: Should this be wrapped in a try/catch to fail gracefully in case the parsing fails, or am I missing parser error handling somewhere?

yyscamper:
@benbp , I suggest we can include some basic npm module  to the micor-kernel, such as lodash, blue-bird promise, so it will make it easy to design such JS script.

data/templates/get_driveid.js
pengz1: Thanks for your review Ben.
I have updated script as you suggested.

yyscamper:
good. :+1:

anhou:
+1

PR #56, opened 2015-12-03
resetting to include a sample SSH key for demo

heckj:
realized the SSH key from legacy (info@renasar) didn't include the private key in any of these directories as an example setup, so I've replaced those templates and included a sample key for use with a vagrant demo.

Yep, whole insecure, but it provides a sample demo setup for now. Need to replace this with API based SSH key loading into the templates


benbp:
:+1:

yyscamper:
:+1:

PR #57, opened 2015-12-03
adding an option 'symbolic link' base argument

heckj:
yeah, I know it's ghetto - but I needed to drop the symlinks into another
directory for the `vagrant up` demo setup, so this option defaults to
`/var/renasar` as the old script, but allows you to add
`--link /home/vagrant/src` to support for Vagrant ISO unpack and installation
instructions I'm cobbling


benbp:
:+1:

zyoung51:
+1

PR #58, opened 2015-12-03
install os only clear partition for target disk, not all

yyscamper:
For ESXi/RHEL/CentOS installation, previously we will clear all disks partition, so the compute node can only install one OS. With this fix, user can install different OS in different disk, the kickstart file will only clear the target drive partition (not all), so different OS can coexist.

@RackHD/corecommitters @iceiilin @sunnyqianzhang @pengz1 @WangWinson


yyscamper:
@zyoung51

zyoung51:
+1

benbp:
:+1:

benbp:
Merge with https://github.com/RackHD/on-tasks/pull/85

PR #59, opened 2015-12-03
GRIP-17: Add scoping to templates

zyoung51:
- Update unittests to switch from findOne to find in on-core
- Depends on GRIP-17 PR in on-core
@RackHD/corecommitters

### DIFF
--- a/spec/lib/api/tasks-spec.js
+++ b/spec/lib/api/tasks-spec.js
@@ -23,7 +23,7 @@ describe('Http.Api.Tasks', function () {
 
         // mock waterline templates to return a template for bootstrap
         waterline = helper.injector.get('Services.Waterline');
-        waterline.templates.findOne = sinon.stub();
+        waterline.templates.find = sinon.stub();
 
         tasksApiService = helper.injector.get('Http.Services.Api.Tasks');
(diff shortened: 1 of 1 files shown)


spec/lib/api/tasks-spec.js
```diff
@@ -23,7 +23,7 @@ describe('Http.Api.Tasks', function () {
 
         // mock waterline templates to return a template for bootstrap
         waterline = helper.injector.get('Services.Waterline');
-        waterline.templates.findOne = sinon.stub();
+        waterline.templates.find = sinon.stub();
```
benbp: Ignoreable.

benbp:
Only the test file was committed, missing the actual changes it seems.

zyoung51:
Just the test files are affected as there are no changes to on-http to use the scope stuff yet.  For the tests, the observable behavior of get/render didn't change, but the tests specifically mock out findOne, which is no longer used.  I had to modify the test to specifically mock out find instead of findOne.

benbp:
Ah ok, sorry I totally missed your "Depends on" comment.

benbp:
:+1:

anhou:
:+1:

PR #63, opened 2015-12-07
fix the case for driveid catalog when only SATADOM in system

yyscamper:
If there is only SATADOM in the system, the previous get_driveid script will fail due to cannot find the directory "/dev/disk/by-path", this script is to ignore such error, so this can ensure we still get correct catalog data for the only SATADOM disk.

@RackHD/corecommitters @iceiilin @pengz1

### DIFF
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -116,6 +116,10 @@ function parseDriveWwid(idList) {
  * @return {Array} include device name, virtual disk name and scsi ID for each disk
  */
 function parseVdInfo(pathList) {
+    if (!pathList) {
+        return [];
+    }
+
     var lines = pathList.split('\n').map(function(line) {
         var split = line.split(/\s+/);
         return [split[8],split[10]].join('->');
@@ -146,6 +150,10 @@ function parseVdInfo(pathList) {
  * @return {Array} include device name and scsi info for each disk
  */
 function parseScsiInfo(lsscsiList) {
+    if (!lsscsiList) {
+        return [];
+    }
+
     var lines = lsscsiList.split('\n');
     return lines.map(function(line) {
         if(line){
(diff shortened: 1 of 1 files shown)


data/templates/get_driveid.js
```diff
@@ -206,11 +214,9 @@ function run() {
             }
             wwidData = stdout0;
             exec(cmdVdInfo, options, function (err1, stdout1) {
-                if (err1) {
-                    console.error(err1.toString());
-                    process.exit(1);
-                }
-                vdData = stdout1;
+                //if only have SATA disk, there will not be /dev/disk/by-path folder, so we should
```
yyscamper: False alarm, Houndci doesn't use our jslintrc

iceiilin:
+1

heckj:
Evening Felix, 

I'm confident that you've verified this manually, but may I suggest that you plan a follow-up pull-request to that separates out the effort of running this program locally from the actual logic and parsing functions within it - basically, make it two files - one that imports and runs the other - and then specifically add unit testing to the file that does the business logic.

There's a lot of relative complexity in that script that will be very sensitive/fragile to the specific output  based on the OS, etc, so having some means of validating it would make a lot of sense.

yyscamper:
@heckj , actually this script is run in compute node, not the code for on-http itself. Computer node will download this script and execute it and then report the data to RackHD service. Currently there is no common framework  to do unit-test for such script, as these scripts may be written in multi language, such as "get_smart.sh" is Shell script, "setup_iso.py" is python script, and this one is JS script.
We will take a follow up to find a framework about how to do unit-test for such script.

Meanwhile, as computer node will download this script to run, so we design it in single script to make it easier to download, this is a comprise between code architecture and interaction logic.

jlongever:
+1

heckj:
@yyscamper totally get that it's running on the compute node - I'm suggesting a means of adding unit tests over the logic in this file. I know it's not run on the RackHD server, but it still merits unit testing.

yyscamper:
@heckj , I agree we'd better have unit-test for such script. I merge it first considering its urgency. 

Do you mean write both the code and unit-test into same file, but use a flag to run unit-test?

heckj:
actually I was suggesting breaking the script up into two files - one that runs the other, so that one of the files is all the functions, and unit tests could be written against that file. 

So one file that's basically:

```
#!/usr/bin/env node
var drive_functions = require('get_drive_functions')()

drive_functions.run()
```

and the other is more or less what you have here already, changing it to provides a constructor function that returns a javascript object with a `run` function that will invoke the code.

That makes sense?

yyscamper:
@heckj , got your idea. I hope this kind of unit-test can be merged into on-http's unit-test framework, so when we run "npm test" will run its unit-test as well. 

I will make a follow up on this and add the unit-test to the master branch.

PR #65, opened 2015-12-09
Modify apidoc about deleting active workflow

iceiilin:
https://hwjiraprd01.corp.emc.com/browse/ODR-346 complaining the inconsistent return code between operation and doc about deleting active workflow api.
Add success code 204 in apidoc, since "204 No Content: The server successfully processed the request, but is not returning any content" from wikipedia; there is no active workflow after this command, so we can view it as success, though no graph id will be returned.


yyscamper:
:+1:

benbp:
+1

PR #67, opened 2015-12-09
adding optional KVM trigger to install KVM related packages

heckj:
to support additional calls into RackHD for installing KVM packages per details at https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/5/html/Virtualization/sect-Virtualization-Installing_the_virtualization_packages-Installing_KVM_packages_on_an_existing_Red_Hat_Enterprise_Linux_system.html

requires https://github.com/RackHD/on-tasks/pull/90 to be functional, but should merge cleanly regardless

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -54,11 +54,11 @@ perl
 libselinux-python
 net-tools
 <% if( undefined !== kvm ) { %>
-kvm
-virt-manager
-libvirt
-libvirt-python
-python-virtinst
+    kvm
+    virt-manager
+    libvirt
+    libvirt-python
+    python-virtinst
 <% } %>
 %end
 
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -53,6 +53,13 @@ sudo
 perl
 libselinux-python
 net-tools
+<% if( undefined !== kvm ) { %>
+kvm
+virt-manager
+libvirt
+libvirt-python
+python-virtinst
+<% } %>
 %end
 
 %pre


data/templates/centos-ks
```diff
@@ -53,6 +53,13 @@ sudo
 perl
 libselinux-python
 net-tools
+<% if( undefined !== kvm ) { %>
```
benbp: Some indentation would be nice.

yyscamper: Since in this PR (https://github.com/RackHD/on-tasks/pull/90/files), the default "kvm" is set to null, here we only judge the "undefined" case, we'd better add a safe-conversion in this function https://github.com/RackHD/on-tasks/blob/master/lib/jobs/install-os.js#L113

heckj: Ah, I think I see what you mean Felix. In this case, the option is just a flat value - either defined or not - for what I'm wanting to trigger, as opposed to an intended array of options, possibly empty, so I'm not sure it makes sense to add this as an option to the `_convertOptions` method that's setting those default `[]` values.

Or are you suggesting that we add it as a default "null" value on the install-os.js job in general, so that it's a clear option to be set on the job/task?

yyscamper: I mean it's better to delete the "kvm" option if it is null, this is just to make the kickstart happy.

Meanwhile, we can also set the default value to "undefined", not "null".

heckj: Understood! I will update!

heckj: https://github.com/RackHD/on-tasks/pull/90 updated with Felix's suggestion

tldavies:
:+1:

yyscamper:
:+1:

heckj:
For future reference, the 4GB Installation ISO that I used to provide my testing repository doesn't include `kvm` or `python-virtinst`, but otherwise I've verified these combined patches work to provide an optional KVM installation, ready to merge!

PR #68, opened 2015-12-14
Add creating proxy function to http-service

pengz1:
Implement http proxy via express-http-proxy module in http-service.js.
Proxy settings should will be input from monorail.json with key name "httpProxies";
"httpProxies" is supposed to be an array and each proxy element could include 3 items:
(1) Server info, which is a must.
(2) Local/remote paths are optional and will be root path by default if not provided.

### DIFF
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -178,7 +178,7 @@ function httpServiceFactory(
 
         var httpProxies = configuration.get('httpProxies');
         if (httpProxies && _.isArray(httpProxies)) {
-            _.forEach(httpProxies, createHttpProxy);
+            _.forEach(httpProxies, createHttpProxy.bind(this, app));
         }
 
         if (configuration.get('httpEnabled', true)) {
@@ -263,7 +263,7 @@ function httpServiceFactory(
      * @param {Object} httpProxy - proxy parameters
      * @returns
      */
-    function createHttpProxy(httpProxy){
+    function createHttpProxy(app, httpProxy){
         try {
             var localPath = httpProxy.localPath || '/',
                 remotePath = httpProxy.remotePath || '/',
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -9,6 +9,8 @@ var onFinished = require('on-finished');
 var http = require('http');
 var https = require('https');
 var fs = require('fs');
+var proxy = require('express-http-proxy');
+var path = require('path');
 
 /**
  * Get remote address of the client.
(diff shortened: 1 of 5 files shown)


yyscamper:
I'm working on the unit-test, so please don't merge until the unit-test is done, but anyway, you can first comment on the code.

zyoung51:
I was wondering if this been benchmarked?  What is the performance impact, if any, of running the proxy within Node vs. using something like squid?

heckj:
@zyoung51 I'm sure it's slower, but it's an easy option that we can build in. Anyone packaging RackHD could easily add a proxy layer like squid+Nginx to get a performance gain if it's suited them.

lib/services/http-service.js
```diff
@@ -229,5 +238,42 @@ function httpServiceFactory(
...
 
+    /**
+     * Creates http proxy
+     * @private
+     * @param {Object} httpProxy - proxy parameters
+     * @returns
+     */
+    function createHttpProxy(httpProxy){
+        try {
+            var localPath = httpProxy.localPath || '/',
```
heckj: if we're already storing static files (say in the directory `common` and you add a configuration for a proxy with `localpath: /common`, which gets precedence?

I think it may be worthwhile always preceding the URL space for the south-bound API set with a unique word to limit potential namespace collisions (i.e. always prefix the URL on the southbound API with `proxy` or something).

If the remote proxy is running on a awkward port - say port 9090 - is that accommodated with the server setting? Any reason to have server and remote path as the totally separate variables they are currently?

heckj:
@pengz1 looking good so far. This PR needs a rebase at this point, and I wanted to ask that you plan on updating https://github.com/RackHD/docs/blob/master/docs/rackhd/configuration.rst with a PR describing the format of the array of proxies allowed in the `monorail.json` file - the 3 sub-components (localPath, remotePath, and server) needed for each proxy are more complex than any of our other configurations today.

heckj:
(-1 until unit tests are added)

pengz1:
@zyoung51 @heckj 
I tried 2 configurations in SH for RHEL bootstrap:
1. Mount mirror server to ORA (use "mount -t cifs"): bootstrap test costs 691s
2. Use proxy codes in this PR: bootstrap test costs 628s
Looks proxy is a little faster than the way we used for bootstrap tests.

I also tried 2 configurations in SH for CentOS bootstrap:
1. Mount mirror server to ORA: bootstrap test costs 691s
2. Use a proxy mirror repo outside EMC (http://mirrors.zju.edu.cn/centos/): bootstrap test costs 1115s
There is performance downgraded but should be acceptable as it is always slower if we want to get a mirror outside our lab.

pengz1:
@RackHD/corecommitters , @yyscamper , @zyoung51 , @jfrey

yyscamper:
@zyoung51 , in this PR, we also fix an unit-test error of SKU pack, please check whether this is expected.

Meanwhile, we found the unit-test of /api/tasks-spec.js will fail in bootstrap.js case, we also fixed it by mocking the template service directly, not via Waterline.

spec/lib/services/sku-pack-service-spec.js
```diff
@@ -51,7 +51,8 @@ describe("SKU Pack Service", function() {
 
     it('should fail to start with an invalid root', function() {
         fs.readdirAsync.rejects();
-        skuService.start('./invalid').should.be.rejected;
+        //SKU service swallowes the exception, so it should not throw error even for invalid root
+        skuService.start('./invalid').should.not.be.rejected;
```
yyscamper: @zyoung51 : please check whether the SKU service should swallow the exception. If you expect not swallow, we will update it.

yyscamper:
@heckj : the unit-test has been added and together with the fix for https://github.com/RackHD/RackHD/issues/54

cgx027:
+1

yyscamper:
:+1:

pengz1:
@heckj 
Unit test is updated, could you help to review and +1 if no other concern?

heckj:
:+1: Unit tests look great, thank you!

yyscamper:
+1

PR #70, opened 2015-12-16
Fix RHEL installtion failure do to kvm issue

pengz1:
It is found that with old centos-ks kvm setting, RHEL bootstrap will fail with kvm errors. New centos-ks can solve this problem.

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -53,7 +53,7 @@ sudo
 perl
 libselinux-python
 net-tools
-<% if( undefined !== kvm ) { %>
+<% if( typeof kvm !== 'undefined' ) { %>
     kvm
     virt-manager
     libvirt


pengz1:
@RackHD/corecommitters @heckj

yyscamper:
I encountered the same issue before but forgot to remind @heckj in this PR (https://github.com/RackHD/on-http/pull/67) :(

Thanks @pengz1 to find it out and fix it. :+1:

heckj:
:+1:

heckj:
@yyscamper and @pengz1 how did you identify the failure? I tried it locally and the previous version worked for me without any errors, so I'm not quite sure what I missed that you ran into. Would like to know for future templating efforts...

pengz1:
I used latest repos for on-http/on-taskgraph/onserve/on-tasks.
On-http will complain kvm error in this case.

pengz1:
@heckj 
You can also find some details in ODR-366.

heckj:
thanks @pengz1 !

PR #73, opened 2015-12-21
rationalizing logs

heckj:
updating mapping to more constricted set of logs per
https://github.com/RackHD/RackHD/issues/34

**NOTE**: Requires https://github.com/RackHD/on-core/pull/48 to be
committed prior as it adds the synonym 'critical' to the existing 'crit'
level.

### DIFF
--- a/lib/services/ws-service.js
+++ b/lib/services/ws-service.js
@@ -48,7 +48,7 @@ function webSocketServiceFactory(
             }
 
             catch (err) {
-                logger.warn('Failed to send message to client: %s'.format(wsConn.id));
+                logger.warning('Failed to send message to client: %s'.format(wsConn.id));
 
                 wsService._closeConnection(wsConn);
             }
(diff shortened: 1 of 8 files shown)


lib/services/ws-service.js
```diff
@@ -151,7 +151,7 @@ function webSocketServiceFactory(
     };
 
     WebSocketService.prototype._forwardResourceMethod = function (method, msg, wsConn, rawMsg) {
-        logger.info('Client: %s requested a resource call.'.format(wsConn.id),
+        logger.debug('Client: %s requested a resource call.'.format(wsConn.id),
```
heckj: @rolandpoulter this sequence seemed more relevant to `debug` than `info`, so I'm suggesting we switch them over. That work for you?

DavidjohnBlodgett:
:+1:

lib/services/ws-service.js
rolandpoulter: Yup, totally agree.

rolandpoulter:
:+1:

yyscamper:
:+1:

PR #74, opened 2015-12-21
Generate swagger client python code

jlongever:
- Cleanup apidoc tags
- Run apidoc to swagger converter
- Generate API python library code
- Add python library to debian package

@RackHD/corecommitters @RackHD/rackhd_dev 

synchronous example:
```
from on_http import NodesApi, ApiClient, Configuration

config = Configuration()
config.debug = True
config.verify_ssl = False

client = ApiClient(host='http://localhost:8080',header_name='Content-Type',header_value='application/json')
nodes = NodesApi(api_client=client)
nodes.api1_1_nodes_get()
print client.last_response.data
```

asynchronous example:
```
def cb_func(resp):
    print 'GET /nodes callback!', resp

thread = nodes.api1_1_nodes_get(callback=cb_func)
```

API module list:
```
~/on-http/python-client/on_http/apis$ ls -l
catalogs_api.py
config_api.py
files_api.py
lookups_api.py
nodes_api.py
obms_api.py
pollers_api.py
profiles_api.py
schemas_api.py
skus_api.py
tasks_api.py
templates_api.py
versions_api.py
workflows_api.py
workflow_tasks_api.py
```


heckj:
-1 : I'm really torn on this issue, but I think including the python client libraries in the service package is a bad precedence to set for several reasons:

 - I think it sets a bad precedence in how do we handle adding other client language libraries: Javascript, Java, Go, Ruby, etc.
 - If you want "this code", the only way to get it from a packaged perspective is to include all the node.js as well
   - There's no obvious cut a release of this independent from the server code - as such, no obvious way to make this available at PyPi (or Go repo, or GEM, or NPM, or...), which would be much more convenient for a variety of client usage solutions, where the end result would be: `pip install python-rackhd` (or some variant name)
 - I'm not clear on what this will look like when we also have a V2 API, where we're going more from swagger.json -> code vs. today's code annotations -> swagger.json. Presumably we'll have two swagger files, not a singular one with two versions embedded within it, then how do we handle the clients?

No qualms with 90% of the effort, but I think the pieces that build and include this into the debian aren't appropriate, and that we should consider making a repository for the python client instead, with it's build set up to do "the right thing" for setting up an end result project with it's own debian, PyPi upload, etc.

heckj:
@jlongever did you mean to close this entirely? I thought most of it was good and valuable - the fixes to the annotations on on-http, etc. I'd love to just set up an additional repo for the python client related packaging and build outputs, but the rest of this is incredibly worthwhile

jlongever:
@heckj I'm going to close this and open a new PR so I can spot check the deployment piece against my own bintray account.

PR #80, opened 2015-12-29
add deleting all types of nodes

iceiilin:
Implement full function about deleting nodes and update its related ones based on "relations" field in nodes/ document. It is invoked by ODR-314 (https://hwjiraprd01.corp.emc.com/browse/ODR-314) requiring removing compute nodes in an enclosure when user asks to delete that enclosure node.

Currently there is only (enclosedBy, encloses) relation in node, but it is easily extensible to cover other relations such as (powers, powerBy) or (compute node use dae, dae used by compute node). When deleting one node, RackHD should also delete or update its corresponding nodes in different conditions.

There are 3 types of operations to deal with corresponding (I call it "target node" in code and here) nodes:
* delete all target nodes. 
   e.g. delete enclosure node -> delete compute node as well
* delete target nodes when it doesn't have other targets in this relation. 
   e.g. delete compute node -> delete enclosure node if this enclosure doesn't have other compute node
* update relation in target node. 
   e.g. delete compute node -> update relations in pdu node

This PR covers all cases above. constants.js in on-core defines mapping table and supported target operations. node-api-service in this PR reads this setting and manipulates database. The required nodes will be deleted only after all its target nodes have been processed successfully.

related PR:
https://github.com/RackHD/on-core/pull/55

### DIFF
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -32,13 +32,12 @@ function nodeApiServiceFactory(
     }
 
     /**
-     * Find nodes downstream that relate with the removing node
+     * Find target nodes that relate with the removing node
      * @param  {Object}     node
      * @param  {Object}     type
-     * @return {Promise}    array of downstream nodes
+     * @return {Promise}    array of target nodes
      */
-    NodeApiService.prototype._findDsNodes = function(node, type) {
-        // Find the enclosure nodes who enclose this compute node
+    NodeApiService.prototype._findTargetNodes = function(node, type) {
         if (!_.has(node, 'relations')) {
             return Promise.resolve();
         }
@@ -48,8 +47,8 @@ function nodeApiServiceFactory(
             return Promise.resolve();
         }
 
-        return Promise.map(relation.targets, function (dsNodeId) {
-            return waterline.nodes.needByIdentifier(dsNodeId)
+        return Promise.map(relation.targets, function (targetNodeId) {
+            return waterline.nodes.needByIdentifier(targetNodeId)
             .catch(function (err) {
                 logger.warning("Error Getting Downstream Node with type " + type,
                     { error: err });
@@ -119,63 +118,71 @@ function nodeApiServiceFactory(
     };
 
     /**
-     * Handle downstream nodes. Update relation type field and remove downstream
+     * Handle target nodes. Update relation type field and remove the target
      * node if needed.
-     * @param  {Object}     dsNodos array of downstream nodes
-     * @param  {Object}     type relation type in upstream node
-     * @param  {Object}     nodeId upstream node id
+     * @param  {Object}     targetNodos array of target nodes
+     * @param  {Object}     type relation type in the removing node
+     * @param  {Object}     nodeId removing node id
      * @return {Promise}
      */
-    NodeApiService.prototype._handleDsNodes = function(dsNodes, type, nodeId) {
+    NodeApiService.prototype._handleTargetNodes = function(targetNodes, type, nodeId) {
         var self = this;
-        var dsType = '';
+        var targetType = '';
         var delOption = null;
 
         if (Constants.NodeRelations.hasOwnProperty(type)) {
-            dsType = Constants.NodeRelations[type].mapping;
-            delOption = Constants.NodeRelations[type].delDsNodeOption;
+            targetType = Constants.NodeRelations[type].mapping;
+            delOption = Constants.NodeRelations[type].delTargetNodeOption;
         }
 
         return Promise.resolve()
         .then(function () {
             if (delOption === 'all') {
-                // Throw error when all downstream nodes need to be deleted
+                // Throw error when all target nodes need to be deleted
                 // but at least one of them have active workflow
-                return Promise.map(dsNodes, function(dsNode) {
-                    return self._delValidityCheck(dsNode.id);
+                return Promise.map(targetNodes, function(targetNode) {
+                    return self._delValidityCheck(targetNode.id);
                 });
             }
         })
         .then(function () {
-            return Promise.map(dsNodes, function (dsNode) {
+            return Promise.map(targetNodes, function (targetNode) {
+                // Delete target nodes if it is reuiqred
                 if (delOption === 'all') {
-                    return self.removeNode(dsNode, dsType);
-                } else {
-                    return self._removeRelation(dsNode, dsType, nodeId)
-                    .then(function (dsNode) {
-                        if (dsNode) {
-                            logger.debug('node updated', {id: dsNode.id, relation: dsType});
-                            if (delOption === 'whenEmpty') {
-                                return self._findEmpty(dsNode, dsType);
-                            } else {
-                                return false;
-                            }
-                        } else {
-                            return false;
-                        }
-                    })
-                    .then(function (del) {
-                        if (del === true) {
-                            return self.removeNode(dsNode, dsType);
-                        }
-                    });
+                    return self.removeNode(targetNode, targetType);
                 }
+
+                // Update relations in the target node
+                return self._removeRelation(targetNode, targetType, nodeId)
+                .then(function (targetNode) {
+
+                    // If target node is unavailable currently, don't need further process
+                    if (!targetNode) {
+                        return false;
+                    }
+
+                    logger.debug('node updated', {id: targetNode.id, relation: targetType});
+
+                    if (delOption === 'whenEmpty') {
+                        // Find out whether there is no targets in this relation type now
+                        // This target node will be deleted in the next promise if no other
+                        // target and delete option is "whenEmpty"
+                        return self._findEmpty(targetNode, targetType);
+                    } else {
+                        return false;
+                    }
+                })
+                .then(function (del) {
+                    if (del === true) {
+                        return self.removeNode(targetNode, targetType);
+                    }
+                });
             });
         });
     };
 
     /**
-     * Check whether a node has no active workflow and can be deleted
+     * Check whether a node is valid to be deleted
      * @param  {String}     nodeId
      * @return {Promise}
      */
@@ -183,6 +190,7 @@ function nodeApiServiceFactory(
         return taskGraphProtocol.getActiveTaskGraph( { target: nodeId })
         .then(function (graph) {
             if (graph) {
+                // If there is active workflow, the node cannot be deleted
                 throw new Errors.BadRequestError('Could not remove node ' + nodeId +
                     ', active workflow is running');
             }
@@ -214,10 +222,10 @@ function nodeApiServiceFactory(
                     return Promise.resolve();
                 }
 
-                // Otherwise update its downstream node
-                return self._findDsNodes(node, type)
-                .then(function (dsNodes) {
-                    return self._handleDsNodes(dsNodes, type, node.id);
+                // Otherwise update targets node in its "relationType"
+                return self._findTargetNodes(node, type)
+                .then(function (targetNodes) {
+                    return self._handleTargetNodes(targetNodes, type, node.id);
                 });
             });
         })
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -13,7 +13,8 @@ di.annotate(nodeApiServiceFactory,
         'Errors',
         'Logger',
         '_',
-        'Promise'
+        'Promise',
+        'Constants'
     )
 );
 function nodeApiServiceFactory(
@@ -22,7 +23,8 @@ function nodeApiServiceFactory(
     Errors,
     Logger,
     _,
-    Promise
+    Promise,
+    Constants
 ) {
     var logger = Logger.initialize(nodeApiServiceFactory);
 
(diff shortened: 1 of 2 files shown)


iceiilin:
Codes uses new constants defined in on-core (PR RackHD/on-core#55). The success of unit test depends on it. PR RackHD/on-core#55 would be merged to master first and rerun the test here.

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,161 @@ function nodeApiServiceFactory(
...
     /**
-     * Find Enclosure nodes which enclose compute node
+     * Find nodes downstream that relate with the removing node
      * @param  {Object}     node
-     * @return {Promise}
+     * @param  {Object}     type
+     * @return {Promise}    array of downstream nodes
      */
-    NodeApiService.prototype._findEnclNodes = function(node) {
+    NodeApiService.prototype._findDsNodes = function(node, type) {
```
yyscamper: what does "ds" mean?

iceiilin: DsNodes = downstream nodes, I stated the latter in PR description and the doc for this function.

yyscamper: I prefer to use "downstream" instead of the short name "ds" though it is a long word. "ds" is not a conventional abbreviation.

iceiilin: I have thought the name a lot before when coding, and got to know from HW engineer that "downstream port" is abbreviated as "DSP", thus I chose the variable name here as it is now (ds). But now it seems that it still brings confusion, changing the function name to "downstream" for clear understanding and keeping the local variable as "ds" for direct use might be good choice.

iceiilin:
@yyscamper @anhou I rename all "downstream"  to "target" node to make it clearer.

lib/services/nodes-api-service.js
```diff
@@ -96,15 +203,31 @@ function nodeApiServiceFactory(
...
 
-        return taskGraphProtocol.getActiveTaskGraph( { target: node.id })
-        .then(function (graph) {
-            if (graph) {
-                throw new Errors.BadRequestError('Could not remove node ' + node.id +
-                    ', active workflow is running');
+        return self._delValidityCheck(node.id)
+        .then(function () {
+            if (!node.hasOwnProperty('relations')) {
+                return Promise.resolve;
```
WangWinson: Shouldn't be ```Promise.resolve()```  ?

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
     }
 
     /**
-     * Find Enclosure nodes which enclose compute node
+     * Find target nodes that relate with the removing node
      * @param  {Object}     node
-     * @return {Promise}
+     * @param  {Object}     type
```
yyscamper: "type" is String.

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
             return Promise.resolve();
         }
 
-        return Promise.map(relation.targets, function (enclNodeId) {
-            return waterline.nodes.needByIdentifier(enclNodeId)
+        return Promise.map(relation.targets, function (targetNodeId) {
+            return waterline.nodes.needByIdentifier(targetNodeId)
             .catch(function (err) {
-                logger.warning("Error Getting Enclosure Node", { error: err });
+                logger.warning("Error Getting Downstream Node with type " + type,
```
yyscamper: Should the word "Downstream" be modified as now you have all changed it to "Target"?

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
+        return waterline.nodes.updateByIdentifier(node.id,
+                                                  {relations: node.relations});
+    };
+
+    /**
+     * Handle target nodes. Update relation type field and remove the target
+     * node if needed.
+     * @param  {Object}     targetNodos array of target nodes
+     * @param  {Object}     type relation type in the removing node
+     * @param  {Object}     nodeId removing node id
```
yyscamper: These parameters type is not OK, "targetNodes" should be Array, "type" should be String, "nodeId" should be String

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
+     * @return {Promise}
+     */
+    NodeApiService.prototype._delValidityCheck = function(nodeId) {
+        return taskGraphProtocol.getActiveTaskGraph( { target: nodeId })
+        .then(function (graph) {
+            if (graph) {
+                // If there is active workflow, the node cannot be deleted
+                throw new Errors.BadRequestError('Could not remove node ' + nodeId +
+                    ', active workflow is running');
             }
```
yyscamper: This can be "return Promise.rejects(new Errors.BadRequestError('xxxx');

lib/services/nodes-api-service.js
```diff
@@ -96,15 +170,31 @@ function nodeApiServiceFactory(
      * @param  {Object}     node
      * @return {Promise}
```
yyscamper: Still JSDoc error, node should be String, and miss the param "srcType"

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -73,88 +40,84 @@ describe("Http.Services.Api.Nodes", function () {
...
-            this.sandbox.stub(waterline.nodes, 'needByIdentifier').rejects(Errors.NotFoundError(''));
+        it("_findTargetNodes should return nothing if cannot find enclosure node", function() {
+            needByIdentifier.rejects(Errors.NotFoundError(''));
 
-            return nodeApiService._findEnclNodes(computeNode)
+            return nodeApiService._findTargetNodes(computeNode, 'enclosedBy')
             .then(function (nodes) {
-                expect(waterline.nodes.needByIdentifier).to.have.been.calledOnce;
+                expect(needByIdentifier).to.have.been.calledOnce;
                 expect(nodes[0]).to.equal(undefined);
```
yyscamper: If target node cannot be found, I think return empty array is better if you don't want throw error. Because I see you have a lot of Promise.map for the returned target node (undefined) in your code.

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -227,25 +186,268 @@ describe("Http.Services.Api.Nodes", function () {
...
+                        ]
+                    }
+                ]
+            };
+            var enclRelationAfter = [];
+            return nodeApiService._removeRelation(enclNode, 'encloses', computeNode.id)
             .then(function () {
+                expect(updateByIdentifier)
+                    .to.have.been.calledWith(enclNode.id,
+                                             {relations: enclRelationAfter});
```
yyscamper: From your test case description, we need an assertion to check the relation filed is empty, such as "expect(enclNode).to.have.property('relation').to.deep.equal([])".

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -227,25 +186,268 @@ describe("Http.Services.Api.Nodes", function () {
...
+            needByIdentifier.withArgs(computeNode2.id).resolves(computeNode2);
+            updateByIdentifier.withArgs(computeNode.id).resolves(computeNodeAfter);
+            updateByIdentifier.withArgs(computeNode2.id).resolves(computeNode2After);
+
+            return nodeApiService.removeNode(enclosureNode)
+            .catch(function (err){
+                expect(err.name).to.equal('BadRequestError');
+                expect(waterline.nodes.destroy).to.not.have.been.called;
+                expect(updateByIdentifier).to.not.have.been.called;
+            });
```
yyscamper: This is a chance that if really the removeNode resolves not rejects, then all assertion in the "catch" block will not be executed, then the test case passes because it doesn't do any assertion. We'd better check the removeNode is actually rejects, or to explicitly fail the test is resolved:

return xxx.removeNode()
.catch() {
xxxx
}.then() {
   return Promise.rejects('xxx');
}

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
+     * @param  {Object}     node node whose relation needs to be updated
+     * @param  {String}     type relation type that needs to be updated
+     * @param  {String}     targetId id in relation type that needs to be deleted
+     * @return {Promise}    node after removing relation
      */
-    NodeApiService.prototype._removeEnclRelations = function(node, enclNodes) {
-        if (!enclNodes) {
+    NodeApiService.prototype._removeRelation = function(node, type, targetId) {
+
+        if (!node) {
```
heckj: do you want to resolve an empty promise here if node isn't specified, or throw an error? I think if the node isn't defined, or if the node doesn't have any relations, you may be better off being explicit about throwing an error. It'll make this function (and those composed from them) much easier to test.

heckj: Either that, or update the description of this function/method to make it clear you're intending that a failure to find/remove the relation specified will result in a Promise with an empty value as opposed to an explicit failure/throw

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
-            // remove this enclosure node, else remove relationships
-            if (enclNode.relations[index].targets.length === 0) {
-                return waterline.nodes.destroy({ id: enclNode.id });
-            } else {
-                return waterline.nodes.updateByIdentifier(
-                    enclNode.id, { relations : enclNode.relations });
+        })
+        .then(function () {
+            return Promise.map(targetNodes, function (targetNode) {
+                // Delete target nodes if it is reuiqred
```
heckj: `reuigred` should be `required`

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
+
+    /**
+     * Handle target nodes. Update relation type field and remove the target
+     * node if needed.
+     * @param  {Object}     targetNodos array of target nodes
+     * @param  {Object}     type relation type in the removing node
+     * @param  {Object}     nodeId removing node id
+     * @return {Promise}
+     */
+    NodeApiService.prototype._handleTargetNodes = function(targetNodes, type, nodeId) {
```
heckj: Suggestion only: you might consider naming this `_deleteTargetNodes` instead of `_handleTargetNodes` since this method appears to have the key logic behind when a node should be deleted and if an error should be thrown (for example, when a workflow is operational on the node)

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
+        var targetType = '';
+        var relationClass = null;
+        var needDel = false;
+
+        if (Constants.NodeRelations.hasOwnProperty(type)) {
+            targetType = Constants.NodeRelations[type].mapping;
+            relationClass = Constants.NodeRelations[type].relationClass;
+        }
+
+        if ((relationClass === 'component') &&
```
heckj: this would probably make more sense to include this logic (to delete or not) within the promise chain (around line 125) since you're starting with a single node in this function and cascading to build a map of all nodes you want to delete. If you're expecting an input of nodes, then you may want to construct a promise chain from that input (assuming the input is an array of nodes) and utilize Promise.each.

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
+        }
 
-            if (_.indexOf(enclNode.relations[index].targets, node.id) !== -1) {
-                _.pull(enclNode.relations[index].targets, node.id);
+        return Promise.resolve()
+        .then(function () {
+            if (needDel === true) {
+                // Throw error when all target nodes need to be deleted
+                // but at least one of them have active workflow
+                return Promise.map(targetNodes, function(targetNode) {
```
heckj: Is the intent that if ANY of the nodes in the target list aren't ready to be deleted, that this entire call should "fail"?

lib/services/nodes-api-service.js
```diff
@@ -30,64 +32,136 @@ function nodeApiServiceFactory(
...
-            if (enclNode.relations[index].targets.length === 0) {
-                return waterline.nodes.destroy({ id: enclNode.id });
-            } else {
-                return waterline.nodes.updateByIdentifier(
-                    enclNode.id, { relations : enclNode.relations });
+        })
+        .then(function () {
+            return Promise.map(targetNodes, function (targetNode) {
+                // Delete target nodes if it is reuiqred
+                if (needDel === true) {
```
heckj: Do you want to remove the node and then the relation, or the relation and then the node? Or does it matter?

lib/services/nodes-api-service.js
heckj: Is this intended to take in a node object, or a nodeID string?

spec/lib/services/nodes-api-service-spec.js
iceiilin: Agree

lib/services/nodes-api-service.js
iceiilin: node object is needed here

spec/lib/services/nodes-api-service-spec.js
iceiilin: This case verifies the relation field in enclNode will be empty after updating, so I check the input parameter for updateByIdentifier().

spec/lib/services/nodes-api-service-spec.js
iceiilin: Good catch. Will update.

lib/services/nodes-api-service.js
iceiilin: I incline to remain it as Promise.resolve(). There is possibility that some node doesn't have desired relations. This function will not update mongoDB and ignore it silently. I will update the function description.

lib/services/nodes-api-service.js
iceiilin: I agree that "handle" is an not explicit action. But this function deals with target nodes with different actions, like delete and update, and "handle" might cover them

lib/services/nodes-api-service.js
iceiilin: Yes. The node that user requires to delete will be removed only after all its target nodes has been deleted (if their relation requires) successfully.

lib/services/nodes-api-service.js
iceiilin: As commented above, remove/update target nodes, and then the node itself. I don't want the target nodes to keep non-existing relations, if the node has been deleted, while the target nodes update fails. What's your opinion?

lib/services/nodes-api-service.js
iceiilin: If I understand your idea correctly, you suggested to move the the needDel assignment just before dealing with every node. I would argue that this function (_handleTargetNodes) deals with all target nodes in one relation (like {relationType:"encloses", targets:['aaa', 'bbb'...]}), thus the deletion action is the same.

benbp:
I know it's a pain, but could you add a `_ = helper.injector.get('_')` (and the subsequent `var _` at the top) to the nodes-api-spec.js file? I removed `_` from the globals exceptions in the .jshintrc, because hound-ci won't let us use a separate spec file for test. Since that's the case, I don't want the linter to treat `_` as global in non-test code.

lib/services/nodes-api-service.js
benbp: Also it's misspelled "targetNodos" instead of "targetNodes".

lib/services/nodes-api-service.js
benbp: Yeah, I see some drawbacks to both approaches:

If we enforce throwing an error, it means that we can never repair hung states through the API (e.g. if a target relation is deleted for some reason, we can now never successfully delete other nodes with a link relation to it).

If we don't throw an error, there could be cases where an intermittent connectivity failure to the database causes us to only partially delete relations when we really meant to delete all of them.

I'm not sure which drawback we want to avoid and which we want to accept. Thoughts?

iceiilin: @benbp current code doesn't hung other nodes from processing, as you mentioned in the first "if" situation, and it also uncovers the real error from database. This function (_removeRelation) throws an error when waterline.nodes.updateByIdentifier() failes; the previous actions, that always return Promise.resolve() when it don't have valid targets, is only about some logic check and don't touch asynchronous process.

iceiilin:
@benbp done for adding local '_' in unittest

spec/lib/services/nodes-api-service-spec.js
yyscamper: Got it. Just realized we cannot test the db output as it is mocked.

yyscamper:
This should be the top 1 PR with highest number of comment :)
:+1:

anhou:
:+1:

PR #84, opened 2016-01-05
Upgrade lodash to version 3.10.1

rolandpoulter:
Depends on https://github.com/RackHD/on-core/pull/61

Tests should fail until `on-core` PR is merged.


package.json
```diff
@@ -24,8 +24,7 @@
     "express": "^4.10.7",
     "express-http-proxy": "^0.6.0",
     "glob": "^4.3.2",
-    "lodash": "^2.4.1",
-    "lru-cache": "^2.5.0",
```
rolandpoulter: Removed `lru-cache` because it was not being used and is already included by `on-core`

index.js
```diff
@@ -72,7 +72,7 @@ function run() {
         });
 
     process.on('SIGINT', function() {
-        http.stop()
```
rolandpoulter: This was causing an uncaught exception when the process was interrupted.

benbp: Thanks, I was noticing this too but forgot to add it to my last PR.

benbp:
:+1:

yyscamper:
:+1:

PR #87, opened 2016-01-08
Adds web ui install script

rolandpoulter:
Github provides a nice ability to download a zip file for any branch. Since the UI is hosted on `gh-pages` it can be used to install the UI for `on-http`. I created a script that will download and extract the UI and serve it at `/ui`. This script will get automatically run with `npm install`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,9 @@ Note: requires MongoDB and RabbitMQ to be running to start correctly.
 
     sudo node index.js
 
+ * http://127.0.0.1/ui -- RackHD Web UI
+ * http://127.0.0.1/docs -- RackHD Docs
+
 ## Config
 
 The `fileService` requires a "fileService" key which holds keys mapping backend
--- a/.gitignore
+++ b/.gitignore
@@ -41,9 +41,11 @@ node_modules
 .idea
 
 # static images
+static/ansible/
 static/files/
 static/http/
-static/tftp/
 static/smb/
-static/ansible/
+static/tftp/
+static/web-ui/
+
 build/
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -184,4 +184,3 @@ function prepareGrunt(grunt) {
     // By default, lint and run all tests.
     grunt.registerTask('default', ['jshint', 'coverage', 'swagger']);
 }
-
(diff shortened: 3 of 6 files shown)


jlongever:
+1

DavidjohnBlodgett:
:+1:

README.md
```diff
@@ -16,6 +16,9 @@ Note: requires MongoDB and RabbitMQ to be running to start correctly.
 
     sudo node index.js
 
+ * http://127.0.0.1:80/ui -- MonoRail Web UI
```
anhou: 'MonoRail' should be 'RackHD' , '80' should be  '8080', 80 port conflicts with OnRack projects's http port, or is it intended?

rolandpoulter: Agreed it should say 'RackHD', however the default port `on-http` uses is 80.

jlongever:
Can one of the admins verify this patch?

jlongever:
Can one of the admins verify this patch?

README.md
anhou: @rolandpoulter I double confirm that the default port of `on-http` is 8080, or your browser's default port is set to 8080, so  you could access 127.0.0.1/docs or 127.0.0.1/ui  correctly without explicit 8080 port? I apply this PR's patch and verify again, the port is 8080, if I miss something, please correct me

WangWinson: When you mention **default** port, does it mean the one hard coded in `http-service.js` or the one in config file `monorail.json`. It will be 80 if no `httpBindPort` key found in config, but we always put 8080 in config.

tldavies: The default is the one hard coded into http-service.js.

While the one in the config file monorail.json will overwrite the default in http-service.js.

tldavies:
@jlongever When you say admins are you talking about @RackHD/corecommitters?

jlongever:
@tldavies Oh, ignore that comment.  It's from an internal Jenkins PR trigger,  it's disabled now.

README.md
rolandpoulter: Thanks @tldavies.

@WangWinson default is whatever port `on-http` uses without any configuration. In this case that is what is hard coded in `http-service.js`.

@anhou my web browser uses port 80 by default same as any other. If you try to run `on-http` by itself according to the README you run `sudo node index.js`, the reason you have to use `sudo` is because by **default** `on-http` uses 80 which requires root access to bind.

We generally run on `8080` to allow it to run alongside something that is already using 80.

benbp:
:+1:

WangWinson:
:+1:

PR #89, opened 2016-01-13
Added sku pack upload/delete routes:

zyoung51:
- PUT /skus/pack: 'curl -sT pack.tar.gz localhost:8080/api/common/skus/pack'
- PUT /skus/:id/pack: 'curl -sT pack.tar.gz localhost:8080/api/1.1/skus/:id/pack'
- DELETE /skus/:id/pack: 'curl -X DELETE localhost:8080/api/1.1/skus/:id/pack'

Only .tar.gz files are supported.  The tarball shall have the format described
in /skupack.d/README.md


zyoung51:
@RackHD/corecommitters @tannoa2 @uppalk1 @keedya

lib/services/sku-pack-service.js
```diff
@@ -82,6 +90,35 @@ function skuPackServiceFactory(
...
+                var httpTemplateRoot = path.resolve('/', conf.httpTemplateRoot);
+                httpTemplateRoot = self.confRoot + '/' + skuid + httpTemplateRoot;
+                promises.push(fs.readdirAsync(httpTemplateRoot));
+            }
+        } catch(error) {
+            logger.warning('Unable to unregister sku configuration for ' + skuid);
+            throw error;
+        }
+        return Promise.map(promises, function(entries) {
+            return Promise.all(_.map(entries, function(entry) {
```
benbp: Doesn't really matter, but you could also do `Promise.map` here like on the previous line.

lib/services/sku-pack-service.js
```diff
@@ -95,5 +132,96 @@ function skuPackServiceFactory(
...
+    };
+
+    SkuPackService.prototype.deletePack = function(skuid) {
+        var self = this;
+        return fs.readFileAsync(self.confRoot + '/' + skuid + '.json')
+            .then(function(contents) {
+                return self.unregisterPack(contents, skuid);
+            })
+            .then(function() {
+                return rimrafAsync(self.confRoot + '/' + skuid);
```
benbp: Just to be extra careful, do you think it's worth asserting that `self.confRoot` is not null or an empty string?

benbp: Since we do run HTTP with root permissions.

zyoung51: Why are we running as root?

package.json
```diff
@@ -31,7 +31,10 @@
     "on-finished": "^2.2.0",
     "on-tasks": "git+https://github.com/RackHD/on-tasks.git",
     "pluralize": "~1.1.2",
-    "ws": "^0.8.0"
+    "ws": "^0.8.0",
+    "os-tmpdir": "~1.0.1",
+    "rimraf": "~2.5.0",
+    "tar": "~2.2.1"
```
yyscamper: if you use npm to automatically update the package.json, npm will arrange these modules name in alphabetical order
```bash
npm install os-tmpdir --save     #auto insert this module to the dependencies list of package.json
npm install os-tmpdir --save-dev   #auto insert this module t o the devDependencies list of packjson.json
```

lib/api/skus.js
```diff
@@ -14,12 +17,64 @@ di.annotate(skusRouterFactory, new di.Inject(
...
+    taskGraphProtocol, 
+    uuid, 
+    skuPack, 
+    tmp,
+    rimraf
+) {
     var router = express.Router();
+    var rimrafAsync = Promise.promisify(rimraf);
+
+    router.skuPackHandler = function(req,res,skuid) {
```
yyscamper: Please add function documentation for your created new function follows [JSDoc](http://usejsdoc.org/) style.

lib/api/skus.js
```diff
@@ -92,6 +147,19 @@ function skusRouterFactory (waterline, rest, _, Promise, taskGraphProtocol) {
     }));
 
     /**
+     * @api {post} /api/1.1/skus/pack PUT /
```
anhou: 'post' should be 'put'

lib/services/sku-pack-service.js
```diff
@@ -95,5 +132,96 @@ function skuPackServiceFactory(
...
+                }
+                return true;
+            });
+        } catch(error) {
+            return Promise.reject(error);
+        }
+        return Promise.resolve(true);
+    };
+
+    SkuPackService.prototype.installPack = function(fromRoot, skuid) {
```
yyscamper: please add the function description

jlongever:
You can ignore the jenkins failure

+1

jlongever:
+1

keedya:
:+1:

lib/api/skus.js
```diff
@@ -118,6 +186,58 @@ function skusRouterFactory (waterline, rest, _, Promise, taskGraphProtocol) {
...
+     *     HTTP/1.1 500 Internal Server Error
+     *     {
+     *       "error": "error message"
+     *     }
+     */
+
+    router.put('/skus/:identifier/pack', function(req, res) {
+        return waterline.skus.needByIdentifier(req.params.identifier)
+            .then(function() {
+                return router.skuPackHandler(req,res,req.params.identifier);
```
yyscamper: better to have a space between each argument

lib/api/skus.js
yyscamper: I'm thinking bind the our own function to express Rounter is good or not. Here we bind to router should only because we need to visit it in unittest, otherwise keep it as internal function in skus.js should work fine as well. So whether we need to sacrifice code architecture design for unittest. Add other @RackHD/corecommitters  to comment.

anhou:
:+1:

yyscamper:
To unpack the SKU tar ball, your design need import many external command utilities, but another solution is design a script (prefer python) to isolate the SKU pack install work, create a ChildProcess to trigger the script. I just think this design is more straightforward and easy to maintain. Of course, the unittest design is a challenge.

PR #90, opened 2016-01-14
Inital support for V2.0 API with Swagger

geoff-reid:
@RackHD/corecommitters @brianparry @abildw @dalebremner @srinia6


jlongever:
test this please

lib/api/1.1/versions.js
```diff
@@ -0,0 +1,53 @@
+// Copyright 2015, EMC, Inc.
```
benbp: It's 2016 now, lol :)

lib/services/http-service.js
```diff
@@ -142,6 +144,17 @@ function httpServiceFactory(
...
         }
 
+        //Swagger support
+        var swaggerConfig = {
+            appRoot: path.resolve(__dirname, '../../'),
+            configDir: path.resolve(__dirname, '../../swagger_config'),
+            swagger: path.resolve(__dirname, '../../static/monorail-2.0.yaml'),
+        };
+        swagger.create(swaggerConfig, function(err, swaggerExpress) {
+            if (err) throw(err);
```
benbp: Yeah we try to avoid braceless if statements as a style rule.

lib/services/http-service.js
```diff
@@ -149,7 +162,7 @@ function httpServiceFactory(
         // Imaging Event Middleware
         app.use(httpEventMiddleware);
 
-        // Override default static directory with sku specific handlers
+        // Override default static eirectory with sku specific handlers
```
benbp: `eirectory`

lib/services/versions-api-service.js
```diff
@@ -2,49 +2,36 @@
...
-     *
+     * Find the installed versions
+     * @param {Object} [req] HTTP request
+     * @param {Object} [res] HTTP response
+     * @return {Promise}
      */
+    VersionApiService.prototype.findVersion = function(req, res) {
 
-    router.get('/versions', rest(function (req, res) {
+        // Find the enclosure nodes who enclose this compute node
```
benbp: Should this comment be removed? Doesn't really seem to make sense here.

benbp:
A couple minor fixes (on commented out code). :+1: after those are addressed.

lib/services/http-service.js
```diff
@@ -142,6 +144,17 @@ function httpServiceFactory(
             });
         }
 
+        //Swagger support
+        var swaggerConfig = {
+            appRoot: path.resolve(__dirname, '../../'),
+            configDir: path.resolve(__dirname, '../../swagger_config'),
+            swagger: path.resolve(__dirname, '../../static/monorail-2.0.yaml'),
```
WangWinson: I think @jlongever has a PR that could leverage `api-doc` to generate `swagger.json`, which I think could just satisfy swagger UI and make it work. Why not align with his solution ? If in that way, do we still need the `monorail-2.0.yaml` ?

WangWinson:
Looks like we are leveraging `swagger-express` in v2 API. This change the way we doing in v1 API. In v1, we write `router` in `express` style, while, in v2 looks like we don't need it. `monorail-2.0.yaml` will cover this , right ?
I have concern on merging API auth functionality to v2 API later, if it's now in the `express` style. @anhou @cgx027

anhou:
@geoff-reid Be curious that use https://www.npmjs.com/package/swagger-express-mw  rather than https://github.com/swagger-api/swagger-node which is more popular. Just think the popular module's potential issues could be fixed soon, or be continuously maintained.

lib/services/http-service.js
dalebremner: @benbp - Thanks! I made the changes...
@WangWinson Thanks for the feedback. Since we have additional work to do on the swagger integration, we'll consider this for future PRs.

brianparry: @WangWinson, @jlongever's apidoc solution is a way of generating a swagger.json from doc comments.  Since we are actually installing swagger middleware now, swagger.json is generated from swagger.yaml by the middleware itself and thus the apidoc method is no longer needed.

jlongever: just to clarify,  the apidoc-swagger won't bee needed for 2.0 APIs since those will use swagger.  It's still needed to generate the swagger.json for existing 1.1 APIs (anything using apidoc annotation).

brianparry:
@anhou This _is_ actually based on https://github.com/swagger-api/swagger-node.  Swagger node is a command line tool that can generate swagger node projects consisting of an app with middleware installed, a config file, a swagger.yaml file and a controllers directory.  The generated projects use swagger-express-mw.  Our approach is very closely related to the sample code generated by swagger-node.

geoff-reid:
Rebased with latest version from master

dalebremner:
:+1:

lib/services/versions-api-service.js
```diff
@@ -1,50 +1,37 @@
...
-     *
+     * Find the installed versions
+     * @param {Object} [req] HTTP request
+     * @param {Object} [res] HTTP response
+     * @return {Promise}
      */
+    VersionApiService.prototype.findVersion = function(req, res) {
 
-    router.get('/versions', rest(function (req, res) {
+        // Find installed packages
```
heckj: this was horrifically ghetto (and I wrote it) and makes us specific to debian (as well as VM based deployments) for talking about versions. Based on the other pieces I'm reading above, we'd be better supported by 'versions' being relevant to the versions of the API we support, rather than "what software versions are also on this machine". Is that what you were heading for? Or is this simply more an "here's an example of an API in both 1.1 and 2.0 w/ swagger"?

(it's terrible when you're embarrassed by code you wrote some time ago to solve a point solution :-) )

abildw: Hi Joe,

Our goal was to get the middleware integrated and Versions was one of the simpler routes to implement so we went with that one.

heckj:
I have a general question about intent with the semantic meaning of /versions URI stubs, but that shouldn't block any merges, it's just me being nosey. :+1:

lib/services/versions-api-service.js
heckj: Awesome, makes sense! Thank you! :+1:

jlongever:
+1

PR #92, opened 2016-01-15
Change ipxe's undionly.kpxe to syslinux's undionly.kkpxe

anhou:
it's the same with master's PR https://github.com/RackHD/on-http/pull/91 , it's for 1.1 release
undionly.kpxe from ipxe is not the original gPXE undionly.kkpxe from syslinux which has been validated for installing ESXi on multiple platforms, correct it here. This will fix the chainload failure issue when installing ESXi OS


iceiilin:
+1

yyscamper:
:+1:

yyscamper:
Merge this due to the urgency of 1.1.x, and We identify this PR should not be the cause of Jenkins test failure.

PR #93, opened 2016-01-19
Feature/rack hd v1.1 api authentication

cgx027:
# What does the code do:

This code will shows how we setup **v1.1** authentication and also how we split northbound and southbound API

# Reference 
1. Spec: 
  * https://github.com/RackHD/specs/blob/master/v1.1_api_authentication.md
2. Google group discussion:
  * [RackHD v1.1 API authentication use case and plan review](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/rackhd/Hw88shMffLw)
  * [v1.1 API authentication --split southbound and northbound APIs](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!searchin/rackhd/split/rackhd/PmP2hosgJr8/eyvuBsVlBAAJ)
  * [Separate North/South API](https://groups.google.com/forum/#!topic/rackhd/n18-mWelhVs)

# Design Notes
* Authentication
    * Passport framework used for authentication
    * Passport-local middleware used for username/password authn when accessing /login, a token will be returned after successful authz. 
    * crypto.pbkdf2 used to generate the hash for the default password with a default salt generated using crypto.randomBytes. The default hash and salt are stored in the configure file (/opt/onrack/etc/monorail.json). Password posted from user is hashed and then compared to the hash read from configure file.
    * Passport-jwt middleware used for token verification. jsonwebtoken is used used to sign and verify the token. 
* Split northbound and southbound api 
    * Multiple http/https endpoints (defined in config file) design see [Separate North/South API](https://groups.google.com/forum/#!topic/rackhd/n18-mWelhVs).
        * Refer to [monorail.json](https://github.com/WangWinson/RackHD/commit/0c37f1bc4932cd9c5f91c954ec4324e08cd8a48f) for configure file change.
    * Separate API routers to northbound and southbound. Leverage 2 endpoints to host north/south API. 

### This PR should be merged together with https://github.com/RackHD/RackHD/pull/83


zyoung51:
Can you rebase this?  It needs to be updated against the 1.1/2.0 changes that were already merged.

yyscamper:
All hound CI warnings except those like "xxx is not defined." should be fixed.

app.js
```diff
@@ -49,18 +48,45 @@ function Runner(
                 return skuPack.start(configuration.get('skuPackRoot', './skupack.d'));
             })
             .then(function() {
-                return app.listen();
+                // TODO: The following endpoint definition is for DEBUG using
+                // It will fianlly been moved to monorail.json config file
+                // var endpoints = configuration.get('httpEndpoints', [{port:8080}]);
```
benbp: I think many users are already using monorail.json for custom HTTP endpoints, and ignoring the config file will cause problems. Perhaps you can do this TODO for this PR?

benbp: Or, at the very least, use `configuration.get` in the endpoints definition, for example:

``` 
{
    "address": configuration.get("apiServerAddress", "172.31.128.1",
    "port": configuration.get("apiServerPort", "9080",
    // etc.
}
```

app.js
```diff
@@ -49,18 +48,45 @@ function Runner(
...
     function stop() {
-        return Promise.resolve()
-            .then(function() {
-                return app.close();
-            })
-            .then(function() {
-                return core.stop();
-            });
+        return Promise.map(services, function(service) {
+             service.stop();
```
benbp: This needs to be returned otherwise `Promise.map` won't wait on the resolution of `service.stop()`

lib/services/auth/auth-service.js
```diff
@@ -0,0 +1,135 @@
...
+                'zlxkgxjvcFwm0M8sWaGojh25qNYO8tuNWUMN4xKPH93PidwkCAvaX2JItLA3p7BSCWIzkw4GwWuezoMvKf3UXg=='),
+            this.encoder);
+
+        if (this.hashFromConf.length !== this.hashConfig.hashBytes ||
+            this.saltFromConf.length !== this.hashConfig.saltBytes){
+            throw new Error('Invalid password hash or salt length in configure file');
+        }
+
+        this.expiresIn = configuration.get('authTokenExpireIn', 86400); //86400 = 24 * 60 * 60, 24hours
+        if (typeof (this.expiresIn) !== 'number'){
```
benbp: FYI, you can use the assert library: `assert.number(this.expiresIn, "Invalid token expire value in configure file")`

lib/services/auth/auth-service.js
```diff
@@ -0,0 +1,135 @@
...
+);
+function authServiceFactory(
+    configuration
+) {
+    function AuthService() {
+    }
+
+    AuthService.prototype.init = function(){
+        this.algorithm = 'HS256';
+        this.secretOrKey = 'RackHDRocks!';
```
benbp: Is there any way to override this (perhaps I'm not understanding the code)?

benbp:
In general, things look pretty great, though -1 from me in regards to the configuration endpoint values, if I'm reading that code correctly.

lib/services/auth/auth-service.js
cgx027: This is actually hard coded here. Are you suggesting that user may want to change it to some other secret and we got need to offer them a easier way to do that? In that case, I would suggest to add a item in monorail.json called tokenSecret.

lib/services/auth/auth-service.js
cgx027: Good suggestion. Taken.

app.js
WangWinson: Yes, I have removed the local definition of `var endpoints`, it will get from `monorail.json`

app.js
WangWinson: Good catch! updated

cgx027:
Thanks @zyoung51, it's now up to date with master.

lib/api/1.1/northbound/authentication.js
```diff
@@ -0,0 +1,64 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: Copyright 2016

lib/api/1.1/northbound/login.js
```diff
@@ -0,0 +1,75 @@
...
+     * @apiSuccess {json} a token to be used when accessing other apis:
+     *      {
+     *       "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyIjoiY
+     *       WRtaW4iLCJpYXQiOjE0NTI0ODA3MTIsImV4cCI6MTQ1MjU2NzExMn0.Qo-9y
+     *       omlkd0WE1jlEA-L1WqA6k_9qJYhkicQL5IUmSg"
+     *      }
+     */
+
+    router.post('/login',
+        function(req, res, next) {          //Middleware to do authentication.
```
yyscamper: Since this anonymous callback function is a little complex, but give the callback a name will help debugging, refer to this link: [You Don't Know JS: Scope & Closures - Anonymous vs. Named] (https://github.com/getify/You-Dont-Know-JS/blob/master/scope%20&%20closures/ch3.md#anonymous-vs-named)

lib/api/1.1/northbound/login.js
```diff
@@ -0,0 +1,75 @@
...
+            passport.authenticate('local', {
+                    session: false
+                },
+                function(err, user, challenges){ // Callback to handle auth error
+                    if(err){
+                        res.status(ERROR_STATUS).send({message: 'Internal server error'});
+                    }
+                    else if (!user){
+                        if(challenges.message === 'Missing credentials'){
+                            res.status(BAD_REQUEST_STATUS).send(challenges);
```
yyscamper: Doesn't the `password.authenticate` returns some specified error code (type)? By directly mapping the error message to response code seems not a good idea.

lib/api/1.1/northbound/router.js
```diff
@@ -0,0 +1,57 @@
...
+        router.use(injector.get(require('./workflows')));
+        router.use(injector.get(require('./versions')));
+    };
+
+    router.unMountAllRouters = function(){
+        this.stack = [];
+    };
+
+    return router;
+}
```
yyscamper: a warning about newline

lib/api/1.1/northbound/router.js
```diff
@@ -0,0 +1,57 @@
...
+di.annotate(northApiFactory, new di.Inject(
+        di.Injector
+    )
+);
+
+function northApiFactory (injector) {
+    var router = express.Router();
+    router.endpoint = {};
+
+    router.mount = function() {
```
yyscamper: Bind the function to express Router instance may bring conflict in future if new version express Rounter designs a function with same name.

lib/api/1.1/northbound/templates.js
```diff
@@ -0,0 +1,87 @@
...
+    parser = require('body-parser');
+
+module.exports = templatesRouterFactory;
+
+di.annotate(templatesRouterFactory, new di.Provide('Http.Api.Templates'));
+di.annotate(templatesRouterFactory,
+    new di.Inject(
+		'Templates',
+		'common-api-presenter'
+	)
```
yyscamper: indent error

lib/api/1.1/northbound/templates.js
```diff
@@ -0,0 +1,87 @@
...
+     * @apiError Error problem was encountered, template was not written.
+     * @apiErrorExample Error-Response:
+     *     HTTP/1.1 500 Error
+     *     {
+     *       "error": "upload failed."
+     *     }
+     */
+
+    router.put('/templates/library/:identifier', parser.raw(),
+               function (req, res) {
```
yyscamper: indent error

lib/api/1.1/northbound/router.js
```diff
@@ -0,0 +1,57 @@
...
+        router.use(injector.get(require('./profiles')));
+        router.use(injector.get(require('./schemas')));
+        router.use(injector.get(require('./skus')));
+        router.use(injector.get(require('./templates')));
+        router.use(injector.get(require('./workflows')));
+        router.use(injector.get(require('./versions')));
+    };
+
+    router.unMountAllRouters = function(){
+        this.stack = [];
```
yyscamper: Could you please add a comment for this statement? seems a little tricky here as no elsewhere references the property `stack`

brianparry:
:-1: 
We'll discuss tomorrow morning, but my general comments:
* I think we need to discuss the merits of splitting northbound and southbound.  I'm not clear on the value that this adds.
* If we decide that the northbound/southbound split makes sense, it should apply to the 2.0 API only.  This should be much easier to do with Swagger than with express routers.
* My suggestion is to reduce this change to authentication only and then change lib/api/index.js to apply authentication middleware to the routes that need it.

lib/api/1.1/southbound/router.js
```diff
@@ -0,0 +1,37 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: This `southbound/router.js` is very alike with the `nothbound/router.js` , the only difference should be router.use. If in future we have v2 api, we need create more this kind router. So whether we can create some common router factory and generate different routers by parameter?

lib/api/1.1/northbound/router.js
```diff
@@ -0,0 +1,57 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+var di = require('di'),
+    express = require('express');
+
+module.exports = northApiFactory;
+
+di.annotate(northApiFactory, new di.Provide('northbound-api-router'));
```
yyscamper: I see you put this file under folder `/api/1.1` , if we have both 1.1 and 2.0 API coexists in future, this injectable name will have conflict, so then we may have to introduce the version number in the injectable name.

jlongever:
-1 
I have some concerns with the amount of changes this PR makes.   My understanding of the task was to add authentication to the v1.1 routes.  What is the value in splitting the APIs between northbound and southbound and creating more ports?  From a user perspective now I have to manage multiple ports I send on depending on which API I want to access?   How does a user access templates from the northbound side (the ISO expansion template for example)?   Could we just limit the scope of this PR to only adding authentication to selective routes?

WangWinson:
@brianparry @jlongever 
Per previous discussion, I think we agreed on split northbound southbound APIs. 
My understanding, some APIs (southbound) is for internally using only, for example, GET /tasks/bootstrap.js will only be call from a computer node side. We don't want to expose these kind of APIs to end user side.  If we don't separate the endpoints to host southbound/northbound API, how can we limit API access from specific path ? In our case, we know IP 172.xxxxx is from the computer node side. 
@anhou may have more details

lib/services/http-service.js
```diff
@@ -190,90 +181,113 @@ function httpServiceFactory(
...
+            }
+        }
 
-            wss.start(httpsServer, {
-                secureOptions: httpsOptions
-            });
+        return {
+            address: endpoint.address || '0.0.0.0',
+            port: endpoint.port || (endpoint.httpsEnabled ? 443 : 80),
+            httpsEnabled : endpoint.httpsEnabled === true,
```
yyscamper: The "===  true" can be omitted

package.json
```diff
@@ -35,7 +35,11 @@
     "swagger-express-mw": "^0.6.0",
     "os-tmpdir": "~1.0.1",
     "rimraf": "~2.5.0",
-    "tar": "~2.2.1"
+    "tar": "~2.2.1",
+    "jsonwebtoken": "^5.4.4",
+    "passport": "^0.3.2",
+    "passport-local": "^1.0.0",
+    "passport-jwt": "^1.2.1"
```
yyscamper: Not a formal convention, but previously, we keep writing the module name by alphabetical order. So change this or not is OK for me.

lib/api/1.1/southbound/router.js
WangWinson: Yes, good suggestion.
Unfortunately, v2 API will adopt `swagger yaml` to replace routers in express style. So, I think we could save this effort

lib/api/1.1/northbound/router.js
WangWinson: Same as above reply

lib/services/http-service.js
WangWinson: Yes, that will also work. But think about illegal input `httpsEnabled: 'Off'` in config file.

lib/api/1.1/northbound/templates.js
WangWinson: Quick eye !  even quicker then hound !
Just a copy pasted here. Anyway, will update

lib/api/1.1/northbound/router.js
WangWinson: Maybe encapsulate the router to a class instead of directly return a express Router ?

lib/api/1.1/northbound/templates.js
yyscamper: Hound will not report indent error, haha

yyscamper:
You just did revolution for the on-http!!!
This PR should also break the RackHD records: most code change, most houndci false alarm, most file change, most hard PR I have reviewed :) 

Since @jlongever @brianparry has some basic design concern, so you first need to reach consensus.

lib/services/http-service.js
yyscamper: shouldn't throw an error if `httpsEnabled` is not boolean? Just have a concern that translate both  `httpsEnabled: 1" or "httpsEnabled: 0" to false may bring confuse..

lib/api/1.1/northbound/router.js
yyscamper: Yes, depends on the effort. I guess add an underscore to the name should also be safe.

lib/api/1.1/northbound/authentication.js
cgx027: good catch

lib/api/1.1/northbound/login.js
cgx027: sounds good~

lib/api/1.1/northbound/router.js
cgx027: Agreed. New comments will be added to the code:
// Router.stack is an array that will grow every time a new
// route is mounted.
// Setting router.stack back to empty array will clear all
// that had previously mounted.

anhou:
@brianparry @jlongever In addition based on @WangWinson 's comments, for splitting southbound and northbound during authentication  is discussed in google group https://groups.google.com/forum/#!topic/rackhd/PmP2hosgJr8  the topic 'v1.1 API authentication --split southbound and northbound APIs'. it also have a full southbound and northbound API list.

anhou:
@WangWinson @cgx027 many file modes is changed to 755, it doesn't make sense, please keep original file mode 644 or 664

zyoung51:
-1, functional tests not pasing.  By semver rules, this is an API breaking change and should be bumped to 2.0?

yyscamper:
houndci warning should due to my previous change for jshintrc, so please ignore houndci warning at present.

WangWinson:
test this please

WangWinson:
test this please

WangWinson:
updated to fix Jenkins issue:
- Replace 8080 port with 8089 to fix port conflict issue during unit test in Jenkins
- Set default http service port to 8080 until `monorail.json` merged

lib/api/1.1/northbound/login.js
brianparry: @yyscamper passport.authenticate returns a connect/express middleware function.  In this case err indicates a problem with the authentication process, not failed authentication so generically mapping this to 500 status code seems reasonable to me.  I think these two conditions can be combined into one:

                       else if(challenges.message === 'Invalid username or password') {
                             res.status(UNAUTHORIZED_STATUS).send(challenges);
                        }
                        else {
                            res.status(UNAUTHORIZED_STATUS).send(challenges);
                       }

lib/api/1.1/northbound/authentication.js
```diff
@@ -0,0 +1,66 @@
...
+
+module.exports = authRouterFactory;
+
+di.annotate(authRouterFactory, new di.Provide('Http.Api.Auth'));
+di.annotate(authRouterFactory, new di.Inject(
+        'Auth.Services'
+        )
+);
+
+function authRouterFactory (
```
brianparry: I'd suggest replacing this with an `authMiddlewareFactory' that returns an express middleware function.  Then inject this function in the router factories that need it.  See https://github.com/RackHD/on-http/compare/master...brianparry:feature/swagger-auth#diff-87f6ea9f0a13097e18945b4d56c26dd5R1.

lib/services/auth-service.js
```diff
@@ -0,0 +1,149 @@
...
+module.exports = authServiceFactory;
+di.annotate(authServiceFactory, new di.Provide('Auth.Services'));
+di.annotate(authServiceFactory,
+    new di.Inject(
+        'Assert',
+        'Services.Configuration'
+    )
+);
+
+function authServiceFactory(
```
brianparry: @cgx027 I think we should look at https://github.com/ncb000gt/node.bcrypt.js for password hashing and verification.

lib/api/1.1/northbound/router.js
```diff
@@ -0,0 +1,61 @@
...
+
+function northApiFactory (injector) {
+    var router = express.Router();
+    router.endpoint = {};
+
+    router.mount = function() {
+
+        // Un-mount existing routes in case router.mount has
+        // been call multiple times.
+        this.unMountAllRouters();
```
brianparry: @cgx027 Why not just do this: `if (this.stack.length > 0) return;`

lib/services/http-service.js
```diff
@@ -62,94 +62,82 @@ di.annotate(httpServiceFactory,
...
+    HttpService.prototype.stop = function() {
+        wss.stop();
+        return this.server.close();
+    };
 
+    HttpService.prototype._setup =  function() {
+        var app = this.app;
+        var endpoint = this.endpoint;
         //Swagger support
         var swaggerConfig = {
```
brianparry: @WangWinson The `_setup` method always uses the same Swagger config.  If you'd like to use different yaml files for northbound and southbound, the swagger file path will need to be added to the endpoint object.

brianparry:
@cgx027 In general, I think the authentication parts are okay.  The only issues I think need to be addressed are:

1. The way passport middleware installed will make it difficult to share between 1.1 and 2.0
2. I think we are doing too much cryptography by hand.  I'd be much more comfortable using a library like https://github.com/ncb000gt/node.bcrypt.js.

@WangWinson I could give this a +0 if the 1.1 appears exactly the same as before by default and all of the functional tests pass.  I still think that we do not need the northbound and southbound router factories and that we should split NB and SB in 2.0 only.  Much of this will be removed or refactored later when the 1.1 API is deprecated.  The plan is to have another core committer meeting to discuss and try to come to consensus.

lib/services/http-service.js
WangWinson: Exactly, your suggestion above was done in my dome code https://github.com/WangWinson/on-http/commit/2f5289d50919b9c0aec61861689baf35e2237bd3 showed on last Wednesday's sync meeting, to support both 1.1 and 2.0 API with split NB/SB endpoints. 

If you think that's worth to be included in this PR, I'll add it.

WangWinson:
Thanks for your comments @brianparry. Looks like we are moving forward. At least +0 > -1. 
Split NB/SB or not in 1.1 API should align with the original 1.1 auth requirement. We discussed the split NB/SB topic a lot in google group. And I believe we reached agreement that we should split them. Honestly, split or not , I'm OK with both way. So code can be updated as long as decision made that split NB/SB is not the requirement for 1.1 API auth.  But still, I'll keep the service endpoints design part.

lib/api/1.1/northbound/router.js
cgx027: Thanks Brian for the comments. The intention here is to unmount everything that has already been mounted, when mount is called. The user case is that in the unit test, we would want to change a configure file (ie, authEnabled = True change to authEnabled = false) and then restart the http-service. The routers(ie, /login) that mounted when authEnable=True should be unmounted after restarting the http-service. 

I would take the action to update the notes to reflect my intention here.

lib/services/auth-service.js
cgx027: Nice suggestion @brianparry. 

Bcrypt is not bad either. Both of them can meets our needs. We did compare bcrypt and crypto.pbkdf2 before we start coding and crypo.pbkdf2 was the choice at that time. 

The discussion is listed in google group discussion at https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/rackhd/Hw88shMffLw

lib/api/1.1/northbound/authentication.js
cgx027: @brianparry .This is a good one. I had updated the code to expose the it as a middleware called authMiddlewareJwt, and moved it to auth-Service.js so that it can be shared. Refer to auth-Service.js line 190.

http-service.js line 186 is an example on how to make use of the middleware. Pretty much similar to what you show on your demo.

lib/services/http-service.js
brianparry: Okay.  I wasn't able to view the demo because my Skype connection dropped.  I do think this should be part of the PR.

lib/services/auth-service.js
brianparry: @cgx027 Could you elaborate on why you chose crypto for this?  As far as I can tell, its only advantage is that it's faster which shouldn't matter in this application since the password hashing is only used during login.  The justification mentioned in google groups seems to be just to avoid bringing in a new module.

tldavies:
test this please

tldavies:
test this please

lib/api/login.js
```diff
@@ -0,0 +1,42 @@
...
+     * @apiParam {String} password password used to login
+     * @apiSuccess {json} a token to be used when accessing other apis:
+     *      {
+     *       "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyIjoiY
+     *       WRtaW4iLCJpYXQiOjE0NTI0ODA3MTIsImV4cCI6MTQ1MjU2NzExMn0.Qo-9y
+     *       omlkd0WE1jlEA-L1WqA6k_9qJYhkicQL5IUmSg"
+     *      }
+     */
+
+    router.all('/login', authService.authMiddlewareLogin.bind(authService));
```
brianparry: @cgx027 I think this should be `router.post` instead of `router.all`

benbp:
:+1:

lib/services/auth-service.js
brianparry: @cgx027 @anhou As we discussed, I think we can think about this some more and make a decision later.  I'm fine with the change as is for now.

lib/services/http-service.js
brianparry: @WangWinson Could you update the diff with this change?  I think we should use the same approach as with V1: add a `swagger` property to each endpoint to configure which swagger file to use, but for now configure only a single endpoint with a single swagger file.

lib/services/auth-service.js
cgx027: Thanks Brian. I will mark it as a refactor task in PivotalTracker.

lib/api/login.js
cgx027: Good catch @brianparry . Since passport-local support auth with credentials in body or in query string. I would modify this to a post(for credentials in body) and a get(for credentials in query string).

lib/services/http-service.js
WangWinson: Of course. 
Lot's of conflict when rebasing code. Will add this update as soon as rebase successful

yyscamper:
+1

WangWinson:
test this please

cgx027:
test this please

lib/services/http-service.js
WangWinson: @brianparry the `yamlName` support in config has been added in latest update.

WangWinson:
Rebased to latest master, and addressed Brian's comments

lib/services/http-service.js
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016, EMC, Inc.
+// Copyright 2015, EMC, Inc.
```
zyoung51: Should this stay 2016?

lib/services/http-service.js
```diff
@@ -62,93 +63,81 @@ di.annotate(httpServiceFactory,
...
 
+    HttpService.prototype.createSwagger = function() {
+        var app = this.app;
+        var endpoint = this.endpoint;
         //Swagger support
         var swaggerConfig = {
             appRoot: path.resolve(__dirname, '../../'),
             configDir: path.resolve(__dirname, '../../swagger_config'),
-            swagger: path.resolve(__dirname, '../../static/monorail-2.0.yaml'),
+            swagger: path.resolve(__dirname, '../../static/' + endpoint.yamlName),
```
zyoung51: Why is this name configurable but the redfish swagger name isn't?

lib/services/http-service.js
```diff
@@ -201,91 +202,125 @@ function httpServiceFactory(
...
+        }
 
-            wss.start(httpsServer, {
-                secureOptions: httpsOptions
-            });
+        return {
+            address: endpoint.address || '0.0.0.0',
+            port: endpoint.port || (endpoint.httpsEnabled ? 443 : 80),
+            httpsEnabled : endpoint.httpsEnabled === true,
+            httpsCert: endpoint.httpsCert || 'data/dev-cert.pem',
```
zyoung51: Should we check to confirm the files exist?

package.json
```diff
@@ -4,11 +4,11 @@
...
   "main": "index.js",
   "scripts": {
-    "doc": "./node_modules/.bin/jsdoc lib -r -d docs",
+    "doc": "./node_modules/.bin/jsdoc lib/api/1.1/northbound -r -d docs",
     "test": "./node_modules/.bin/mocha $(find spec -name '*-spec.js') -R spec --require spec/helper.js",
     "install": "./install-web-ui.sh",
-    "apidoc": "./node_modules/.bin/apidoc -i lib -f '.*\\.js$' -o build/apidoc",
-    "apidoc-swagger": "./node_modules/.bin/apidoc-swagger --verbose -i lib -f '.*\\.js$' -o build/apidoc-swagger"
+    "apidoc": "./node_modules/.bin/apidoc -i lib/api/1.1/northbound -f '.*\\.js$' -o build/apidoc",
+    "apidoc-swagger": "./node_modules/.bin/apidoc-swagger --verbose -i lib/api/1.1/northbound -f '.*\\.js$' -o build/apidoc-swagger"
```
zyoung51: Is the documentation for southbound removed?  I don't quite understand what this section is doing.

package.json
```diff
@@ -25,19 +25,26 @@
...
-    "pluralize": "~1.1.2",
-    "ws": "^0.8.0",
-    "swagger-express-mw": "^0.6.0",
     "os-tmpdir": "~1.0.1",
     "rimraf": "~2.5.0",
     "tar": "~2.2.1",
     "tv4": "~1.2.7",
-    "swagger-tools": "~0.9.11"
+    "swagger-tools": "~0.9.11",
+    "jsonwebtoken": "^5.4.4",
```
zyoung51: Duplicate of line 28

lib/api/login.js
```diff
@@ -0,0 +1,60 @@
...
+     * @apiParam {query} password password used to login
+     * @apiSuccess {json} a token to be used when accessing other apis:
+     *      {
+     *       "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyIjoiY
+     *       WRtaW4iLCJpYXQiOjE0NTI0ODA3MTIsImV4cCI6MTQ1MjU2NzExMn0.Qo-9y
+     *       omlkd0WE1jlEA-L1WqA6k_9qJYhkicQL5IUmSg"
+     *      }
+     */
+
+    router.get('/login', authService.authMiddlewareLogin.bind(authService));
```
brianparry: @cgx027 I don't understand why a GET handler is needed.  The semantics for the GET method ignore the request body, so the only way this can work is if username and password are provided as a query string in plain text.

spec/lib/api/login-spec.js
```diff
@@ -0,0 +1,377 @@
...
+                "port": 8089,
+                "httpsEnabled": false,
+                "authEnabled": true,
+                "routers": "northbound-api-router"
+            };
+            startServer(endpointHttp);
+        });
+
+        //give a shoot on http instead of https.
+        it('should success auth with http instead of https', function() {
```
zyoung51: Do we want to support the authenticate over insecure case?

spec/lib/api/login-spec.js
```diff
@@ -0,0 +1,377 @@
...
+        });
+    });
+
+    describe('test with authentication disabled', function () {
+        before('start HTTPs server', function () {
+            this.timeout(5000);
+            endpoint.authEnabled = false;
+            startServer(endpoint);
+            //restore endpoint
+            endpoint.authEnabled = true;
```
zyoung51: Do this in 'after' ?

lib/api/login.js
cgx027: Hi @brianparry , it's true that when you are sending the query using GET, a request body is not needed, instead, you have to provide the username and password in query string, as plain text, as you noticed. I guess you are concerning about security of having username/password in plain text but SSL by nature will secure the URL thus a man-in-the-middle attacker will not get to know the URL using a sniffer. Refer to [How secure are query string over https](https://blog.httpwatch.com/2009/02/20/how-secure-are-query-strings-over-https/). Please let me know if this does not address your concern. 

The reason we keep GET is that someone may use a plain web-browser (not using any browser plugins) to testify our APIs, in which case a POST body is not feasible. This seems to be a user case discussion thus I include @leoyjchang and @heckj for a comment here.

spec/lib/api/login-spec.js
cgx027: @zyoung51 we do not recommend user to do that but the code did not restrict them from doing it. This test case is to make sure we will not crash even user choose to do auth over http.

lib/services/http-service.js
WangWinson: Well, this is just update to this file. It was originally created on 2015. I guess the copyright should reflect the first time the file created ?

lib/services/http-service.js
WangWinson: Of course! I could do the same way if you think it's good. I made this 2.0 swagger filename configurable as Brain required.

lib/api/login.js
jlongever: @cgx027 that blog seems to infer that its best to not pass sensitive data as a query string. It suggests using a token ID or session cookies.  IMHO we should not be passing user/password info in a query string, writing user/password credentials to a log, or allow authentication over HTTP on external network (denied by firewall).

package.json
WangWinson: Since the southbound well not expose to user, it may confuse user that why some APIs listed in the docs could not be accessed. This update remove the southbound APIs from docs.
@anhou correct me.

package.json
WangWinson: Ooops~ Good catch.

lib/api/login.js
cgx027: @jlongever I agree GET can be more insecure if care is not taken. It seems a fine trade off for people not able to test the API using a plain browser. I will submit a patch to remove GET.

spec/lib/api/login-spec.js
cgx027: Thanks @zyoung51 Good suggestion.

lib/services/http-service.js
cgx027: From what I captured in Slack discussion, only the newly added files need to update to copyright 2016. @benbp @yyscamper @anhou should know better.

spec/lib/api/login-spec.js
zyoung51: Where is that recommendation documented?

package.json
zyoung51: How are developers supposed to access the southbound api and use it? The documentation is needed...

lib/services/http-service.js
zyoung51: I can do that later, was just wondering why only one of them was configurable. Thanks.

lib/services/http-service.js
WangWinson: Good suggestion. I can add the fs.exists logic here.

spec/lib/api/login-spec.js
cgx027: It's not yet documented but I believe the best place is here: [configuration.rst](https://github.com/RackHD/docs/blob/master/docs/rackhd/configuration.rst). This will be a pull request later on.

lib/services/http-service.js
yyscamper: Yes, only for newly created file.

PR #97, opened 2016-01-23
Add module path for di injector

iceiilin:
Previous calls of helper.requireWrapper() didn't designate required module path, so the default ones are the path of on-core/di.js and its parent directories. Modules can be successfully injected when on-core is a real module under RackHD/on-http/node_modules/on-core, but will throw error about failure to find modules when using "npm link" in on-http to create a symbolic link of on-core to the local source code, since in the latter case, the working directory is RackHD/on-core/lib/di.js, and nodejs cannot find  modules in RackHD/on-http/node_modules.
So __dirname is added as a parameter in helper.requireWrapper() to ask di.js to find modules in current directory in on-http.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,7 +19,6 @@ function onHttpContextFactory(di, directory) {
 
     var core = onCore(di, directory),
         helper = core.helper;
-
     return {
         expressApp: function () {
             return helper.simpleWrapper(express(), 'express-app', undefined, __dirname);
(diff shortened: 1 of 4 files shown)


yyscamper:
:+1:

benbp:
Thank you! I made these changes in my local branch is well to work around the issue. I didn't catch that it was because of npm link though, that explains how it passed tests.

:+1:

PR #104, opened 2016-01-28
Redfish 1.0 /TaskService:

zyoung51:
- Implement GET for the following endpoints:
  - /TaskService
  - /TaskService/Tasks
  - /TaskService/Oem/Tasks/{nodeid}
  - /TaskService/Tasks/{taskid}
- Update redfish swagger definition
- Add unittests


zyoung51:
@RackHD/corecommitters @keedya @tannoa2 @uppalk1

keedya:
:+1:

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+                });
+                return redfish.render('redfish.1.0.0.taskcollection.json', 
+                        'TaskCollection.json#/definitions/TaskCollection',
+                        options);
+            });
+    }).then(function(output) {
+        res.status(200).json(output);
+    })
+    .catch(function(error) {
+        res.status(500).json(error);
```
heckj: "teach me" question - what's the error data returned here with the 500? trying to understand what's exposed if/when we hit this sort of thing.

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+    options.identifier = identifier;
+
+    waterline.nodes.needByIdentifier(identifier)
+    .then(function() {
+        return Promise.all([ dataFactory('', 'graphObjects') ])
+            .spread(function(graphs) {
+                options.graphs = _.filter(graphs, function(graph) {
+                    return graph.systemId === identifier;
+                });
+                return redfish.render('redfish.1.0.0.taskcollection.json',
```
heckj: another dummy question - if the result is an empty set, does this end up returning a 404 response?

lib/api/redfish-1.0/task-service.js
zyoung51: Whatever 'error' was thrown is what comes out.  None of the response codes or data models are defined in the swagger file yet.

lib/api/redfish-1.0/task-service.js
zyoung51: No, it returns a 200 with a valid task collection object that contains 0 members in the tasks array.

lib/api/redfish-1.0/task-service.js
heckj: heh, ok. This is where I don't know the Redfish spec - is that what Redfish is supposed to do, or is this a gap and it should be a 404 response?

lib/api/redfish-1.0/task-service.js
heckj: heh, ok. This is where I don't know the Redfish spec - is that what Redfish is supposed to do, or is this a gap and it should be a 404 response?

lib/api/redfish-1.0/task-service.js
heckj: was hoping you could illuminate what was in an "error" object - don't know this code yet

zyoung51: A correct error object would contain fields for a code, message and an array of message objects with extended information:
``
{
"error": {
"code": "Base.1.0.GeneralError",
"message": "A general error has occurred. See ExtendedInfo for more information.",
"@Message.ExtendedInfo": []
}
}
``
This error object is not formatted and contains what was thrown by EJS, TV4, or another exception

jlongever:
+1

Not sure what Hound is doing... 

@RackHD/corecommitters is this OK to merge?

lib/api/redfish-1.0/task-service.js
zyoung51: Re-reading your question, I answered if the graph was an empty set, not if the nodeid was an empty set.  For a nodeid empty set it should return a 404 (the URI requested a resource that doesn't exist).  The swagger model and error handling work will be completed further out.

lib/api/redfish-1.0/task-service.js
heckj: Huh - interesting. Maybe not entirely useful, but not exposing guts or anything either.

lib/api/redfish-1.0/task-service.js
heckj: Cool, thanks!

heckj:
:+1:

spec/lib/api/redfish-1.0/task-service-spec.js
```diff
@@ -0,0 +1,158 @@
...
+        validator.render.reset();
+
+        function resetStubs(obj) {
+            _(obj).methods().forEach(function (method) {
+                if (obj[method] && obj[method].reset) {
+                  obj[method].reset();
+                }
+            }).value();
+        }
+
```
iceiilin: Will sandbox be used here to simplify the reset and restore process?

yyscamper:
Hi @zyoung51 : I expect move all Redfish template JSON file to a particular sub-folder under `templates`. Especially when more and more Redfish APIs are supported, there will be more and more such templates be created.

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+        'valid': 'Running',
+        'succeeded': 'Completed',
+        'cancelled': 'Killed',
+        'failed':  'Exception'
+    };
+
+    switch(dataName)  {
+        case 'graphObjects':
+            return Promise.map(waterline.graphobjects.find({}), function(graph) {
+                    return {
```
yyscamper: incorrect indent.

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+    Promise.all([ dataFactory('', 'graphObjects') ])
+        .spread(function(graphs) {
+            _.forEach(graphs, function(graph) {
+                if (graph.systemId && -1 === options.systems.indexOf(graph.systemId))  {
+                    options.systems.push(graph.systemId);
+                }
+            });
+            options.graphs = graphs;
+            return redfish.render('redfish.1.0.0.taskservice.1.0.0.json', 
+                 'TaskService.1.0.0.json#/definitions/TaskService',
```
yyscamper: Is their any better method to not hard code these template file name deep in the code? Thus will ease our future maintenance.
Such as creating a redfish-template-service.js, which exposes an API `render(resourceName, version)`, so for `TaskService`, just need to call `render('TaskService', '1.0.0')`

Just my proposal, you may have better idea.

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+                        systemId: graph.node
+                    };
+                });
+    }
+};
+
+function taskServiceRoot(req, res) {
+    var options = {};
+    options.basepath = req.swagger.operation.api.basePath;
+    options.templateScope = ['global'];
```
yyscamper: I wonder how the `templateScope` is used, I don't find any where to reference it.

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+    options.url = req.url;
+    options.dateTime = moment.utc().format();
+    options.systems = [];
+
+    Promise.all([ dataFactory('', 'graphObjects') ])
+        .spread(function(graphs) {
```
yyscamper: whether this should be `then`? `spread` will flat all array elements to multi arguments, like `.spread(function(graph1, graph2, graph3, ...)`

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+    });
+}
+
+function listTasks(req, res) {
+    var options = {};
+    options.basepath = req.swagger.operation.api.basePath;
+    options.templateScope = ['global'];
+    options.url = req.url;
+    options.dateTime = moment.utc().format();
+
```
yyscamper: I see these `options.basepath`, `options.templateScope`, `options.url`, `options.dateTime` are exactly  the same for every function, so wrap them to a common function will be better.

package.json
```diff
@@ -37,7 +37,8 @@
     "rimraf": "~2.5.0",
     "tar": "~2.2.1",
     "tv4": "~1.2.7",
-    "swagger-tools": "~0.9.11"
+    "swagger-tools": "~0.9.11",
+    "moment": "~2.11.1"
```
yyscamper: If could, please write these module name in alphabet order, because our previous modules are all defined in such a convention. Of course, if you use following command to add the module, it will be automatically added by alphabetic order:
```
npm install swagger-tools --save
```

spec/lib/api/redfish-1.0/task-service-spec.js
```diff
@@ -0,0 +1,158 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: Copyright 2016

spec/lib/api/redfish-1.0/task-service-spec.js
yyscamper: Yes, sinon.sandbox is a good tool here to manage multi stub and spy.

spec/lib/api/redfish-1.0/task-service-spec.js
```diff
@@ -0,0 +1,158 @@
...
+        node: 'abcdefg'
+    };
+
+    it('should return a valid task service root', function () {
+        waterline.graphobjects.find.resolves([graph]);
+        return helper.request().get('/redfish/v1/TaskService')
+            .expect('Content-Type', /^application\/json/)
+            .expect(200)
+            .expect(function(res) {
+                expect(tv4.validate.called).to.be.true;
```
yyscamper: I think the better writing should be `expect(tv4.validate).to.have.been.called`

spec/lib/api/redfish-1.0/task-service-spec.js
```diff
@@ -0,0 +1,158 @@
...
+            .expect(500);
+    });
+
+    it('should return a valid task from a system', function () {
+        waterline.graphobjects.find.resolves([graph]);
+        waterline.nodes.needByIdentifier.resolves();
+        return helper.request().get('/redfish/v1/TaskService/Oem/Tasks/' + graph.node)
+            .expect('Content-Type', /^application\/json/)
+            .expect(200)
+            .expect(function(res) {
```
yyscamper: You can explicitly use jshint ignore to make the houndci happy
```
.expect(function(res) { //jshint ignore: true
```

yyscamper:
@jlongever : houndci's warning is positive, The code defines a variable `res` but it never been used within its scope. In case, to make the houndci happy, we can explicitly tell jshint to ignore such warning, jut like:
```
.then(function(res) { //jshint ignore: true
```

jlongever:
@yyscamper sub-directories in /templates isn't supported today.  
re: hound,  my question was regarding why it was stuck on 'in-progress' status.  Seems fine after the updated push.

lib/api/redfish-1.0/task-service.js
zyoung51: For example, Redfish has both a 'TaskService' and 'TaskService.1.0.0' with different schema definitions.  Additionally, I would prefer to be able to declare the definition to template relationship within the implementor to avoid creating an inflexible hardcoded mapping of relationships somewhere else.

lib/api/redfish-1.0/task-service.js
zyoung51: Great point about the options.  @jlongever and I have discussed a plan to refactor the options generation into a factory.  The templateScope is used when retrieving the templates in the render() function as part of the SKU package.

lib/api/redfish-1.0/task-service.js
zyoung51: I want the array constructed above to be flattened by the spread.  I used an array for consistency with all the other Redfish implementations.  Sometime later, all the code could be refactored as the pattern is the same in all of them.

package.json
zyoung51: You've made this comment a few times, but this is the exact command I am using, and that is what the command did.  

Note that all of these packages are not alphabetized:
    "swagger-express-mw": "^0.6.0",
    "os-tmpdir": "~1.0.1",
    "rimraf": "~2.5.0",
    "tar": "~2.2.1",
    "tv4": "~1.2.7",
    "swagger-tools": "~0.9.11",
    "moment": "~2.11.1"
and:
   "nock": "~3.5.0"

zyoung51:
@yyscamper Previous PR to fix directories in templates is here: https://github.com/RackHD/on-core/pull/42.  While this fixed the code from crashing when a directory is in templates, it did not add recursion.  @RackHD/corecommitters should the directories be traversed now?

jlongever:
@zyoung51 Supporting recursive directories for static files seems useful.

yyscamper:
@jlongever @zyoung51 : sure the templates/profiles don't support recursive directories at present, but we can enable it (open another PR should be better).

package.json
yyscamper: I just did experiment on this, I also see these package names are not alphabetized,  I remembered it should work in my previous testing, I see some people said about this: http://stackoverflow.com/questions/26934886/how-to-prevent-npm-install-package-save-dev-from-reordering-devdependencies

@zyoung51 : please ignore this comment, maybe my testing has something wrong. Alphabetical order or not will not impact function, so I'm OK with either way.

lib/api/redfish-1.0/task-service.js
yyscamper: What's the relationship between `TaskService` and `TaskService 1.0.0`? 

I think create another lookup function will be more flexible, for example, in future, porting your code to Redfish 2.x will not need to change the hard-coded version string everywhere; another example, if we move those templates file to sub-directory, we just need to add the sub-directory name in the lookup function, don't need to change every template path.

yyscamper:
@jlongever Sometime I can also see the houndci stuck in progress, trigger it again may resolve, I don't know why. If we see it again, we can report a bug for houndci.

lib/api/redfish-1.0/task-service.js
zyoung51: This doesn't reference a file path, it references database name field. I understand the directory setup will make it easier for you to look at the files, but it has no bearing on the 'name' field in the database unless you are advocating that they become coupled? Or that we stop storing the templates in the database and start reading from the filesystem?

WangWinson: @zyoung51 *'Or that we stop storing the templates in the database and start reading from the filesystem?'*  This is a good idea. Actually, do not save file `data` in database, just save a link/path/meta of the file should be good. We may need to track this for later discussion.

zyoung51: This is probably not a good idea. The database will be clustered/shared and accessed from (potentially) many rackhd instances. Isolating this information on the filesystem would probably be a step backwards in the architecture. With that said, if the core committees want to move this out of the database and into a filesystem because my template names are not to @yyscamper liking, then I'm up for trying it out

yyscamper: I think @WangWinson 's concern here is if we move it from database to filesystem, it will be a way to reduce the RackHD footprint, I know some people is working on this. 

Meanwhile, put it into filesytem will get the dynamic loading feature, but our current database design only load templates one time when app runs, if we put a new template file (especially during debugging), we have to reboot the app to let the new template file takes effect.

Of course, this discussion is not push you to change it in this PR, first we need a discussion, if agree, we can do it in another PR.

lib/api/redfish-1.0/task-service.js
yyscamper: yes, `spread` is to flatten the arguments, but I see your function arguments is not written in flattened style, maybe `spread(function(graphs)` should be `then(function(graphs)` or `spread(function(graph1, graph2, graph3, ...)`?

lib/api/redfish-1.0/task-service.js
zyoung51: it is really a conversation for @RackHD/corecommitters.  Moving the contents of templates/profiles out of the database and into the filesystem will impact the entire design.  It seems like a huge change to be made because you don't like the names, with little benefit to be had.

The contents are sourced from the database and not cached.  Update the database and stop reloading the app.  The new template takes affect immediately.

yyscamper: I think two topic here:

(1) whether hard-coded such long but seems well-regular string in so many places is a good design? we have to do this or is there any better design?

(2) Since redfish api will introduce more and more templates, so whether we need change templates/profiles from database to filesystem. database is better for sharing across apps, filesystem is good for dynamic loading and footprint reduce, but filesystem need much code change.


Felix

-----Original Message-----

lib/api/redfish-1.0/task-service.js
```diff
@@ -0,0 +1,145 @@
...
+var moment = require('moment');
+
+module.exports = {
+    taskServiceRoot: taskServiceRoot,
+    listTasks: listTasks,
+    getTask: getTask,
+    getSystemTasks: getSystemTasks
+};
+
+var dataFactory = function(identifier, dataName) {
```
yyscamper: just find the argument `identifier` is not referenced anywhere, wondering why houndci not report a warning, and is the`identifier` should be the argument for the following `find` function?

zyoung51: This is to align with the other dataFactory functions in Chassis and System.  The intention has always been to refactor them into a real factory in a future commit.

amymullins:
At this point, any requested topics for additional discussion have been moved to google groups, this code review has met the merge criteria, and I request one of the @RackHD/corecommitters to make the merge so we can move on to the remaining development stories in the backlog.

amymullins:
Once the rebase occurs of course:)

PR #105, opened 2016-01-28
Extend redfish API for Chassis

jlongever:
- Add Chassis root and collections path
- Add Thermal reference object
- Add Power reference object
- Add unit-tests

@RackHD/corecommitters @zyoung51 @keedya @uppalk1 @tannoa2

### DIFF
--- a/lib/api/redfish-1.0/chassis.js
+++ b/lib/api/redfish-1.0/chassis.js
@@ -1,0 +1,226 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var redfish = injector.get('Http.Api.Services.Redfish');
+var _ = injector.get('_');
+var waterline = injector.get('Services.Waterline');
+var Promise = injector.get('Promise');
+var Errors = injector.get('Errors');
+var taskProtocol = injector.get('Protocol.Task');
+var basepath = '/redfish/v1';
+var Errors = injector.get('Errors');
+
+module.exports = {
+    listChassis: listChassis,
+    getChassis: getChassis,
+    getThermal: getThermal,
+    getPower: getPower
+};
+
+var dataFactory = function(identifier, dataName) {
+    switch(dataName)  {
+        case 'ohai':
+        case 'dmi':
+            return waterline.nodes.needByIdentifier(identifier)
+            .then(function(node) {
+                return waterline.catalogs.findLatestCatalogOfSource(node.id, dataName);
+            })
+            .then(function (catalogs) {
+                if (_.isEmpty(catalogs)) {
+                    throw new Errors.NotFoundError(
+                        'No Catalogs Found for Source (' + dataName + ').'
+                    );
+                }
+                return catalogs;
+            });
+        case 'system':
+            return waterline.nodes.needByIdentifier(identifier)
+            .then(function(node) {
+                return _.filter(node.relations, function(relation) {
+                    return relation.relationType === 'encloses';
+                }).map(function(relation) {
+                    return relation.targets[0];
+                });
+            });
+        case 'chassis':
+            return waterline.nodes.needByIdentifier(identifier)
+            .then(function(node) {
+                return _.filter(node.relations, function(relation) {
+                    return relation.relationType === 'encloses';
+                }).map(function(relation) {
+                    return relation.targets;
+                });
+            })
+            .then(function(targets) {
+                return targets[0];
+            });
+        case 'chassisData':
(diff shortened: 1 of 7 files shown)


jlongever:
^^ just pushed a fix for that jenkins failure, next build should pass.

jlongever:
test this please

static/redfish.yaml
```diff
@@ -3,7 +3,7 @@ info:
   version: "0.0.1"
   title: Redfish POC
 # during dev, should point to your local machine
-host: 10.240.19.38:8080
+host: localhost:8080
```
zyoung51: Good catch, thanks for fixing.

zyoung51:
+1

lib/api/redfish-1.0/chassis.js
```diff
@@ -0,0 +1,232 @@
...
+
+var dataFactory = function(identifier, dataName) {
+    switch(dataName)  {
+        case 'ohai':
+        case 'dmi':
+            return waterline.nodes.needByIdentifier(identifier)
+            .then(function(node) {
+                return waterline.catalogs.findLatestCatalogOfSource(
+                        node.id, dataName
+                    ).then(function (catalogs) {
```
benbp: Flatten out this promise chain please, so we can avoid pyramid of doom-iness.

lib/api/redfish-1.0/chassis.js
```diff
@@ -0,0 +1,232 @@
...
+                return redfish.render('redfish.1.0.0.chassis.1.0.0.json', 
+                                'Chassis.1.0.0.json#/definitions/Chassis',
+                                options);
+        }).then(function(output) {
+            res.status(200).json(output);
+        })
+        .catch(function(error) {
+            res.status(500).json(error);
+        });
+    });
```
benbp: There doesn't appear to be any catch handler for the data factory call to 'system', just the inner data factory calls. We should flatten this promise chain as well I believe so that we don't get any silent errors.

lib/api/redfish-1.0/chassis.js
```diff
@@ -0,0 +1,232 @@
...
+                options.fanList = fanList;
+                return redfish.render('redfish.1.0.0.thermal.1.0.0.json', 
+                                'Thermal.1.0.0.json#/definitions/Thermal',
+                                options);
+        }).then(function(output) {
+            res.status(200).json(output);
+        })
+        .catch(function(error) {
+            res.status(500).json(error);
+        });
```
benbp: Same as above re: missing catch handler.

lib/api/redfish-1.0/chassis.js
```diff
@@ -0,0 +1,232 @@
...
+                return redfish.render('redfish.1.0.0.power.1.0.0.json', 
+                                'Power.1.0.0.json#/definitions/Power',
+                                options);
+        }).then(function(output) {
+            res.status(200).json(output);
+        })
+        .catch(function(error) {
+            res.status(500).json(error);
+        });
+    });
```
benbp: Same as above.

lib/api/redfish-1.0/chassis.js
```diff
@@ -0,0 +1,232 @@
...
+}
+
+function getChassis(req, res) {
+    var identifier = req.swagger.params.identifier.value;
+    var options = {};
+    options.basepath = basepath;
+    options.templateScope = ['global'];
+    options.url = req.url;
+    options.identifier = identifier;
+    options.chassisType = "Other";
```
benbp: Style question: why not just define all this as an object literal?

```
var options = {
    url: req.url,
    basepath: basepath,
    < etc. >
}
```

benbp:
-1 until the promise chains are fixed, then +1

lib/api/redfish-1.0/chassis.js
jlongever: no special reason,  carried over from the initial redfish work.   I'll change it if you prefer an object literal definition.

jlongever:
@benbp updated to address comments.

benbp:
Thanks! :+1:

lib/api/redfish-1.0/chassis.js
zyoung51: @jlongever this is such a boilerplate thing, maybe we should generate it from a factory instead?

jlongever: @zyoung51 Works for me.

jlongever:
@RackHD/corecommitters is this OK to merge now?

PR #106, opened 2016-01-28
ON-2719 - Add additional POST workflow route necessary to

jimturnquist:
ON-2719 - Add additional POST workflow route necessary to support node refresh workflow.

@benbp

PR #107, opened 2016-01-28
ON-2719 - Add additional POST workflow route necessary to

jimturnquist:
support node refresh workflow.

(cherry picked from commit d4ccb4438f9e4a25d3f9caea42eadf6fb79dc5be)


benbp:
Thanks! I've actually got an addition/overhaul of this same route going up against master in the next week. So if you're okay with it, I'd like to recommend we only merge the PR up against 1.1 (https://github.com/RackHD/on-http/pull/106) so that we have a short term solution for your team, but not this one against master since it's all going to be updated anyways.

jimturnquist:
Sounds fine Ben.  We won't be actively testing from master in the next few weeks so we should be good.

benbp:
Ok, thanks.

PR #112, opened 2016-02-03
Taskgraph improvements

benbp:
Refactoring the workflow engine for:

- High Availability
- Fault Tolerance
- Better performance
- Better execution strategy

This PR primarily adds up front creation, validation and persistence of TaskGraph objects on workflow requests, instead of sending out requests via AMQP before. This prevents a lot of the failure cases and timeouts under load that we used to see, and gives the user more immediate feedback on failure. Also adding a workflow-api-service that is now called out to by many of the other APIs and services, reducing the direct dependency on some waterline collection methods.

Relies on https://github.com/RackHD/on-tasks/pull/116, RackHD/on-core#75, and RackHD/on-http#112

See the description of RackHD/on-taskgraph#56 for architectural slides.


benbp:
Mentioning a mix of those who came to the presentation/should review this: @RackHD/corecommitters @VulpesArtificem @heckj @king-jam @johren @stuart-stanley @jimturnquist @larry-dean @davequick @zyoung51 @zhangy42 @mattgam @tldavies @DavidjohnBlodgett @rolandpoulter

heckj:
needs a little rebase love :-)

benbp:
Yep, I merged the nodes-api-service PR yesterday which of course broke this. I'm worried if I rebase now I'll have to do it again though for some other reason, so I'm just waiting until the code contents are reviewed and ready to go, and then I'll plan on the rebase if that sounds good to you?

benbp:
It's the same situation with on-core.

lib/api/1.1/northbound/nodes.js
```diff
@@ -540,12 +546,14 @@ function nodesRouterFactory (
     router.get('/nodes/:identifier/workflows/active', rest(function (req) {
         return waterline.nodes.needByIdentifier(req.params.identifier)
         .then(function (node) {
-            return taskGraphProtocol.getActiveTaskGraph( { target: node.id })
-            .then(function (graph) {
-                if (graph) {
-                    return waterline.graphobjects.findOne({ instanceId: graph.instanceId });
-                }
-            });
+            return workflowApiService.findActiveGraphForTarget(node.id);
```
heckj: Nothing wrong here, but begged the question - if you leverage the new entrypoint to start a workflow that doesn't have a target set to a `node`/`node.id`, how do you query for it's state or otherwise lookup status like you do here for `/nodes/:id/workflows/active`?

lib/api/1.1/northbound/nodes.js
```diff
@@ -569,9 +577,12 @@ function nodesRouterFactory (
     router.delete('/nodes/:identifier/workflows/active', rest(function (req) {
         return waterline.nodes.needByIdentifier(req.params.identifier)
         .then(function (node) {
-            return taskGraphProtocol.cancelTaskGraph({ target: node.id });
+            return workflowApiService.findActiveGraphForTarget(node.id);
+        })
+        .then(function(graph) {
+            return workflowApiService.cancelTaskGraph(graph.instanceId);
```
heckj: if there is no active workflow found, does this `500` error or pass through with a successful cancel? (i.e. what does `workflowApiService.cancelTaskGraph` do when graph is empty? Will there be a graph.instanceId on that object no matter what, or is that a bad assumption here?

lib/api/1.1/northbound/nodes.js
benbp: @heckj it's not super obvious, but the `/workflows` route returns all active workflows, and also supports tacking on query parameters. I think it would be more clear as something like `/workflows/active` and `/workflows/active/:id`, but I was hesitant to make too many changes to existing api routes, though I'm happy to add some redundant routes if you want.

lib/api/1.1/southbound/profiles.js
```diff
@@ -47,16 +45,6 @@ function profilesRouterFactory (
                     });
                 }
             })
```
heckj: Why are we removing this back-off mechanism? Is it not effective/functional and now cruft (was replaced by something else), or just no longer an error condition the workflow engine will generate, so not an issue?

lib/api/1.1/northbound/nodes.js
heckj: I think we may want to consider that as follow-on work while we're pressing out more of the V2 API rather than adding lots of extensions under the 1.1 umbrella - @brianparry thoughts there? I normally balk at redundant routes, but in this case I think it would makes sense - either now (in the 1.1 API) or as an additional item under the v2.0 API set.

lib/api/1.1/northbound/nodes.js
benbp: Good question, I hadn't noticed this edge case. In the case that the graph finishes after the `findActiveGraphForTarget` call but before the `cancelTaskGraph` call, the user will get back a 204 No Content (just tested this out). Just brainstorming, perhaps we want to tack on another step that returns the instanceId no matter what on success, or do something else?

lib/api/1.1/northbound/nodes.js
benbp: Yep, I agree. I have a bunch of route change suggestions that I need to reach out to @brianparry about. Guess this is as good a time as any :)

spec/lib/api/login-spec.js
```diff
@@ -54,10 +54,12 @@ describe('Http.Api.Login', function () {
         return [
             dihelper.simpleWrapper(require('swagger-express-mw'), 'swagger'),
             dihelper.simpleWrapper(ws.Server, 'WebSocketServer'),
+            dihelper.simpleWrapper({}, 'TaskGraph.TaskGraph'),
+            dihelper.simpleWrapper({}, 'TaskGraph.Store'),
             dihelper.simpleWrapper({}, 'Task.Services.OBM'),
             dihelper.simpleWrapper({}, 'ipmi-obm-service'),
-            dihelper.requireWrapper('rimraf', 'rimraf'),
-            dihelper.requireWrapper('os-tmpdir', 'osTmpdir'),
+            dihelper.requireWrapper('rimraf', 'rimraf', undefined, __dirname),
```
heckj: Okay - call me clueless - why the additional `undefined` and `__dirname` passed into the DI system here?

lib/api/1.1/southbound/profiles.js
benbp: I considered a short term workaround that was quite brittle because it relies on AMQP events only to track workflow start/finished state and can quickly become inaccurate if any of those are missed. My hope is that with the performance gains we are getting from the refactor, plus the logging updates and other improvements we made since the introduction of this, that we won't need it anymore.

If we start to see this become an issue again, or anticipate right now that it will, we should add this back in, but I think we should re-design it to be more robust in its tracking of state than just an AMQP event listener.

spec/lib/services/workflow-api-service-spec.js
```diff
@@ -0,0 +1,187 @@
...
+        return workflowApiService.defineTaskGraph(graphDefinition)
+        .then(function() {
+            expect(store.persistGraphDefinition).to.have.been.calledOnce;
+            expect(store.persistGraphDefinition).to.have.been.calledWith(graphDefinition);
+            expect(workflowApiService.createGraph).to.have.been.calledOnce;
+            expect(workflowApiService.createGraph).to.have.been.calledWith(graphDefinition);
+        });
+    });
+
+    it('should validate a graph definition', function () {
```
heckj: are there other explicit validation cases that we want to cover or verify here? Wondering what the gap is on logic for the validation logic in defineTaskGraph?

heckj:
:+1: with the rebase and tests passin', you know - the usual. Not sure how to play the corner case of cancelling a graph that doesn't exist - feels like that should properly be a 404 response, but I think you'll 500 error out with the current code.

spec/lib/api/login-spec.js
benbp: Since `rimraf` is actually an on-core dependency, the default search path for the `requireWrapper` call searches in the wrong place by default when any npm links have been made with on-core. This bug was fixed in a couple other places in https://github.com/RackHD/on-http/pull/97, but has been re-introduced with the `login-spec.js` file and I think I was just the first to run it with npm link and notice it. `__dirname` specifies the right relative search path, `undefined` is the slot in the function signature for extra injector overrides or context I think, which we don't want here.

spec/lib/services/workflow-api-service-spec.js
benbp: Primarily what I wanted to test here was just that a validation error would bubble up. I do a full exercising of all the validation failure cases in these two sections of tests within on-core:

https://github.com/benbp/on-core/blob/taskgraph-improvements/spec/lib/workflow/task-graph-spec.js#L97-L200
https://github.com/benbp/on-core/blob/taskgraph-improvements/spec/lib/workflow/task-graph-traversal-spec.js#L19-L110

benbp:
Clarified offline with @heckj about what he was asking: we *will* return a 404 from the `findActiveGraphForNode` call if the workflow does not exist, as that's what that call is designed to do on that case. 

Also decided that it makes the most sense from a strict API design standpoint to return a 400 on the chance when a workflow that was just completed is attempted to be cancelled, so I'll make that update.

benbp:
Also @heckj tests won't pass until I merge on-core and on-tasks, so I'll do that in that order.

lib/api/1.1/northbound/nodes.js
tldavies: :+1: for api updates

lib/api/1.1/northbound/nodes.js
tldavies: add a tests for the edge case.

lib/services/profiles-api-service.js
```diff
@@ -34,9 +34,6 @@ function profileApiServiceFactory(
     var logger = Logger.initialize(profileApiServiceFactory);
 
     function ProfileApiService() {
```
tldavies: Do we still need this function if there is nothing in it?

lib/services/workflow-api-service.js
```diff
@@ -0,0 +1,157 @@
...
+        'Services.Waterline',
+        'TaskGraph.TaskGraph',
+        'Logger',
+        'Errors',
+        'Promise',
+        'Constants',
+        '_'
+    )
+);
+function workflowApiServiceFactory(
```
tldavies: newline

lib/api/1.1/northbound/workflows.js
```diff
@@ -35,7 +44,21 @@ function workflowsRouterFactory (taskgraphRunner, rest, waterline, _) {
...
-     * @api {put} /api/1.1/workflows PUT /
+     * @api {post} /api/1.1/workflows POST /
+     * @apiVersion 1.1.0
+     * @apiDescription run a new workflow
+     * @apiName workflows-run
+     * @apiGroup workflows
+     * @apiError Error problem was encountered, workflow was not run.
+     */
+
+    router.post('/workflows/', rest(function(req) {
```
jlongever: Probably can delete this PR https://github.com/RackHD/on-http/pull/142.

brianparry:
@benbp +1 on the 2.0 API and nodes API service changes.

jlongever:
+1 assuming those check failures are due to dependent PR's.

anhou:
+1

PR #113, opened 2016-02-03
Ensure the accepted exit code is an array of number in bootstrap.js

iceiilin:
As discussed in https://github.com/RackHD/on-core/pull/70, add check of accept exit code to ensure it is an array of number. Otherwise this value won't take effect in evaluating stdout. 
It can cover the case when accepted exit code is '127', and actual exit code is 1 resulting accepting this error unexpectedly.
This PR can be merged independently without impacting discovery process.

### DIFF
--- a/data/templates/bootstrap.js
+++ b/data/templates/bootstrap.js
@@ -138,7 +138,7 @@ function executeTasks(data, timeout) {
                         _task.stderr + "\n" +
                         _task.error.toString());
             console.log("ACCEPTED RESPONSES " + _task.acceptedResponseCodes);
-            if (_task.acceptedResponseCodes &&
+            if (checkValidAcceptCode(_task.acceptedResponseCodes) &&
                 _task.acceptedResponseCodes.indexOf(_task.error.code) >= 0) {
 
                 console.log("_task " + _task.cmd + " error code " + _task.error.code +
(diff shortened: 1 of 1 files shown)


data/templates/bootstrap.js
```diff
@@ -273,4 +273,22 @@ function getFile(downloadUrl, cb) {
...
+ * @private
+ * @param code
+ */
+function checkValidAcceptCode(code) {
+    if (!(code instanceof Array)) {
+        return false;
+    }
+
+    return code.every(function(item) {
+        if (typeof item !== 'number') {
```
benbp: This is fine, although if we wanted to be more flexible we could try to do parseInt() calls against each item, and only return false if that fails. That way `"127"` just becomes `127`

benbp:
:+1: with a minor suggestion about coercing string types to numbers.

data/templates/bootstrap.js
yyscamper: Even further, if the `acceptResponseCodes` is a single valid number, not an array, we can internally wraps this number to an array (` 1 --> [1]`), this should be more user friendly. I proposal this becomes I remembered for OS bootstrap, users often input a single vland ID and then cause rendering error, because actually we required it as an Array of number. 

Of course, this kinds of conversion can be better done in on-tasks who publish the command.

I'm OK to do this or not.

yyscamper:
:+1: though has a minor improvement suggestion.

iceiilin:
@benbp @yyscamper Good suggestions! Both of them are about adjusting user(task)'s input to an array of number, and I totally agree with that. I think a better place to add this function is task.js in on-tasks before running the job. There are two reasons: 
- acceptedResponseCode is used in several jobs like linux-command, linux-catalogs and ipmi-catalog
- child process in on-core only checks array of number, and I prefer to keep the similar logic here to conform on the interfaces
So I think this PR is ready.

PR #114, opened 2016-02-03
Add swagger-ui with login and combined swaggers

zyoung51:
- Install the swagger-ui by running install-swagger-ui.sh
- Swagger UI will be at /swagger-ui

### DIFF
--- a/install-swagger-ui.sh
+++ b/install-swagger-ui.sh
@@ -2,7 +2,7 @@
 
 cd ./static
 rm -rf swagger-ui
-curl -L -o swagger-ui.zip https://github.com/zyoung51/swagger-ui/archive/master.zip
+curl -L -o swagger-ui.zip https://github.com/rackhd/swagger-ui/archive/master.zip
 unzip swagger-ui.zip swagger-ui-master/dist* -d swagger-ui
 mv swagger-ui/swagger-ui-master/dist/* swagger-ui
 rm swagger-ui.zip
--- a/install-swagger-ui.sh
+++ b/install-swagger-ui.sh
@@ -1,0 +1,8 @@
+# Copyright 2015, EMC, Inc.
+
+cd ./static
+rm -rf swagger-ui
+curl -L -o swagger-ui.zip https://github.com/zyoung51/swagger-ui/archive/master.zip
+unzip swagger-ui.zip swagger-ui-master/dist* -d swagger-ui
+mv swagger-ui/swagger-ui-master/dist/* swagger-ui
+rm swagger-ui.zip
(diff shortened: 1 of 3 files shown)


zyoung51:
@RackHD/corecommitters @uppalk1 @keedya @tannoa2 @brianparry

jlongever:
yikes
test this please

install-swagger-ui.sh
```diff
@@ -0,0 +1,8 @@
+# Copyright 2015, EMC, Inc.
+
+cd ./static
+rm -rf swagger-ui
+curl -L -o swagger-ui.zip https://github.com/zyoung51/swagger-ui/archive/master.zip
```
benbp: Do we want to just put this in our bintray?

jlongever:
-1

@zyoung51 could you fix this?  then +1
Error Details: options.swaggerUiDir path does not exist: /var/lib/jenkins/jobs/on-http/workspace/build/static/swagger-ui
Stack Trace: Error: options.swaggerUiDir path does not exist: static/swagger-ui

install-swagger-ui.sh
```diff
@@ -0,0 +1,8 @@
...
+unzip swagger-ui.zip swagger-ui-master/dist* -d swagger-ui
```
benbp: Also, I've found that unzip usually isn't installed by default on some distros including Ubuntu, maybe gzip is a more friendly choice?

install-swagger-ui.sh
jlongever: @benbp I suggested forking into RackHD, but bintray is another option.   Looks like Zack is copying what @rolandpoulter did with web-ui. Should this be applied across the board for things like this?

benbp:
-1 I would prefer not to keep assets in privately controlled github accounts, even though it's not like the system will break without it. If our bintray isn't appropriate (our you need admin access?) maybe that means we have a larger problem to solve about where we host assets like this.

install-swagger-ui.sh
zyoung51: This is a copy paste of the webUI install script...

install-swagger-ui.sh
benbp: Looks like @heckj just forked swagger-ui for that.

Yeah, I think it makes sense to fork as a pattern for upstream dependencies we want to host ourselves, we do it with DI, web-ui as you mention, swagger-ui now, lib-ffi, etc. It can be a pain if you have to keep them in sync but seems like an okay pattern for just grabbing assets. Of course, it still means we're tied to Github, but you have to draw the line somewhere :)

heckj: yep - per conversation in Slack channel, seemed most sensible to keep a local fork in RackHD org instead of a personal one if we're going to use it internally

jlongever: Should this be applied for web-ui stuff that @rolandpoulter committed?  I believe that references a personal fork.

heckj: yep

zyoung51:
@benbp @jlongever Fixed the missed swagger-ui installation and changed the script to pull from RackHD's swagger-ui repo

jlongever:
+1

benbp:
:+1: thanks!

PR #116, opened 2016-02-04
updated node service

abildw:
updated 1.1 controller and node service in preparation for Swagger 2.0 Node Controller.
Still passes all unit testing.

@geoff-reid @dalebremner @brianparry @srinia6 @RackHD/corecommitters

### DIFF
--- a/lib/api/1.1/northbound/nodes.js
+++ b/lib/api/1.1/northbound/nodes.js
@@ -15,7 +15,6 @@ di.annotate(nodesRouterFactory, new di.Inject(
         'Http.Services.Api.Nodes',
         'Services.Configuration',
         'Task.Services.OBM',
-        'ipmi-obm-service',
         'Logger',
         '_',
         'Errors'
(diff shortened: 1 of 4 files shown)


lib/api/1.1/northbound/nodes.js
```diff
@@ -430,12 +318,7 @@ function nodesRouterFactory (
      */
 
     router.post('/nodes/:macaddress/dhcp/whitelist', rest(function(req) {
```
benbp: These whitelist routes are actually just dead routes if you feel so inclined to delete them.

benbp:
:+1:

benbp:
As a general trend, I think we should move the unit test code for services out of the api tests and into the services tests directory when we move logic out of the api and into services, but it's a mix all over the place right now so it's not like I think this PR should be blocked or anything (just something to keep in mind). Good work!

heckj:
:+1:

PR #130, opened 2016-02-11
Modified Swagger rendering fitting

dalebremner:
* Allows a rendering template to be specified in the Swagger definition file

### DIFF
--- a/lib/fittings/swagger_render.js
+++ b/lib/fittings/swagger_render.js
@@ -1,11 +1,24 @@
+// Copyright 2016, EMC, Inc.
+
 'use strict';
 
+var injector = require('../../index.js').injector;
+
 module.exports = function create(fittingDef, bagpipes) {
+    var swag = injector.get('Http.Services.Swagger');
     return function swagger_render(context, next) {
         if (!context.response.headersSent) {
             var status = context.request.swagger.options.success || 200;
-            context.response.status(status).json(context.response.body);
+            var templateNameKey = fittingDef.templateNameKey;
+            var operation = context.request.swagger.operation;
+            var template = operation[templateNameKey];
+            if (template === undefined) {
+                context.response.status(status).json(context.response.body);
+            } else {
+                swag.renderer(template, context.response.body, next).then(function (renderedOutput) {
+                    context.response.status(status).json(renderedOutput);
+                });
+            }
         }
     };
 };
-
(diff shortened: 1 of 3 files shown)


dalebremner:
@RackHD/corecommitters @abildw @geoff-reid @srinia6

geoff-reid:
:+1:

abildw:
:+1:

srinia6:
+1

brianparry:
:+1:

lib/fittings/swagger_render.js
```diff
@@ -1,11 +1,26 @@
...
 'use strict';
 
+var injector = require('../../index.js').injector;
+
 module.exports = function create(fittingDef, bagpipes) {
     return function swagger_render(context, next) {
         if (!context.response.headersSent) {
             var status = context.request.swagger.options.success || 200;
-            context.response.status(status).json(context.response.body);
+            var swag = injector.get('Http.Services.Swagger');
```
brianparry: @dalebremner I'd move this injector call outside the swagger_render function.  That way it only needs to be injected when the fitting is created, rather than each time the fitting is called.

lib/services/swagger-api-service.js
```diff
@@ -105,10 +109,23 @@ function swaggerFactory(
...
+        return templates.get(templateName, ['global'])
+            .then(function(template) {
+                return template.contents;
+            })
+            .then(function(contents) {
+                var output = Promise.map(data, function(item) {
+                    return JSON.parse(ejs.render(contents, item));
+                });
+                return output;
+            });
```
brianparry: @dalebremner Should be a `.catch(error)` here that calls `next(_processError(err)`

lib/fittings/swagger_render.js
```diff
@@ -1,11 +1,26 @@
...
+            var templateNameKey = fittingDef.templateNameKey;
+            var operation = context.request.swagger.operation;
+            var template = operation[templateNameKey];
+            if (template === undefined) {
+                context.response.status(status).json(context.response.body);
+            } else {
+                swag.renderer(template, context.response.body).then(function (renderedOutput) {
+                    context.response.status(status).json(renderedOutput);
+                }, function (error) {
```
brianparry: @dalebremner I think error handling should happen in the swagger api service using a `.catch`.  That way you can use the API service error parsing.

brianparry:
:+1:

PR #131, opened 2016-02-11
Add profile loading to skupack bundle

zyoung51:


zyoung51:
Depends on: https://github.com/RackHD/on-core/pull/84

spec/lib/services/sku-pack-service-spec.js
```diff
@@ -107,8 +117,22 @@ describe("SKU Pack Service", function() {
                 destroy: sinon.stub()
             };
 
+            waterline.templates = {
```
yyscamper: I checked these stub for waterline service has not been restored, so this may impact other unit test.

zyoung51: Not sure what you want restored?  There is no 'waterline.<model>' when this is run, which is why I create them.  I cannot restore that which doesn't exist.

yyscamper: Since you mock a dummy templates property to  the waterline service, if not remove it from waterline, then other unit test will see the same dummy templates property as well. This may impact some unit-test (if we have defined one), like check whether the waterline service has a function 'create'.

zyoung51: @yyscamper This identical pattern appears throughout the unittests.  Each unittest sets up what they need at the start (as I did) and overrides what was there by doing so. I can understand your concern of someone accidentally forgetting to setup their mock and unintentionally using mine.  I will make a change to delete these in the after().

benbp:
:+1:

jlongever:
+1

PR #134, opened 2016-02-17
pollers unit-test for 2.0 and services 

srinia6:
@dalebremner @brianparry @geoff-reid @abildw @RackHD/corecommitters

### DIFF
--- a/spec/lib/api/2.0/pollers-spec.js
+++ b/spec/lib/api/2.0/pollers-spec.js
@@ -1,4 +1,4 @@
-// Copyright 2015, EMC, Inc.
+// Copyright 2016, EMC, Inc.
 /* jshint node:true */
 
 'use strict';
--- a/spec/lib/api/2.0/pollers-spec.js
+++ b/spec/lib/api/2.0/pollers-spec.js
@@ -1,0 +1,330 @@
+// Copyright 2015, EMC, Inc.
+/* jshint node:true */
+
+'use strict';
+
+describe('Http.Api.Pollers', function () {
+    var taskProtocol;
+    before('start HTTP server', function () {
+        this.timeout(5000);
+        taskProtocol = {
+            requestPollerCache: sinon.stub()
+        };
+        return helper.startServer([
+            dihelper.simpleWrapper(taskProtocol, 'Protocol.Task')
+        ]);
+    });
+
+    beforeEach('reset test DB', function () {
+        return helper.reset();
+    });
+
+    after('stop HTTP server', function () {
+        return helper.stopServer();
+    });
+
+    it('should return an empty array from GET /pollers', function () {
+        return helper.request().get('/api/2.0/pollers')
+            .expect('Content-Type', /^application\/json/)
+            .expect(200, []);
+    });
+
+    it('should create an IPMI poller with POST /pollers', function () {
+        return helper.request().post('/api/2.0/pollers')
+            .send({
+                type: 'ipmi',
+                pollInterval: 5000,
+                config: {
+                    command: 'sdr',
+                    host: '0.0.0.0',
+                    user: 'myuser',
+                    password: 'mypass'
+                }
+            })
+            .expect('Content-Type', /^application\/json/)
+            .expect(201)
+            .expect(function (res) {
+                expect(res.body).to.have.property('type', 'ipmi');
+                expect(res.body).to.have.property('node', null);
+                expect(res.body).to.have.property('failureCount', 0);
+                expect(res.body).to.have.property('pollInterval', 5000);
+                expect(res.body).to.have.property('config').to.be.an('object');
+                expect(res.body).to.have.deep.property('config.command', 'sdr');
+                expect(res.body).to.have.deep.property('config.host', '0.0.0.0');
+                expect(res.body).to.have.deep.property('config.user', 'myuser');
+                expect(res.body).to.have.deep.property('config.password', 'mypass');
+                expect(res.body).to.have.property('id').to.be.a('string');
+                expect(res.body).to.have.property('createdAt').to.be.a('string');
+                expect(res.body).to.have.property('updatedAt').to.be.a('string');
+                expect(res.body).to.have.property('lastStarted', null);
+                expect(res.body).to.have.property('lastFinished', null);
+            });
+    });
+
+    it('should create an SNMP poller with POST /pollers', function () {
+        return helper.request().post('/api/2.0/pollers')
+            .send({
+                type: 'snmp',
+                pollInterval: 6000,
+                config: {
+                    host: '0.0.0.0',
+                    communityString: 'public',
+                    extensionMibs: [
+                        'PowerNet-MIB::rPDULoadDevMaxPhaseLoad', // max load for this PDU
+                        'PowerNet-MIB::sPDUOutletCtl', // state of each outlet (1 on, 2 off)
+                        'PowerNet-MIB::rPDULoadStatusLoad', // current in 10ths of amps
+                        'PowerNet-MIB::rPDUIdentDeviceLinetoLineVoltage' // voltage
+                    ]
+                }
+            })
+            .expect('Content-Type', /^application\/json/)
+            .expect(201)
+            .expect(function (res) {
+                expect(res.body).to.have.property('type', 'snmp');
+                expect(res.body).to.have.property('node', null);
+                expect(res.body).to.have.property('failureCount', 0);
+                expect(res.body).to.have.property('pollInterval', 6000);
+                expect(res.body).to.have.property('config').to.be.an('object');
+                expect(res.body).to.have.deep.property('config.host', '0.0.0.0');
+                expect(res.body).to.have.deep.property('config.communityString', 'public');
+
+                expect(res.body).to.have.deep.property('config.extensionMibs')
+                    .to.be.an.instanceOf(Array);
+
+                expect(res.body).to.have.deep.property('config.extensionMibs')
+                    .with.length(4)
+                    .and.to.include('PowerNet-MIB::rPDULoadDevMaxPhaseLoad')
+                    .and.to.include('PowerNet-MIB::sPDUOutletCtl')
+                    .and.to.include('PowerNet-MIB::rPDULoadStatusLoad')
+                    .and.to.include('PowerNet-MIB::rPDUIdentDeviceLinetoLineVoltage');
+
+                expect(res.body).to.have.property('id').to.be.a('string');
+                expect(res.body).to.have.property('createdAt').to.be.a('string');
+                expect(res.body).to.have.property('updatedAt').to.be.a('string');
+                expect(res.body).to.have.property('lastStarted', null);
+                expect(res.body).to.have.property('lastFinished', null);
+            });
+    });
+
+    // TODO: probably change this later to not just be a static check
+    var library = [
+        {
+        name: 'ipmi',
+        node: true,
+        config: [
+            {
+            key: 'host',
+            type: 'string'
+        },
+        {
+            key: 'user',
+            type: 'string',
+            defaultsTo: 'admin'
+        },
+        {
+            key: 'password',
+            type: 'string',
+            defaultsTo: 'admin'
+        },
+        {
+            key: 'alerts',
+            type: 'json',
+            required: false
+        }
+        ]
+    },
+    {
+        name: 'snmp',
+        config: [
+            {
+            key: 'host',
+            type: 'string',
+            required: true
+        },
+        {
+            key: 'communityString',
+            type: 'string',
+            required: true
+        },
+        {
+            key: 'extensionMibs',
+            type: 'string[]'
+        }
+        ]
+    }
+    ];
+
+    it('should retrieve the library with GET /pollers/library', function () {
+        return helper.request().get('/api/2.0/pollers/library')
+            .expect('Content-Type', /^application\/json/)
+            .expect(200, library);
+    });
+
+    it('should retrieve the ipmi library entry with GET /pollers/library/ipmi', function () {
+        return helper.request().get('/api/2.0/pollers/library/ipmi')
+            .expect('Content-Type', /^application\/json/)
+            .expect(200, library[0]);
+    });
+
+    it('should retrieve the snmp library entry with GET /pollers/library/snmp', function () {
+        return helper.request().get('/api/2.0/pollers/library/snmp')
+            .expect('Content-Type', /^application\/json/)
+            .expect(200, library[1]);
+    });
+
+    describe('non-existent node id', function () {
+        var badId = 'bad';
+        it('should 404 with GET /pollers/:id ', function () {
+            return helper.request().get('/api/2.0/pollers/' + badId)
+            .expect(404);
+        });
+
+        it('should 404 with PATCH /pollers/:id', function () {
+            return helper.request().patch('/api/2.0/pollers/' + badId)
+            .send()
+            .expect(404);
+        });
+
+        it('should 404 with DELETE /pollers/:id', function () {
+            return helper.request().delete('/api/2.0/pollers/' + badId)
+            .expect(404);
+        });
+    });
+
+    describe('POST /pollers afterwards', function () {
+        var poller;
+        beforeEach(function () {
+           return helper.request().post('/api/2.0/pollers')
+            .send({
+                type: 'ipmi',
+                pollInterval: 5000,
+                config: {
+                    command: 'sdr',
+                    host: '0.0.0.0',
+                    user: 'myuser',
+                    password: 'mypass'
+                }
+            })
+            .expect('Content-Type', /^application\/json/)
+            .expect(201)
+            .then(function (req) {
+                poller = req.body;
+            });
+        });
+        it('should contain the new poller in GET /pollers', function () {
+            delete poller.name;
+            return helper.request().get('/api/2.0/pollers')
+            .expect('Content-Type', /^application\/json/)
+            .expect(200, [poller]);
+        });
+
+        it('should return the same poller from GET /pollers/:id', function () {
+            delete poller.name;
+            return helper.request().get('/api/2.0/pollers/' + poller.id)
+            .expect('Content-Type', /^application\/json/)
+            .expect(200, poller);
+        });
+
+        it('should update with PATCH /pollers', function () {
+            poller.pollInterval = 20000;
+            return helper.request().patch('/api/2.0/pollers/' + poller.id)
+            .send(poller)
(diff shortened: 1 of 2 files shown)


dalebremner:
:+1:

jlongever:
Interesting, no coveralls coverage check?

srinia6:
@jlongever Does it need to be turned on ? If so, how?

jlongever:
@srinia6 it should be on by default.   not sure why it didn't run.

spec/lib/api/2.0/pollers-spec.js
```diff
@@ -0,0 +1,330 @@
+// Copyright 2015, EMC, Inc.
```
abildw: update to 2016

abildw:
I just had a minor comment about the date, but other than that it looks good to me

:+1:

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+        waterline = helper.injector.get('Services.Waterline');
+        waterline.workitems = {
+            needByIdentifier: sinon.stub().resolves(),
+            find: sinon.stub().resolves([]),
+            create: sinon.stub().resolves(),
+            updateByIdentifier: sinon.stub().resolves(),
+            destroyByIdentifier: sinon.stub().resolves()
+        };
+
+	    taskProtocol = helper.injector.get("Protocol.Task");
```
yyscamper: indent error

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+        taskProtocol.requestPollerCache = sinon.stub();
+
+        pollerService = helper.injector.get("Http.Services.Api.Pollers");
+
+    });
+
+    describe("getPollers", function() {
+        it('should expose the appropriate methods', function() {
+            pollerService.should.have.property('getPollers')
+                .that.is.a('function').with.length(1);
```
yyscamper: why expect a function has length 1?

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+        it('Run getPollers', function() {
+            var mockPoller = [{
+                id: '4532',
+                name: 'Pollers.IPMI',
+                config: {}
+            }];
+
+            waterline.workitems.find.resolves([mockPoller]);
+            return pollerService.getPollers().then(function (pollers) {
+                expect(pollers).to.deep.equal([mockPoller]);
```
yyscamper: the `mockPoller` is already an Array, why still wraps it into another Array? I would expect  the `pollers' is exactly equals to the `mockPoller`

spec/lib/services/pollers-api-service-spec.js
srinia6: getPollers function takes one argument. thats why.

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+            });
+        });
+       
+        it('should return error message  if no poller informations are found', function () {
+            var mockPoller = {"message":"NotFoundError: Could not find workitem with identifier"};
+            waterline.workitems.find.resolves(mockPoller);
+            return pollerService.getPollers().should.eventually.become(mockPoller);
```
yyscamper: `eventually` is superfluous here, either `should.eventually.equal.to(mockPoller)` or `should.become(mockPoller)`

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+        });
+    });
+
+    describe("getPollersById", function() {
+        it('should expose the appropriate methods', function() {
+            pollerService.should.have.property('getPollersById')
+            .that.is.a('function').with.length(1);
```
yyscamper: why expect a function has length 1?

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+        });
+
+        it('Run getPollersById', function() {
+            var mockPoller = [{
+                id: '4532',
+                name: 'Pollers.IPMI',
+                config: {}
+            }];
+            waterline.workitems.needByIdentifier.resolves([mockPoller]);
+            return pollerService.getPollersById().then(function (pollers) {
```
yyscamper: Since you are testing getPollersById, so I think you should mock at least 2 pollers in db and then pass a poller id into getPollersById and expect to get the correct poller, something like this:
```javascript
var mockPoller = [{
    id: '4532',
    name: 'Pollers.IPMI',
    config: {}
}, {
    id: '1234',
    name: 'Pollers.TEST',
    config: {}
}];
waterline.workitems.needByIdentifier.withArgs({'id': '4532'}).resolves(mockPoller[0]);
waterline.workitems.needByIdentifier.withArgs({'id': '1234'}).resolves(mockPoller[1]);
return Promise.all(['4532', '1234'].map(function(id, index) {
    return pollerService.getPollersById(index).then(function (poller) {
    expect(poller).to.deep.equal(mockPoller[index]);
});
```

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+                expect(pollers).to.deep.equal([mockPoller]);
+            });
+        });
+
+        it('should return error message if no specific poller information are found', function () {
+            var mockPoller = {"message":"NotFoundError: Could not find workitem with identifier"};
+            waterline.workitems.needByIdentifier.resolves(mockPoller);
+            return pollerService.getPollersById().should.eventually.become(mockPoller);
```
yyscamper: This test case design is still good for me. Since you are testing an error case, shouldn't mock the `waterline.workitems.needByIdentifier` to be rejected and expect the getPollersById is rejected as well?

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,258 @@
...
+        });
+
+    });
+
+    describe("getPollerLib", function() {
+        it('should expose the appropriate methods', function() {
+            pollerService.should.have.property('getPollerLib')
+            .that.is.a('function').with.length(0);
```
yyscamper: Why testing a function has length 0?

spec/lib/services/pollers-api-service-spec.js
yyscamper: Thanks, it's my fault, I didn't realized the function has a property `length`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length

yyscamper:
:-1: for this, since I see a lot bad test case design. @srinia6 please help fix this

srinia6:
@ yyscamper: updated. thank you!

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,251 @@
...
+        waterline = helper.injector.get("Services.Waterline");
+        waterline.workitems = {
+            needByIdentifier: sinon.stub().resolves(),
+            find: sinon.stub().resolves([]),
+            create: sinon.stub().resolves(),
+            updateByIdentifier: sinon.stub().resolves(),
+            destroyByIdentifier: sinon.stub().resolves()
+        };
+
+	taskProtocol = helper.injector.get("Protocol.Task");
```
brianparry: nit: indent

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,251 @@
...
+            waterline.workitems.needByIdentifier.withArgs({"id": "4532"}).resolves(mockPoller[0]);
+            waterline.workitems.needByIdentifier.withArgs({"id": "1234"}).resolves(mockPoller[1]);
+            Promise.all(["4532", "1234"].map(function(id, index) {
+                return pollerService.getPollersById(index)})).then(function (pollers) {
+                    expect(pollers).to.deep.equal(mockPoller[index]);
+                });
+         });
+
+        it("should return error if specific poller info is not found", function () {
+            var mockPoller = {"message":"NotFoundError: Could not find workitem with identifier"};
```
brianparry: @srinia6 I think we want to make this an error object: `var mockPollerError = new Errors.NotFoundError('Could not find workitem with identifier')`.

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,251 @@
...
+                config: {}
+            }];
+
+            waterline.workitems.find.resolves(mockPoller);
+            return pollerService.getPollers().then(function (pollers) {
+                expect(pollers).to.deep.equal(mockPoller);
+            });
+        });
+        it('should return error if poller informations is not found', function () {
+            var mockPoller = {"message":"NotFoundError: Could not find workitem with identifier"};
```
brianparry: @srinia6 I think we want to make this an error object: `var mockPollerError = new Errors.NotFoundError('Could not find workitem with identifier')`.

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,251 @@
...
+                config: {}
+            }];
+            waterline.workitems.create.resolves(mockPoller);
+            return pollerService.postPollers().then(function (pollers) {
+                expect(pollers).to.deep.equal(mockPoller);
+            });
+        });
+
+        it("Throws error when  postPollers runs with invalid input", function() {
+            var mockPoller = {
```
brianparry: @srinia6 I think we want to make this an error object: `var mockPollerError = new Errors.ValidationError('Validation errors')`.

spec/lib/services/pollers-api-service-spec.js
srinia6: Hmm i fixed that. I will remove white-spaces/tabs.

jlongever:
LGTM +1

yyscamper:
:+1: after fix the hounci warning "Line is tool long"

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,251 @@
...
+            {
+                id: "4532",
+                name: "Pollers.TEST",
+                config: {}
+            }];
+            waterline.workitems.needByIdentifier.withArgs({"id": "4532"}).resolves(mockPoller[0]);
+            waterline.workitems.needByIdentifier.withArgs({"id": "1234"}).resolves(mockPoller[1]);
+            Promise.all(["4532", "1234"].map(function(id, index) {
+                return pollerService.getPollersById(index)})).then(function (pollers) {
+                    expect(pollers).to.deep.equal(mockPoller[index]);
```
anhou: return Promise.all, 'index' is not defined may be caused by this?, and please check 'Promise.all(' 's another ‘)'

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,251 @@
...
+            },
+            {
+                id: "4532",
+                name: "Pollers.TEST",
+                config: {}
+            }];
+            waterline.workitems.needByIdentifier.withArgs({"id": "4532"}).resolves(mockPoller[0]);
+            waterline.workitems.needByIdentifier.withArgs({"id": "1234"}).resolves(mockPoller[1]);
+            Promise.all(["4532", "1234"].map(function(id, index) {
+                return pollerService.getPollersById(index)})).then(function (pollers) {
```
anhou: 'Missing' semicolon' should be caused by 'pollerService.getPollersById(index)}))' 's  "}" , please have a check. and also please help check all the test cases, I'm curious that unit test cases could pass with these typos.

brianparry:
:+1:

anhou:
@srinia6 all houndci's review comments should be fixed. if it could be ignored, please comment it.

:+1: after fix all houndci issues

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,250 @@
...
+                config: {}
+            },
+            {
+                id: "4532",
+                name: "Pollers.TEST",
+                config: {}
+            }];
+            waterline.workitems.needByIdentifier.withArgs({"id": "4532"}).resolves(mockPoller[0]);
+            waterline.workitems.needByIdentifier.withArgs({"id": "1234"}).resolves(mockPoller[1]);
+            Promise.all(["4532", "1234"].map(function(id, index) {
```
yyscamper: just realized this promise has not been returned, this should be fixed before merge.

spec/lib/services/pollers-api-service-spec.js
yyscamper: Yes, this is a critical problem, I checked  the code, the brackets don't match.

spec/lib/services/pollers-api-service-spec.js
yyscamper: I checked the brackets don't match, so this need be fixed.

yyscamper:
I have to withdraw my approval, because I still see a critical error:
(1) The Promise.all need be returned.
(2) The houndci warning about index and semicolon need be fixed, because it reflects a critical error that  the brackets don't match. (I wonder why the travis ci pass the test).

Meanwhile, the hounci warning about "Line is too long" should also be fixed.

spec/lib/services/pollers-api-service-spec.js
yyscamper: Let the code be well aligned, it will help identifier some bug:
```javascript
return Promise.all(
    ["4532", "1234"].map(function(id, index) {
        return pollerService.getPollersById(index)
            .then(function (poller) {
                expect(poller).to.deep.equal(mockPoller[index]);
            });
    });
);
```

brianparry: I think this may be the best way to do this.

```javascript
Promise.map(["4532", "1234"], function(id, index) {
    return pollerService.getPollersById(index);
}).then(function (pollers) {
    _.forEach(pollers, function(poller, index) {
        expect(poller).to.deep.equal(mockPoller[index]);
    });
});
```

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,250 @@
...
+                return pollerService.getPollersById(index)})).then(function (pollers) {
+                    expect(pollers).to.deep.equal(mockPoller[index]);
+                });
+         });
+
+        it("should return error if specific poller info is not found", function () {
+            var mockPollerError = new Errors.NotFoundError("Could not find workitem with identifier");
+            waterline.workitems.needByIdentifier.rejects(mockPollerError);
+            return pollerService.getPollersById().should.eventually.be.rejectedWith(mockPollerError);
```
srinia6: Ignoring this error.

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,250 @@
...
+                config: {}
+            }];
+
+            waterline.workitems.find.resolves(mockPoller);
+            return pollerService.getPollers().then(function (pollers) {
+                expect(pollers).to.deep.equal(mockPoller);
+            });
+        });
+        it('should return error if poller informations is not found', function () {
+            var mockPollerError = new Errors.NotFoundError("Could not find workitem with identifier");
```
srinia6: Ignoring this error.

spec/lib/services/pollers-api-service-spec.js
```diff
@@ -0,0 +1,255 @@
...
+                return pollerService.getPollersById(id);
+            }).then(function (pollers) {
+                _.forEach(pollers, function(poller, index) {
+                    expect(poller).to.deep.equal(mockPoller[index]);
+                });
+            });
+         });
+
+        it("should return error if specific poller info is not found", function () {
+            var mockPollerError = new Errors.NotFoundError("Could not find workitem with identifier");
```
srinia6: Ignoring this error.

srinia6:
test this please

yyscamper:
@srinia6, Ithink you miss another fix: the promise.map need be returned otherwise your assertion logic will not take effect.

Meanwhile, for "Line is too long", if you do have reason to ignore it, you'd use jshint ignore synatx(please search jshint manual) and append it after the line you want to ingore.

srinia6:
@yyscamper I looked at the test case. It is working as expected. The reason it was passing previously was the data that resolved came back as undefined . This was fixed by replacing ["id" : "1234" ] with ["1234"]. Also, the brackets were closed appropriately.  @brianparry the build passed.

yyscamper:
@srinia6 : understand the brackets design, your current code is OK for me. 
My primary concern is that the promise in the test case doesn't be returned. You can do an experiment if not return the Promise. Just change the "expect(xxx).to.equal(xxx)" to the inverted logic "expect(xxx).to.not.equal(xxx)", you will find both pass!!

This is not a style problem, this is a problem about library/framework usage problem, so this need be fixed. If you add your assertion logic in the Promise, that promise must be returned, otherwise mocha will not wait the test case is actually done. I see all your code's Promises are returned except this one. Please refer to the mocha guideline about Promise: https://mochajs.org/#working-with-promises.

yyscamper:
:+1: after fix the problem that Promise doesn't been returned

srinia6:
@yyscamper : Not sure if I understood your concern. You suggested these changes!

I  did try substituting per your suggestion/concern i.e.- "** Just change the "expect(xxx).to.equal(xxx)" to the inverted logic "expect(xxx).to.not.equal(xxx)", you will find both pass!! **" 

I see "crisp" error  showing-up as below:

**Unhandled rejection AssertionError: expected { Object (id, name, ...) } to not equal { Object (id, name, ...) }**
    at .........
    at arrayEach (.....)
    at Function.<anonymous> (.....)
    at .........
    at tryCatcher (.......)
    at Promise._settlePromiseFromHandler (.....)
    at Promise._settlePromiseAt (.....)
    at Promise._settlePromises (....)
    at Async._drainQueue (......)
    at Async._drainQueues (....)
    at Async.drainQueues (,,,,)
    at process._tickDomainCallback (node.js:459:13)

SO i am guessing it IS doing all the "right" thing?

srinia6:
@yyscamper Fixed it. Turns out that i am not seeing the failing case because the promise is mocked-out (here).

brianparry:
@srinia6: It works now because the Promise returned by Promise.map is immediately fulfilled because we're just stubbing the waterline calls.  If those calls were asynchronous, your `then` callback would get invoked later which could give you unexpected results.  You just need to return the promise that gets returned by `Promise.map(..).then`.

yyscamper:
good. @srinia6

srinia6:
@RackHD/corecommitters - Can you review/merge this please?

jlongever:
+1

PR #135, opened 2016-02-18
Node Tagging

zyoung51:
- Add 'tags' to nodes to identify groups of nodes by arbitrary tag names
- Add 1.1 and 2.0 paths for new /nodes and /tags routes.

### DIFF
--- a/lib/api/1.1/northbound/nodes.js
+++ b/lib/api/1.1/northbound/nodes.js
@@ -424,16 +424,47 @@ function nodesRouterFactory (
         return nodeApiService.delActiveWorkflowById(req.params.identifier);
     }));
 
-    router.get('/nodes/:identifier/tags', rest(function(req) {
+    /**
+     * @api {get} /api/1.1/nodes/:identifier/tags GET /:id/tags
+     * @apiVersion 1.1.0
+     * @apiDescription Retrieve the tags on the node
+     * @apiName get-node-tags
+     * @apiGroup nodes
+     * @apiParam {String} identifier node identifier
+     * @apiSuccess (Success 200) {json} tags  the list of tags for that node.
+     * @apiError NotFound The node with the <code>identifier</code> was not found.
+     */
+     router.get('/nodes/:identifier/tags', rest(function(req) {
         return nodeApiService.getTagsById(req.params.identifier);
     }));
 
+    /**
+     * @api {patch} /api/1.1/nodes/:identifier/tags PATCH /:id/tags
+     * @apiVersion 1.1.0
+     * @apiDescription Add tags to a node
+     * @apiName patch-node-tags
+     * @apiGroup nodes
+     * @apiParam {String} identifier node identifier
+     * @apiSuccess (Success 200) {json} tags  the list of tags for that node.
+     * @apiError NotFound The node with the <code>identifier</code> was not found.
+     */
     router.patch('/nodes/:identifier/tags', rest(function (req) {
         if(req.body.tags) {
             return nodeApiService.addTagsById(req.params.identifier, req.body.tags);
         }
     }));
 
+    /**
+     * @api {delete} /api/1.1/nodes/:identifier/tags/:name DELETE /:id/tags/:name
+     * @apiVersion 1.1.0
+     * @apiDescription Delete tag from a node
+     * @apiName del-node-tag
+     * @apiGroup nodes
+     * @apiParam {String} identifier node identifier
+     * @apiParam {String} Name of tag to remove
+     * @apiSuccess (Success 204) Empty
+     * @apiError NotFound The node with the <code>identifier</code> was not found.
+     */
     router.delete('/nodes/:identifier/tags/:name', rest(function(req) {
         return nodeApiService.removeTagsById(req.params.identifier, req.params.name);
     }));
--- a/lib/api/1.1/northbound/nodes.js
+++ b/lib/api/1.1/northbound/nodes.js
@@ -424,6 +424,19 @@ function nodesRouterFactory (
         return nodeApiService.delActiveWorkflowById(req.params.identifier);
     }));
 
+    router.get('/nodes/:identifier/tags', rest(function(req) {
+        return nodeApiService.getTagsById(req.params.identifier);
+    }));
+
+    router.patch('/nodes/:identifier/tags', rest(function (req) {
+        if(req.body.tags) {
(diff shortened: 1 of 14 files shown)


zyoung51:
Depends on https://github.com/RackHD/on-core/pull/88, https://github.com/RackHD/on-taskgraph/pull/62

lib/api/1.1/northbound/nodes.js
```diff
@@ -426,6 +426,19 @@ function nodesRouterFactory (
...
+        return nodeApiService.getTagsById(req.params.identifier);
+    }));
+
+    router.patch('/nodes/:identifier/tags', rest(function (req) {
+        if(req.body.tags) {
+            return nodeApiService.addTagsById(req.params.identifier, req.body.tags);
+        }
+    }));
+
+    router.delete('/nodes/:identifier/tags/:name', rest(function(req) {
```
anhou: need annotations, v1.1 api apidoc, and python-client is based on router's annotations.

lib/api/1.1/northbound/tags.js
```diff
@@ -0,0 +1,69 @@
...
+function tagsRouterFactory (
+    Tags,
+    Nodes, 
+    rest, 
+    _, 
+    Promise
+) {
+    var router = express.Router();
+
+    router.get('/tags', rest(function (req) {
```
anhou: the same with above,  need annotations, v1.1 api apidoc, and python-client is based on router's annotations.

lib/services/nodes-api-service.js
```diff
@@ -451,5 +453,80 @@ function nodeApiServiceFactory(
...
+        });
+    };
+
+    /**
+     * Get a list of tags applied to the specified id
+     * @param  {String}     id
+     * @return {Promise}    Resolves to an array of tags
+     */
+    NodeApiService.prototype.getTagsById = function(id) {
+        return Promise.resolve().then(function() {
```
anhou: this part of codes could be simplified as below. type in editor, if somethings is wrong, please correct it.

	assert.isMongoId(id, 'the id must be a valid mongo id');
        return waterline.nodes.needByIdentifier(id)
	.then(function (node) {
            return node.tags;
        });

zyoung51: The assert is intentionally within the Promise so I can Promise.catch() them.  Simplifying it as you suggest means I need to add complexity in try/catch somewhere else, which I didn't want to do.

WangWinson:
General question, does add new API route cause a API version bump ?

lib/services/nodes-api-service.js
anhou: IMHO, the exception errors could (already)be handled in a area of concentration. don't need to be promised in every function. if don't want to use try/catch, could use promise.resolve(..).catch(...) somewhere else to catch the exception.

anhou:
:+1:  after fix annotation, unused 'res', travisCI failure, and if others have no concerns about promise assertion

lib/services/nodes-api-service.js
zyoung51: @anhou the promise.resolve().catch() that you are advising me to do is what is done already.  The .catch() is in the caller of the service.  They are already handled there in an area of concentration.  Doing as you suggest would spread out the handling to each individual function in the service and swallow the errors preventing them from being reported up.  Are you sure that's what you want me to do?

jlongever:
test this please

lib/services/nodes-api-service.js
anhou: @zyoung51 I mean if error handing is already in an area of concentration, the throwed assert exception in each function could be handled there,  so don't need to be promised in each function here. and if don't want complexity of try/catch in somewhere else, promise's catch also could work there.

zyoung51: @anhou: This is how to get them thrown into the promise chain.  If I move the asserts outside the promise chain, the promise chain cannot catch them.  I did it with the asserts outside the promise chain originally, and it works if something else calls the service from within a promise chain, but that doesn't happen in the express routers.  IE: making the change you are asking for means we have to rewrite the express routers.  Which is fine with me, but I want to be sure it is what you want and you understand what you are asking me to do.

zyoung51:
@RackHD/corecommitters Can you answer Winson's question above?

jlongever:
@WangWinson 

http://rackhd.readthedocs.org/en/latest/devguide/index.html#api-versioning-conventions
 
Specifically:
 
8.3.3. API Version Guidelines
Use the following guide lines when determining if a new API version is needed.
 
The following changes require a new API version:
 
- changing the semantic meaning of a URI route
- removing a URI route
 
The following changes do not require a new API version:
 
- adding an entirely new URI route
- changing the query parameters (pagination, filtering, etc.) accepted by the URI route
- changing the return values on error conditions
- changing the data structure for a resource at a given URI

jlongever:
test this please

WangWinson:
Thanks @jlongever
:+1:

lib/services/nodes-api-service.js
anhou: @zyoung51 I remember that the express middle in RackHD had handle it at https://github.com/RackHD/on-http/blob/master/lib/services/rest-api-service.js#L315 , if something is wrong, please correct me.

jlongever:
@zyoung51 assuming unit-test failures are due to on-core dependency: https://github.com/RackHD/on-core/pull/88.

+1 when conflicts resolved.

lib/services/nodes-api-service.js
brianparry: @anhou Similarly, error handling for the 2.0 APIs is here: https://github.com/RackHD/on-http/blob/master/lib/fittings/json_error_handler.js.  The 2.0 API will catch these errors in the swagger API service here: https://github.com/RackHD/on-http/blob/master/lib/services/swagger-api-service.js#L52.  Errors are handed off to the error handler fitting by calling `next(err)`.

brianparry: @zyoung51 I believe these assertion failures will turn into HTTP 400 responses.  I'm not sure if that's what you intended, but it seems like we might want to return HTTP 500 if any of these asserts fail.

jlongever:
test this please

jlongever:
Closing this PR,  new PR here resolving merge conflicts https://github.com/RackHD/on-http/pull/154

PR #136, opened 2016-02-18
Redact encrypted fields in OBM and SNMP serializer

geoff-reid:
@srinia6 @dalebremner @abildw @RackHD/corecommitters

### DIFF
--- a/lib/serializables/v1/obm.js
+++ b/lib/serializables/v1/obm.js
@@ -53,7 +53,7 @@ function ObmFactory (
             switch (key) {
                 case 'password':
                 case 'community':
-                    object[key] = encryption.decrypt(value);
+                    object[key] = 'REDACTED';
                     break;
 
                 default:
--- a/lib/serializables/v1/snmp.js
+++ b/lib/serializables/v1/snmp.js
@@ -35,7 +35,7 @@ function SnmpFactory (Promise, Serializable, encryption) {
 
     Snmp.prototype.serialize = function (target) {
         this.defaults(target);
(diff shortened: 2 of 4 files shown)


abildw:
+1

heckj:
@geoff-reid I thought there was somewhere that we were doing auto-redaction in middleware... never mind, I was wrong, but so you're aware of it - here's what we did for redacting this kind of information from the event logs. https://github.com/RackHD/on-core/blob/master/lib/serializables/log-event.js#L144-L162

I don't know if you'd want to follow that pattern of having what values you explicitly redact in constants or not. The "serializables" construct was something Frey set up ages ago, but was never applied to any of the HTTP endpoints, so it may not be worth the trouble, especially as we get into the swagger and redfish spaces.

benbp:
:+1:

yyscamper:
I think we should move the Redact functionality to a common service in on-core, neither like our current design in LogEvent nor implement a brand new one in on-http, so that it can be shared.

brianparry:
@yyscamper: While I think a common mechanism for redacting passwords is a good idea, it's outside the scope of this PR.  The idea here is to implement redaction for OBMs as quickly as possible to eliminate the possibility of revealing passwords.  I'd suggest we open a new google groups thread to discuss a more general solution.

@geoff-reid  :+1:

brianparry:
@yyscamper Looks like the pollers route also has passwords that should be redacted: https://github.com/RackHD/on-http/pull/134.

PR #138, opened 2016-02-18
Updating ubuntu ipxe and preseed files

ytjohn:
* Replaced /ubuntu_trusty (which seems to be a personal mirror a dev setup) with correct locations that will work when /ubuntu is proxying http://archive.ubuntu.com/ubuntu/
* Removed the squashfs referenced in the preseed. It is not needed, and attempting to inject one creates a fragile dependency.
* Added the port to the host string.


benbp:
:+1: thanks!

heckj:
:+1: loving it! @leoyjchang @yyscamper @anhou FYI, give you a bit of a jumpstart...

zyoung51:
LGTM +1

anhou:
:+1: Thanks Joe

larry-dean:
@stuart-stanley

PR #140, opened 2016-02-19
[redfish] Validate objects using odata.type if it is supplied

zyoung51:
- The object is parsed and all objects within that contain an
  'odata.type' are extracted.  Each odata.type is looked up in
  the schema dataset and then the object is validated against
  the schema.


jlongever:
+1

anhou:
:+1:

PR #144, opened 2016-02-22
[skupack] Add configuration/option import and rendering

zyoung51:
- Import skuConfig object from config.json into environment collection
- Add 'sku' and 'env' to profile/template rendering objects

### DIFF
--- a/lib/services/common-api-presenter.js
+++ b/lib/services/common-api-presenter.js
@@ -19,7 +19,8 @@ di.annotate(CommonApiPresenterFactory,
         'Services.Lookup',
         'Profiles',
         'Templates',
-        '_'
+        '_',
+        'Services.Environment'
     )
 );
 
@@ -39,7 +40,8 @@ function CommonApiPresenterFactory(
     lookupService,
     profiles,
     templates,
-    _
+    _,
+    Env
 ) {
     var logger = Logger.initialize(CommonApiPresenterFactory);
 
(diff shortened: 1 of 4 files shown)


lib/services/common-api-presenter.js
```diff
@@ -209,28 +211,24 @@ function CommonApiPresenterFactory(
...
-                lookupService.ipAddressToMacAddress(self.res.locals.ipAddress),
-                templates.get(name, scope)
+                Promise.props({
+                    server: configuration.get('apiServerAddress', '10.1.1.1'),
+                    port: configuration.get('apiServerPort', 80),
+                    ipaddress: self.res.locals.ipAddress,
+                    netmask: configuration.get('dhcpSubnetMask', '255.255.255.0'),
+                    gateway: configuration.get('dhcpGateway', '10.1.1.1'),
+                    macaddress: lookupService.ipAddressToMacAddress(self.res.locals.ipAddress),
+                    sku: Env.get('config', {}, [ scope[0] ]),
```
heckj: is scope guaranteed to be an array/list? Looks like it is in `lib/services/http-service.js` but wanted to check to see if it was fiddled with elsewhere...

heckj:
question about where scope may be twiddled, but :+1: regardless

lib/services/common-api-presenter.js
zyoung51: The array is default to an array in http-service at the start of the middleware and stays an array here: https://github.com/RackHD/on-http/pull/144/files#diff-51cccc923efc7fc45ae2bcf6846680c5R57

yyscamper: What the `'config'` here means? Only reads those configuration entries from sku pack's "skuConfig" definition? User may confuse it is from RackHD configuration file (monorail.json).  

Meanwhile, why only use the scope[0]? It is the SKU scope? I think it's better to have a comment here.

lib/services/common-api-presenter.js
```diff
@@ -39,7 +40,8 @@ function CommonApiPresenterFactory(
     lookupService,
     profiles,
     templates,
-    _
+    _,
+    Env
```
yyscamper: Since `Env` is an instance (https://github.com/RackHD/on-core/blob/master/lib/services/environment.js#L82), so as convention, its name should begin with lowercase. (if it is a Class, it should begin with uppercase) So change it from `Env` to `env`

jlongever:
test this please

jlongever:
test this please

jlongever:
@zyoung51 the test_list_simple_storage test fails pretty consistently,  it's failing to look up smart catalog source.  Could you take a look? 

test this please

jlongever:
+1

keedya:
:+1:

PR #145, opened 2016-02-24
Workflow related API fixes

benbp:
More smoke test fixes:

- Missed one occurrence of replacing the deprecated way to check for an active graph for a node.
- Making some changes to the workflow library routes. These routes *already* do not work (https://github.com/RackHD/RackHD/issues/89), but tests have passed with false positives so far because the `/workflows/library` route gets redirected to the `/workflows/<id>` route. Previously, this route improperly returned a 200 no matter what. Because it was changed to return a 404 on failure to find a workflow, a bunch of the false positive tests are now failing as they should be.

I think this may be controversial in regards to the `/workflows/library/*` addition/change, but I'm kind of at a loss for what to do. I don't want to continue with the current broken behavior just to keep incorrect tests passing.

Resolves https://github.com/RackHD/RackHD/issues/89

@RackHD/corecommitters @jlongever @heckj @VulpesArtificem

### DIFF
--- a/lib/api/1.1/northbound/workflows.js
+++ b/lib/api/1.1/northbound/workflows.js
@@ -104,34 +104,25 @@ function workflowsRouterFactory (
     }));
 
     /**
-     * @api {get} /api/1.1/workflows/library GET /library
+     * @api {get} /api/1.1/workflows/library/:injectableName GET /library/:injectableName
      * @apiVersion 1.1.0
-     * @apiDescription get list of workflows possible to run
-     * @apiName workflows-library
-     * @apiSuccess {json} workflows list of workflows that are possible to run
-     * @apiGroup workflows
-     */
-
-    router.get('/workflows/library', rest(function () {
-        return workflowApiService.getGraphDefinitions();
-    }));
-
-    /**
-     * @api {get} /api/1.1/workflows/library/:identifier GET /library/:identifier
-     * @apiVersion 1.1.0
-     * @apiDescription get a specific workflow from the library
-     * @apiParam {String} identifier of workflow, must cast to ObjectId
+     * @apiDescription get a specific workflow definition from the library. To get all
+     *                 known workflow definitions, use the '*' wildcard (GET /library/*)
+     * @apiParam {String} injectableName of workflow definition
      * @apiName workflows-library-item
-     * @apiSuccess {json} workflow the workflow that have the <code>identifier</code>
+     * @apiSuccess {json} workflow the workflow that have the <code>injectableName</code>
      * @apiGroup workflows
      */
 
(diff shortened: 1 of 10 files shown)


jlongever:
+1

anhou:
:+1:

lib/api/1.1/northbound/workflows.js
```diff
@@ -104,34 +104,24 @@ function workflowsRouterFactory (
...
      * @apiName workflows-library-item
-     * @apiSuccess {json} workflow the workflow that have the <code>identifier</code>
+     * @apiSuccess {json} workflow the workflow that have the <code>injectableName</code>
      * @apiGroup workflows
      */
 
-    router.get('/workflows/library/:identifier', rest(function (req) {
-        return workflowApiService.getGraphDefinitions(req.params.identifier)
+    router.get('/workflows/library/:injectableName', rest(function (req) {
+        if (req.params.injectableName === '*') {
```
yyscamper: Whether it means the `get /workflows/library` will never work well, instead, we need to use `get /workflows/library/*`? But this seems not a good design.

jlongever: probably should be two routes:  GET /workflows/library and GET /workflows/library/:injectableName

heckj: In terms of good API design, you're right - we got ourselves into a pickle with this setup, and there's no clean way to escape it with the 1.1 API that doesn't do something to muck up backwards compatibility. The optimal place to resolve this will be the 2.0 API that @brianparry is working on. As a workaround I'm fine with this with updated documentation - maybe an honorable mention in the APIdoc strings above for how to handle what we'd previously intended with a `GET /workflows/library` and relevant notes in the user's guide as well. 

@jlongever if I'm understanding correctly, the two routes is what we intended, but the system wasn't routing the calls to the relevant control code correctly, or routing them to two places at one, which is what we were seeing in #89, but hadn't stumbled over explicitly until we dug out the bugs with the workflow refactor knowledge.

heckj:
:+1:  with some tacked-in documentation - to the API strings in particular since we're removing an API in a `1.1` API release, and notes in the user's guide documentation under workflows? This is probably something worthwhile to call out in a short email to the Corsair team and mailing list with this change - since it's a longer-standing bug in the 1.1 API that we can't seem to really resolve with breaking some level of backwards compatibility. I'm pretty sure this will impact the BOSH CPI team.

FYI: @kaleo211 @meganmurawski @oskoss  - I recall seeing the `GET /workflows/library` call in your code, so wanted to make sure you had a specific head's up on this.

FYI: @king-jam @jfrey @justink0

jlongever:
Function scripts pass with all dependencies locally.
+1 !

PR #146, opened 2016-02-24
add external repo support for CoreOS installation

yyscamper:
This is discussed in this link: https://github.com/RackHD/RackHD/issues/90
Since http proxy often fails for large file transmission especially in vagrant environment, so add the external repo support will provide alternative way to do CoreOS installation.

@RackHD/corecommitters @pengz1 @heckj

### DIFF
--- a/data/profiles/install-coreos.ipxe
+++ b/data/profiles/install-coreos.ipxe
@@ -1,5 +1,5 @@
 echo Starting CoreOS Stable installer
-set base-url http://<%=server%>:<%=port%>/coreos
+set base-url <%=repo%>
 # coreos.autologin <-- enable as kernel param to autologin for easier debugging
 # reference: https://coreos.com/os/docs/latest/booting-with-ipxe.html
 # JUST BOOT AND RUN COREOS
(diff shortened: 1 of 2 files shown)


heckj:
:+1:

heckj:
:+1: @RackHD/corecommitters question related to this kind of change - how should we add documentation to existing structures like this to share that we're made new variables like this that need to be accommodated by any workflows invoking them?

yyscamper:
@heckj : I think this is not a kind of new variable, because CentOS/RHEL/ESXi has already supported such option. I consider this as a technical debt. We need to sync our all OS bootstrap design, I see we already have an issue opened for this: https://github.com/RackHD/RackHD/issues/50. If this issue is resolved, then a common document will cover all OS bootstrap for these common features.

In future, we may find some feature is hard to be added into other OS, like CoreOS may not support kvm, and Other OS excepts ESXi cannot support to configure the virtual switch, so we do need a separated document for each OS to describe those special features.

anhou:
+1

yyscamper:
test this please

jlongever:
Fyi..The unit test fixes included in this PR https://github.com/RackHD/on-http/pull/157 resolves random port conflicts introduced in the 1.1 authentication commits.

yyscamper:
@jlongever , thanks for this info, so I can ignore the Jenkins failure.

jlongever:
test this please

WangWinson:
:+1:

PR #147, opened 2016-02-24
added fixes and tests for diffrent content types being passed to prof…

DavidjohnBlodgett:
…ile and template PUT requests

@RackHD/corecommitters @rolandpoulter @tldavies @VulpesArtificem

### DIFF
--- a/lib/api/1.1/northbound/profiles.js
+++ b/lib/api/1.1/northbound/profiles.js
@@ -2,9 +2,9 @@
 
 'use strict';
 
-var di = require('di'),
-    express = require('express'),
-    parser = require('body-parser');
+var di = require('di');
+var express = require('express');
+var parser = require('body-parser');
 
 module.exports = profilesRouterFactory;
 
@@ -75,13 +75,21 @@ function profilesRouterFactory (
      *     }
      */
 
-    router.put('/profiles/library/:identifier', parser.raw(), function (req, res) {
-        profiles.put(req.params.identifier, req.body.toString()).then(function () {
-            res.status(200).send();
-        }).catch(function (error) {
-            res.status(500).json(error);
+    router.put('/profiles/library/:identifier',
+        function (req, res, next) {
+            if (req.get('Content-Type') === 'application/octet-stream') {
+               parser.raw({type: '*/*'})(req, res, next);
+            } else {
+                parser.text({type: '*/*'})(req, res, next);
+            }
+        },
+        function (req, res) {
+            profiles.put(req.params.identifier, req.body.toString()).then(function () {
+                res.status(200).send();
+            }).catch(function (error) {
+                res.status(500).json(error);
+            });
         });
-    });
 
     return router;
 }
--- a/lib/api/1.1/northbound/templates.js
+++ b/lib/api/1.1/northbound/templates.js
@@ -2,9 +2,9 @@
 
 'use strict';
 
-var di = require('di'),
-    express = require('express'),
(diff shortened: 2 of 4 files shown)


lib/api/1.1/northbound/profiles.js
```diff
@@ -4,7 +4,7 @@
 
 var di = require('di'),
     express = require('express'),
-    parser = require('body-parser');
+    ambiguous_parser = require('./../../../fittings/ambiguous_text_parser');
```
benbp: What is this and where does it come from?

lib/api/1.1/northbound/profiles.js
```diff
@@ -75,7 +75,9 @@ function profilesRouterFactory (
      *     }
      */
 
-    router.put('/profiles/library/:identifier', parser.raw(), function (req, res) {
+    //router.put('/profiles/library/:identifier', parser.text("*/*"), function (req, res) {
```
benbp: Delete this?

lib/api/1.1/northbound/profiles.js
DavidjohnBlodgett: I added a function to decide what parser should be used based on the content-type of the request... I had to do this because parser.raw() fails to achieve the desired result when using 'text/*' content-types, and parser.text() fails when you have 'application/octet-stream' content-type... so both were context specific. That being said, I would have had to feed this function in to both PUT definitions, both profiles.js and templates.js so it was suggested by @rolandpoulter to abstract this function call.

lib/fittings/ambiguous_text_parser.js
```diff
@@ -0,0 +1,11 @@
+parser = require('body-parser');
+
+module.exports = {
+  parse: function (req, res, next) {
+      if (req.get('Content-Type') === 'application/octet-stream') {
+          parser.raw()(req, res, next);
+      }else {
+          parser.text({type: 'text/*'})(req, res, next);
```
rolandpoulter: `var parser = require('body-parser');

lib/api/1.1/northbound/profiles.js
```diff
@@ -2,9 +2,9 @@
 
 'use strict';
 
-var di = require('di'),
-    express = require('express'),
-    parser = require('body-parser');
+var di = require('di');
+var express = require('express');
+var ambiguousParser = require('./../../../fittings/ambiguous_text_parser');
```
benbp: I'm not too excited about the deep relative path here, partly because it circumvents how we do dependency management. But I know we're also going in a direction with the 2.0 API that can't use DI because of swagger. And is fittings the right place for this? @brianparry do you have any thoughts here?

benbp: I'm almost inclined to suggest just embedding the middleware function into both profiles.js and templates.js, since it's dead simple and only used in two places, but I don't know how that will play with the 2.0 API transition either.

DavidjohnBlodgett: @benbp I'm ok with embedding it, if thats the direction we want to go with it. :)

DavidjohnBlodgett:
@RackHD/corecommitters @rolandpoulter @tldavies @VulpesArtificem addressed feedback and fixed issues. I think this is merge ready.

rolandpoulter:
:+1:

VulpesArtificem:
:+1:

benbp:
:+1:

PR #150, opened 2016-02-25
Swagger 2.0 Workflow Routes

dalebremner:
@RackHD/corecommitters @srinia6 @geoff-reid @abildw


heckj:
semantically, what's the intention with our URI scheme for adding, updating, or even deleting workflow definitions vs. invoking a workflow and being able to retrieve lists of workflows that have run in the past (or are currently running) to understand their current or historical state?

benbp:
I've actually got some feedback on breaking changes we need/should make to these route structures, since they are not so great right now. Let me put that feedback together please before anyone merges this (I'll do it today).

benbp:
So here are a couple of notes, perhaps we should move this conversation outside of the PR if it gets too big. The general point is we really don't want to make a 1:1 copy of the 1.1 workflows API because the 1.1 workflows API is not well designed.

We need more semantic differences in the route structure than what we have in 1.1, in order to make it clear which routes are for workflow definitions, which are for active workflows, and which are for creating active workflows.

For example, right now we have: `PUT /workflows` and `POST /workflows` which is kind of confusing, because POST is for running a workflow and PUT is for defining one. It would be better as something like `PUT /workflows/definitions` and `POST /workflows` or something.

PR #156, opened 2016-02-26
Don't return a 404 when there is no active workflow for a node

benbp:
At the request of @stuart-stanley. A 404 indicates an illegal request, most likely that the node ID doesn't exist at all, as compared to a legitimate request that just returns empty (e.g. are there active workflows? no).

@RackHD/corecommitters

### DIFF
--- a/lib/api/1.1/northbound/nodes.js
+++ b/lib/api/1.1/northbound/nodes.js
@@ -391,8 +391,6 @@ function nodesRouterFactory (
      * @apiParam {String} identifier node identifier
      * @apiSuccess {json} workflow  the active workflow for that node.
      * @apiError NotFound1 The node with the <code>identifier</code> was not found.
-     * @apiError NotFound2 The node with the <code>identifier</code> does not have an
-     *                     active workflow.
      * @apiErrorExample Error-Response:
      *     HTTP/1.1 404 Not Found
      *     {
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -431,13 +431,6 @@ function nodeApiServiceFactory(
         return waterline.nodes.needByIdentifier(id)
         .then(function (node) {
             return workflowApiService.findActiveGraphForTarget(node.id);
-        })
-        .then(function(graph) {
-            if (_.isEmpty(graph)) {
-                throw new Errors.NotFoundError('Not Found');
-            } else {
-                return graph;
-            }
         });
     };
 
--- a/spec/lib/services/nodes-api-service-spec.js
+++ b/spec/lib/services/nodes-api-service-spec.js
@@ -268,8 +268,7 @@ describe("Http.Services.Api.Nodes", function () {
         it('should throw a NotFoundError if the node has no active graph', function () {
             waterline.nodes.needByIdentifier.resolves({ id: 'testid' });
             findActiveGraphForTarget.resolves(null);
-            return expect(nodeApiService.getActiveNodeWorkflowById('test'))
-                    .to.be.rejectedWith(Errors.NotFoundError);
+            return expect(nodeApiService.getActiveNodeWorkflowById('test')).to.become(null);
         });
     });
 


benbp:
@heckj FYI

heckj:
`404` is an illegal request? Are you confusing a `400` response with `404`? (ref: https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) @stuart-stanley ?

stuart-stanley:
@heckj this came out of a slack conversation. I wish I could find the bleeping page I saw the put-way-better-than-I-will-here discussion of this, but the nutshell is that something like:
GET node/<nid>/workflows/active ends up being the most useful if:
* it returns 404 if the nid does not exist.
* it returns a blank list/204 if the _node_ is there, but there doesn't happen to be an active workflow at the moment. 

It gets a little muddy, because "active" isn't a list, of course.

As an FYI (very very much FYI/ramblings, but hopefully interesting), the view on "active" would be that it is a "short cut" (or utility resource): it is a handy way to do GET node/<id>/workflows and then scan the resulting list for the active one. Well, a handy way to NOT have to do that ;). But the more complete API would look like:
* GET node/<id>/workflows -> gets entire list both past, present, and future. Ok, kidding on the future.
* GET node/<id>/workflows/active -> get the current active list (with list here being [] or [1]).
* DELETE node/<id>workflow/<work-flow-id> -> delete that workflow:
  * 404 if never-existed work-flow
  * 410 if already deleted (this presumes we keep all completed workflows and can tell if they have been deleted, of course).
 and so on.
What there isn't is a DELETE node/<id>/workflows/active... that ends up being dangerous, since this can occur:
* I look at GET node/<id>/workflows/active and see workflow "foo", which I want to kill.
* workflow foo completes and workflow bar starts
* I DELETE node/<id>/workflows/active and blow away bar. ooops...

benbp:
@stuart-stanley I'm kind of flip-flopping now from our conversation, but isn't the active workflow treated as its own resource, in which case a 404 makes sense because that resource is not found?

heckj:
@benbp @stuart-stanley That's where I was coming from - that the URI down to the workflow represented an indepdent resource. Clearly this would have been more sane with the URI's as parallel, rather than subservient hierarchical resources, but I was still leaning towards tracking them as independent. If it's unclear if the node itself exists or not, then a URI check to /api/1.1/nodes/[id] would (to me) be the discerning factor by returning a 404 explicitly for that.

This is something we can definitely do better with API 2.0, especially as we're heading to divorcing workflows from being node specific (well, already have, but the 1.1 URI structure doesn't really reflect that). There's some active conversation/feedback going on know for the 2.0 API structure and it's semantics that differentiate workflow definitions from workflows run and workflow's current state when running. All need to be accounted for.

I'm hoping that in the future, something like /nodes/workflows would be a convenience only/shortcut to know about workflows per a node, and that something like /nodes/workflows/active wouldn't even need to specifically exist, and would instead be referenced by a workflow ID if you wanted to be clear about history or current state of a specific workflow.

cc. @brianparry (since you've been dipping into the 2.0 semantics and where we can improve things)

stuart-stanley:
Agreed on looking forward to 2.0.

I think the big thing with the GET */active returning 404 comes down to being nice to the caller (or replace "nice" with communicating well with): If I do a GET node/<id>/workflows/active and I get an error, I don't know without hunting if thats a totally normal condition (the workflow finished before I could delete it) or a real problem (my node-id doesn't exist or ceased to exist or something). To ground this a little bit, we did have a case at one point where a full node would vanish on us during the "wait for fully discovered" part of smoke. To be fair, it was a kind of stupid case, but it happened. 

So, we end up having a loop that polls for the active workflow to be gone. Oh good, we got a 204, it's done. If a 404 came back it told us ERROR! ERROR! ERROR! DANGER! Now we see a 404 and go "oh, I wonder if this is a real problem or not?"

All of that ties back to */active being a pointer to something else. I mean, call it three times and you may get three radically different resources back, _each with a different resource-unique-id_. That's the hint @benbp that tells you this resource is a short-cut resource vs a "real" end-point. Does that make more sense?

johren:
The difference to me was something analogous to telling my messenger to go pick up a package at 110 Main Street.

He/she may come back to me and say, "I went to 110 Main Street, but they had no package to pick up."  (204)
or
He/she may come back to me and say, "I could not find any building at 110 Main Street, are you sure you gave me the right address?"  (404)


Wouldn't that be the same thing here?  "That node had no active workflow" vs. "I have no idea what node that is"

johren:
According to the HTTP Code Definition (/Protocols/rfc2616/rfc2616-sec10.html), 404 is returned when
"The server has not found anything matching the Request-URI."  So doesn't that mean that 
/api/1.1/nodes/12345/workflows/active would return 204 for a valid node with ID 12345, but
/api/1.1/nodes/54321/workflows/active (assuming node 54321 doesn't exist) or /api/1.1/nodes/12345/workflows/actiev would return 404?

johren:
Okay, one last argument...A 404 means a "Client-Side Error".  Is it the error of the client that they asked for active workflow for a valid node but there were none?  I can see in the case that the node doesn't exist because they can get the list of node IDs before they request the active workflow.  But they have no idea until they make the API call for the active workflow if it exists or not.

yyscamper:
I prefer to 204, 404 is client error, it is a kind of error request. But no active workflow means nothing to response but user's request is valid.

heckj:
I get your point about `204`, but fundamentally a 2xx response to me indicates a success, when asking for "the active workflow for a node" (if none exists) doesn't semantically appear correct to me. All of the points above are good and valid, and even more reinforcement that the resources should have been handled more in a parallel fashion than hierarchical/subservient, which leads directly to this confusion.

My preference and earlier insistence on 404 for this is that I read the entire URI as the resource descriptor address that was either there (or missing), thinking of it more like a database's composed key construct - where the entire URI was the definitive resource location (even with all the kind of stupidity that entails, as "current" doesn't always mean the same workflow from one invocation to the next even in short order). I personally feel 404 is more generally representative here in the space of overall REST APIs.

This isn't a new argument at all, and there's good arguments for both sides. I'll reference a couple of topics from StackOverflow:
 - http://stackoverflow.com/questions/26845631/is-it-correct-to-return-404-when-a-rest-resource-is-not-found
 - http://stackoverflow.com/questions/2195639/restful-resource-not-found-404-or-204-jersey-returns-204-on-null-being-returne
 - http://stackoverflow.com/questions/11746894/what-is-the-proper-rest-response-code-for-a-valid-request-but-an-empty-data
 - http://blog.ploeh.dk/2013/04/30/rest-lesson-learned-avoid-204-responses/
 - http://programmers.stackexchange.com/questions/203492/when-to-use-http-status-code-404-in-an-api

(these were top hits from `https://www.google.com/search?client=safari&rls=en&q=http+response+204&ie=UTF-8&oe=UTF-8#q=http+response+204+vs+404` in google search)

I will absolutely defer to the @RackHD/corecommitters in making a consistency ruling on this if you all think the 404 response needs to be changed for the existing 1.1 API

benbp:
@heckj given that the 1.1 API *was* returning a 204 until I changed it to a 404 with the workflow refactor, perhaps the tiebreaker here is making fewer breaking changes to the 1.1 API and restoring the 204 response. In the 2.0 API we can get around this debate by having a better structure as you mention. @brianparry ?

brianparry:
@benbp @heckj @stuart-stanley A few thoughts here:

My interpretation of 404 is: "the thing you were looking for isn't here."
And 204 is: "your request was successful but there is no applicable data to return in the response body"

404 makes sense to me only when requesting a _specific_ thing (i.e. /nodes/workflows/:Identifier).  In the case of /nodes/workflows/active, the client is not requesting anything specific, but rather is asking if any active workflow exists.  It's sort of a shorthand for something like GET:/nodes/workflows?state=active so I think it makes sense to either return an empty collection or 204.

For the 2.0 API, I'd propose we change the semantics to use the query string approach and return a 200 and an empty collection for any query that finds no matching resources.

heckj:
Sounds like a plan, I like the thought process. Let's revert this back to 204 response as it was originally, and in the future (V2 API) lets run with a 200 response and empty list representing nothing in that sequence/query string response.

benbp:
kk, I still need the +1s though :)

@heckj @RackHD/corecommitters

heckj:
Uh, duh - sorry :+1:

jlongever:
+1

PR #157, opened 2016-02-28
Add dynamic and static IP/MAC lookups

jlongever:
- Option to remove DHCP-Proxy dependency (example dhcpd.conf PR https://github.com/RackHD/RackHD/pull/114) and decouple DHCP service/least file tailing.
- Creates lookup entry based on DHCP bootfile option request query parameters
- Fixed unit-tests to use unique http/https ports that are causing random ADDRINUSE errors

@RackHD/corecommitters @geoff-reid @zyoung51 @uppalk1 @tannoa2 @anhou

### DIFF
--- a/lib/api/1.1/southbound/profiles.js
+++ b/lib/api/1.1/southbound/profiles.js
@@ -27,6 +27,11 @@ function profilesRouterFactory (
     /**
      * @api {get} /api/1.1/profiles GET /
      * @apiVersion 1.1.0
+     * @apiParam {query} macs List of valid MAC addresses to lookup
+     * @apiParam {query} mac When macs parameter is not passed, 
+     *                       passed with IP adds MAC address to lookup
+     * @apiParam {query} ip When macs parameters is not passed, 
+     *                      passed with MAC adds IP address to lookup
      * @apiDescription used internally by the system -- will NOT get a list of all profiles,
      *                 look at api/current/profiles/library
      * @apiName profiles-get
--- a/lib/api/1.1/southbound/profiles.js
+++ b/lib/api/1.1/southbound/profiles.js
@@ -31,9 +31,13 @@ function profilesRouterFactory (
      */
 
     router.get('/profiles', function (req, res) {
+        if (req.query.mac && req.query.ip) {
+            profileApiService.setNode(req.query);
+            req.query.macs = req.query.mac
+        }
+        
         if (req.query.macs) {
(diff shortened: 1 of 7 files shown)


keedya:
:+1:

lib/api/1.1/southbound/profiles.js
```diff
@@ -31,9 +31,13 @@ function profilesRouterFactory (
      */
 
     router.get('/profiles', function (req, res) {
+        if (req.query.mac && req.query.ip) {
+            profileApiService.setNode(req.query);
+            req.query.macs = req.query.mac;
+        }
+
```
benbp: Should put a catch either here or in the service otherwise we'll lose any errors that occur with this call.

lib/services/profiles-api-service.js
```diff
@@ -61,6 +63,18 @@ function profileApiServiceFactory(
         return _.flattenDeep([macs]);
     };
 
+    ProfileApiService.prototype.setNode = function(query) {
```
benbp: This function name doesn't really make sense to me? Since you're setting Mac and Ip and not node Id?

benbp:
:+1: if you add a catch block/log.

lib/api/1.1/southbound/profiles.js
```diff
@@ -31,9 +31,13 @@ function profilesRouterFactory (
      */
 
     router.get('/profiles', function (req, res) {
+        if (req.query.mac && req.query.ip) {
```
anhou: Could you help move the description like 'Creates lookup entry based on DHCP bootfile option request query parameters' to the annotations here? the message in PR  wouldn't be included in codes, and annotations here will help a lot for code readablility

anhou:
:+1: after addressing the comments

jlongever:
@benbp

lib/api/1.1/southbound/profiles.js
jlongever: @anhou  Are you looking for comments?  There are other routes missing query parameter annotations;  maybe there should be a story to track updating those as well.

anhou: no, not for the routes, but for if() {... self.lookup...} 's

jlongever: Right,  there are a lot of route code that reference query parameters that offer no description.

jlongever:
@anhou added query param description

anhou:
@jlongever Thanks for refinement!

PR #160, opened 2016-02-29
Clone workflow definitions before validating them on upload

benbp:
So, workflow definition uploading is broken without this fix because task graph creation/validation made edits to the definition by reference, turning task waitOn values into UUIDs.

@RackHD/corecommitters @VulpesArtificem


benbp:
Actually, this is better in on-taskgraph. We used to do it there but it was changed to `_.defaultsDeep`. Closing this in favor of a forthcoming change there.

PR #164, opened 2016-03-02
Improve the http proxy performance for big file transimission

yyscamper:
The issue link is: https://github.com/RackHD/RackHD/issues/90, and some discussion is in the google group: https://groups.google.com/forum/#!topic/rackhd/NAKw1ZXixlQ

This fix is to replace the old proxy library `express-http-proxy` with `http-proxy-middleware`, because `express-http-proxy` has bad performance for big file transmission, because it will first cache all content into memory and then passed to requester. But `http-proxy-middleware` works in a stream way,  the content is gradually passed to the requester, this will dramatically require less memory usage.

I have tested in vagrant environment (with 512M memory), use the `http-proxy-middleware` succeed to download a 3GB file via proxy. But the `express-http-proxy` will make the ORA hangs even I tried to increase the memory to 2GB.

@RackHD/corecommitters @pengz1 @iceiilin @WangWinson @tedstockwell @sunnyqianzhang

### DIFF
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -342,19 +342,19 @@ function httpServiceFactory(
                 serverAddr = httpProxy.server;
 
             //localPath & remotePath should always starts with /
-            if (!_.startsWith(localPath, sep)) {
+            if (!localPath.startsWith(sep)) {
                 localPath = sep + localPath;
             }
-            if (!_.startsWith(remotePath, sep)) {
+            if (!remotePath.startsWith(sep)) {
                 remotePath = sep + remotePath;
             }
 
-            //localPath & remotePath should both ends with / or not, otherwise the pathRewite will
+            //localPath & remotePath should both ends with / or not, otherwise the pathRewrite will
             //not work well
-            if (_.endsWith(remotePath, sep) && !_.endsWith(localPath, sep)) {
+            if (remotePath.endsWith(sep) && !localPath.endsWith(sep)) {
                 localPath = localPath + sep;
             }
-            else if (!_.endsWith(remotePath, sep) && _.endsWith(localPath, sep)) {
+            else if (!remotePath.endsWith(sep) && localPath.endsWith(sep)) {
                 remotePath = remotePath + sep;
             }
 
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -9,7 +9,7 @@ var onFinished = require('on-finished');
 var http = require('http');
 var https = require('https');
 var fs = require('fs');
-var proxy = require('express-http-proxy');
+var proxyMiddleware = require('http-proxy-middleware');
 var path = require('path');
 var bodyParser = require('body-parser');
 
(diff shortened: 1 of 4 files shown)


spec/lib/services/obm-api-service-spec.js
```diff
@@ -6,7 +6,6 @@
 describe("Http.Services.Api.Obms", function () {
     var obmService;
     var Promise;
-    var chai = require('chai');
     chai.use(require('chai-string'));
```
yyscamper: Remove this line is just to fix the unit-test error, the unit-test will fail with following error information if keep this line:
```
  1) Http.Services.Api.Obms getObmLib Run getObm:
     TypeError: Object [object Object] has no method 'equalIgnoreCase'
      at Context.<anonymous> (spec/lib/services/obm-api-service-spec.js:205:40)

  2) Http.Services.Api.Obms getObmLibById Run getObmLibById:
     TypeError: Object [object Object] has no method 'equalIgnoreCase'
      at Context.<anonymous> (spec/lib/services/obm-api-service-spec.js:225:40)
```

If remove this line, the error disappears. I don't understand why at present. But anyway, I  think the best is to move the [chai-string](http://chaijs.com/plugins/chai-string/) to on-core/spec/helper.js, As I see the chai-string is good for string assertion and potentially will be used by other repos.

package.json
```diff
@@ -41,8 +42,7 @@
     "swagger-tools": "~0.9.11",
     "tar": "~2.2.1",
     "tv4": "~1.2.7",
-    "ws": "^0.8.0",
-    "debug": "^2.2.0"
+    "ws": "^0.8.0"
```
yyscamper: I move these library in alphabetic order

package.json
```diff
@@ -18,12 +18,13 @@
   "dependencies": {
     "body-parser": "^1.10.1",
     "cors": "^2.5.2",
+    "debug": "^2.2.0",
```
jlongever: why was debug moved to dependencies?

jlongever: never mind,  you moved it up.

spec/lib/services/obm-api-service-spec.js
yyscamper: Just open a PR for on-core to move `chai-string` to on-core: https://github.com/RackHD/on-core/pull/94

lib/services/http-service.js
```diff
@@ -335,19 +335,41 @@ function httpServiceFactory(
...
-                    return path.join(remotePath, require('url').parse(req.url).path);
-                },
-                decorateRequest: function (req) {
-                    //For HTTPS access
-                    req.rejectUnauthorized = false;
-                }
-            }));
+
+            //localPath & remotePath should always starts with /
+            if (!_.startsWith(localPath, sep)) {
```
benbp: FYI We also support the ES6 native `String.startsWith` method because of the es6-shim library within di.js (https://github.com/RackHD/di.js/blob/master/package.json#L16)

lib/services/http-service.js
```diff
@@ -335,19 +335,41 @@ function httpServiceFactory(
...
+                localPath = sep + localPath;
+            }
+            if (!_.startsWith(remotePath, sep)) {
+                remotePath = sep + remotePath;
+            }
+
+            //localPath & remotePath should both ends with / or not, otherwise the pathRewite will
```
benbp: "pathRewrite"

package.json
benbp: As far as I can tell, we're not even using this dependency. Can we just get rid of it?

benbp:
:+1:

benbp:
FYI @heckj

pengz1:
+1

lib/services/http-service.js
yyscamper: This is better, thanks for awareness,  I will change it.

lib/services/http-service.js
yyscamper: Got it, I'll change it.

PR #165, opened 2016-03-02
RackHD Issue #116

zyoung51:
- Fix ipv4 and ipv6 fields
- Add enum properties to osName
- Add required fields to json schema
- Update Task objects to align with workflow changes

### DIFF
--- a/static/DSP8010_1.0.0/json-schema-oem/RackHD.BootImage.json
+++ b/static/DSP8010_1.0.0/json-schema-oem/RackHD.BootImage.json
@@ -35,7 +35,14 @@
                     "type": "string"
                 },
                 "osName": {
-                    "type": "string"
+                    "type": "string",
+                    "enum": [
+                        "CentOS",
+                        "CentOS+KVM",
+                        "ESXi",
+                        "RHEL",
+                        "RHEL+KVM"
+                    ]
                 },
                 "repo": {
                     "type": "string"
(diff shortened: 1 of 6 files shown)


benbp:
:+1:

benbp:
@zyoung51 are these all the references to instanceId? If so I'm going to remove the `id` field from the graph objects again before anyone has the chance to use it again.

zyoung51:
@benbp This should be it in the redfish handlers.

benbp:
Awesome, thank you.

static/DSP8010_1.0.0/json-schema-oem/RackHD.BootImage.json
```diff
@@ -62,6 +69,13 @@
                     "type": "string"
                 }
             },
+            "required": [
+                "domain",
+                "hostname",
+                "osName",
+                "rootPassword",
+                "dnsServers"
+            ],
```
yyscamper: by RackHD design, all these fields are optional because all of them have a default value ( see https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/install-centos.js) , so we still want to enforce this required criteria in Redfish API?

zyoung51:
@yyscamper These are required items in the redfish versions that we are porting into RackHD.  Compatibility is a requirement, so they will stay required.

@RackHD/corecommitters Should we be using the defaults Felix linked above?  For example, a domain is defined but no DNS servers are set and the password is clear text.

jlongever:
+1

heckj:
FYI - when you make a PR that resolves an issue, you can put in:

`resolves https://github.com/RackHD/RackHD/issues/116` that will auto-close the issue on merge, so you don't need to manually close after the fact...

PR #158, opened 2016-02-29
Fix task graph running in SKU api

benbp:
Missed an update to change the graph running calls.

@RackHD/corecommitters @zyoung51

PR #166, opened 2016-03-03
Fix post sku pack error, change the regenerateSku to use latest workflow API

yyscamper:
When I tried to POST a new SKU pack, it will always fail. I find it is due to the regenerateSku is not changed together with @benbp 's workflow enhancement design. Previously, we use taskGraphRunner to trigger a workflow, but now we use a helper workflowApiService to do the same work.

@RackHD/corecommitters @zyoung51 @pengz1 @WangWinson @iceiilin @cgx027 @sunnyqianzhang

### DIFF
--- a/lib/api/1.1/northbound/skus.js
+++ b/lib/api/1.1/northbound/skus.js
@@ -15,9 +15,9 @@ di.annotate(skusRouterFactory, new di.Provide('Http.Api.Skus'));
 di.annotate(skusRouterFactory, new di.Inject(
         'Services.Waterline',
         'Http.Services.RestApi',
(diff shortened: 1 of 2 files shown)


yyscamper:
I just saw @benbp made the same change in this PR (https://github.com/RackHD/on-http/pull/158). So I close mine.

zyoung51:
Isn't this already fixed here: https://github.com/RackHD/on-http/pull/158

yyscamper:
yes, I just saw it, so I close mine.

Felix

-----Original Message-----

benbp:
Heh, good catch :)

PR #167, opened 2016-03-03
Work around npm 1.3.10 bug with prune.

johren:
npm prune --production doesn't work correctly in version 1.3.10.  It ends up removing the redfish-node module form on-tasks node_modules which is needed for on-http.

To work around this, we have to remove node_modules and run 'npm install --production' rather than do 'npm prune --production'.  We cannot just do 'npm install --production on line 14 because we need apidoc for line 15.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,12 @@ static/tftp/
 static/web-ui/
 
 build/
+
+# HWIMO-BUILD artifacts
+on-http_*.dsc
+on-http_*.tar.gz
+on-http_*.build
+on-http_*.changes
+on-http_*.deb
+packagebuild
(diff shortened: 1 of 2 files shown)


johren:
@davidzuhn @stuart-stanley @RackHD/corecommitters 

Please review ASAP.  This is blocking our development and testing on master and has had our smoke tests failing for the past week.

stuart-stanley:
+1 spooooooky.

jlongever:
+1

johren:
Thanks, Joel.  Can you merge this for me, please?

jlongever:
will do.
test this please

PR #170, opened 2016-03-04
[redfish] Add detailed error message responses

zyoung51:
- Add redfish detailed error message
- Change 500 on invalid object to 404 with redfish detailed error message
- Change 202 Accepted to 201 Created when Tasks are created
- Add redfish message registry
- Refactor redfish validator to separate schema into a separate service

### DIFF
--- a/data/templates/redfish.1.0.0.message.1.0.0.json
+++ b/data/templates/redfish.1.0.0.message.1.0.0.json
@@ -1,0 +1,12 @@
+{
+    "error": {
+        "code": "<%= code %>",
+        "message": "<%= message %>",
+        "@Message.ExtendedInfo": [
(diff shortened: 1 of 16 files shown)


brianparry:
@zyoung51 I'm thinking it might make sense to add an error handler callback argument to the `controller` factory function.  Th

```javascript
.catch(function(err) {
    if (errorHandler) {

});
```

zyoung51:
@brianparry This errorhandler could take additional arguments, so it would not eliminate all cases, but that would clean up most of them.  Another option that I explored, but decided not to use, would be to generate a redfish error with the message already formatted.

keedya:
:+1:

jlongever:
+1

PR #171, opened 2016-03-07
fix SKU Pack defineTask/defineTaskGraph error

yyscamper:
When inject the tasks/workflows of SKU pack, it will fail due to defineTask and defineWorkflow error. The root cause is that after @benbp 's workflow change, the previous TaskGraphRunner should be replaced with WorkflowApiService. This PR is to fix this problem together some other minor enhancement:

Changes include:
1. Fix SKU pack defineTask/defineTaskGraph error
2. remove unused taskGraphRunner inject (because it has been replaced with workflowApiService)
3. change POST the SKU pack returns a JSON object, this is the standard response format for all other APIs

@RackHD/corecommitters @zyoung51 @pengz1 @sunnyqianzhang

### DIFF
--- a/lib/services/common-api-presenter.js
+++ b/lib/services/common-api-presenter.js
@@ -300,7 +300,7 @@ function CommonApiPresenterFactory(
                             // If we failed to render the error then we've got larger problems.
 
                             logger.error('Unable to render error template.', {
-                                macaddress: macaddress,
+                                macaddress: options.macaddress,
                                 error: terrible
                             });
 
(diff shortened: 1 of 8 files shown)


lib/services/common-api-presenter.js
```diff
@@ -300,7 +300,7 @@ function CommonApiPresenterFactory(
                             // If we failed to render the error then we've got larger problems.
 
                             logger.error('Unable to render error template.', {
-                                macaddress: macaddress,
+                                macaddress: options.macaddress,
```
yyscamper: macaddress is `undefined` in current scope and even closure, so this deserve a fix.

zyoung51:
+1

benbp:
:+1:

PR #172, opened 2016-03-07
[skupack] Get pack information with sku information

zyoung51:

### DIFF
--- a/lib/services/sku-pack-service.js
+++ b/lib/services/sku-pack-service.js
@@ -371,6 +371,22 @@ function skuPackServiceFactory(
             });
     };
 
+    SkuPackService.prototype.getPackInfo = function(skuid) {
+        var self = this;
+        return self.loader.getAll(self.confRoot)
+        .then(function (conf) {
+            var skuConf = conf[skuid + '.json'];
+            if(skuConf) {
+                var json = JSON.parse(skuConf);
+                return {
+                    description: json.description || null,
+                    version: json.version || null
+                };
+            }
+            return null;
+        });
+    };
+
     function getObjectsWithKey(obj, key) {
         if (_.has(obj, key)) {
             return [obj];
--- a/spec/lib/api/1.1/skus-spec.js
+++ b/spec/lib/api/1.1/skus-spec.js
@@ -5,6 +5,7 @@
 
 describe('Http.Api.Skus', function () {
     var workflowApiService;
+    var _;
     before('start HTTP server', function () {
         this.timeout(5000);
         workflowApiService = {
@@ -16,7 +17,9 @@ describe('Http.Api.Skus', function () {
         ]);
         return helper.startServer([
             dihelper.simpleWrapper(workflowApiService, 'Http.Services.Api.Workflows')
-        ]);
+        ]).then(function() {
+            _ = helper.injector.get('_');
+        });
     });
 
     beforeEach('reset test DB', function () {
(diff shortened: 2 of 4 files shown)


zyoung51:
Test this please

spec/lib/api/1.1/skus-spec.js
```diff
@@ -16,7 +17,9 @@ describe('Http.Api.Skus', function () {
         ]);
         return helper.startServer([
             dihelper.simpleWrapper(workflowApiService, 'Http.Services.Api.Workflows')
-        ]);
+        ]).then(function() {
+            _ = helper.injector.get('_');
```
yyscamper: `_` has already in the global environment, you don't need to do this. see this line https://github.com/RackHD/on-core/blob/master/spec/helper.js#L15

spec/lib/services/sku-pack-service-spec.js
```diff
@@ -227,6 +229,16 @@ describe("SKU Pack Service", function() {
...
+
+        it('should get details on a pack', function() {
+            return skuService.getPackInfo('a')
+            .then(function(obj) {
+                expect(obj.description).to.be.a.string;
+                expect(obj.description).to.equal(data.description);
+                expect(obj.version).to.be.a.string;
+                expect(obj.version).to.equal(data.version);
+            });
+        });
```
yyscamper: Test coverage problem, you forgot the negative test cases, if no description or version in the sku pack's config.json, they will be defaults to `null` as you defined.

lib/services/sku-pack-service.js
```diff
@@ -371,6 +371,22 @@ function skuPackServiceFactory(
...
+        .then(function (conf) {
+            var skuConf = conf[skuid + '.json'];
+            if(skuConf) {
+                var json = JSON.parse(skuConf);
+                return {
+                    description: json.description || null,
+                    version: json.version || null
+                };
+            }
+            return null;
```
yyscamper: Since in line  382-383 you defaults the `description` and `version` to null, so why not return {description: null, version: null} here? I think this will match with your previous code.

spec/lib/api/1.1/skus-spec.js
zyoung51: As a CC you want me to trace through the code and out of this repository to find that you inject this as a global variable somewhere else?  You feel so strongly about it that you nitpicked this line of code?

yyscamper: Just a suggestion or reminder, your code will work as well, but this is not necessary, because we have put it into global variables, on-http/spec/helper.js will includes on-core/spec/helper.js, this is our design.

zyoung51: The design you want reinforced is to stash global variables in another repo with no documentation or comments that you are doing so, and then expect people to use them?  

Should I use that pattern more?  Should I be stashing globals in on-core for other things and not using the injector as an interface?  I use 'fs' sometimes.  Should I put it in a global in on-core?

lib/services/sku-pack-service.js
zyoung51: Because there is no sku pack.  You want me to reflect that no pack is installed by pretending one is installed without a description or version defined?

spec/lib/api/1.1/skus-spec.js
yyscamper: Lack the development doc is a gap. As whether we need to put it into on-core's global, I think that depends. If we see many unit-testing will use it I suggest to move it to on-core, this will sure bring much convenience. So you will not need to require `mocha`, `sinon`, `chai`, you use it without require because we do in the common helper.js. Of course you can reject this shortcut.

So this is just a suggestion for you, not an imperative command.

lib/services/sku-pack-service.js
yyscamper: If this is a kind of error case (but not causing critical problem because it is optional), so this is OK for me.

yyscamper:
:+1: after fix the test coverage

zyoung51:
test this please

yyscamper:
:+1:

jlongever:
test this please

zyoung51:
test this please

jlongever:
+1

PR #174, opened 2016-03-08
Add openSUSE/SLES bootstrap support

anhou:
remove old openSUSE profile and add new support for both SLES and openSUSE multiple version.
SLES11, SLES12 and and the latest openSUSE version leap/42.1 is verified. the usage is the same with CentOS/RHEL bootstrap. the document will be updated, the verification is via mounted ISO as http server, the validated iso image is below.
openSUSE-Leap-42.1-DVD-x86_64.iso
SLE-12-SP1-Server-DVD-x86_64-GM-DVD1.iso
SLES-11-SP4-DVD-x86_64-GM-DVD1.iso
and the bootstrap by  NAT/proxy access to online openSUSE repo is also verified.

Updated: full set parameters in template are supported, they are the same with CentOS/RHEL installation

related PR:
https://github.com/RackHD/on-tasks/pull/140
https://github.com/RackHD/on-taskgraph/pull/74


data/templates/suse-autoinst.xml
```diff
@@ -0,0 +1,107 @@
...
+
+  <scripts>
+    <chroot-scripts config:type="list"> 
+      <script>
+        <chrooted config:type="boolean">true</chrooted>
+        <interpreter>shell</interpreter>
+        <!-- scripts for workflow completion -->
+        <source> <![CDATA[
+          #!/bin/sh
+          curl http://<%=server%>:<%=port%>/api/common/templates/renasar-ansible.pub
```
zyoung51: This should be the completionUrl and not hardcoded to renasar-ansible.pub

data/templates/suse-autoinst.xml
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0"?>
```
zyoung51: Where is the networkDevice setup?

data/templates/suse-autoinst.xml
```diff
@@ -0,0 +1,107 @@
...
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+
+  <partitioning config:type="list">
```
zyoung51: Should this use the installDisk parameter?

zyoung51:
+0.  This is fine as a starting point.  It would be a +1 if it supported the full set of existing options that we already support on centos, redhat and esxi.

anhou:
@zyoung51 yes, it's a starting point, not a fully support, networkdevice and installDisk is not included yet. let me mention it in PR message, and I'll fix the hardcoded completionUri

data/templates/suse-autoinst.xml
```diff
@@ -0,0 +1,107 @@
...
+      <user_password><%=rootEncryptedPassword%></user_password>
+      <username>root</username>
+    </user>
+<% users.forEach(function(user) { %>
+    <user>
+      <encrypted config:type="boolean">true</encrypted>
+      <fullname><%=user.name%></fullname>
+      <user_password><%=user.encryptedPassword%></user_password>
+      <username><%=user.name%></username>
+    </user>
```
yyscamper: For CentOS/RHLE/ESXi, it supports config each user's `uid` (`gid` is a miss), so this need a follow up action.

yyscamper:
:+1: , and make follow up action to support the full list of bootstrap parameters, like `installDisk`, 'networkDevices`, each user's `uid`

anhou:
@yyscamper just be curious that why set user's 'uid', is it essential? if yes, why don't have a 'gid' (user group id) ? any history for this? IMHO, it don't need the ability to set uid and gid for users.

yyscamper:
@anhou , it is from product team's requirement.

data/templates/suse-autoinst.xml
```diff
@@ -0,0 +1,187 @@
...
+      <gid>0</gid>
+      <uid>0</uid>
+      <user_password><%=rootEncryptedPassword%></user_password>
+      <username>root</username>
+    </user>
+<% users.forEach(function(user) { %>
+    <user>
+      <encrypted config:type="boolean">true</encrypted>
+      <fullname><%=user.name%></fullname>
+      <uid><%=user.uid%></uid>
```
yyscamper: `uid` is better to be optional

anhou:
@zyoung51 @yyscamper  Other parameters such as network device, kvm, installDisk, dns are updated, originally plan to another PR due to thinking a little complex for these parameters.  but found it's not complex after digging and verifications, so include these in this PR.

data/templates/suse-autoinst.xml
```diff
@@ -0,0 +1,187 @@
...
+
+  <software>
+    <packages config:type="list">
+      <package>openssh</package>
+<% if( typeof kvm !== 'undefined' ) { %>
+      <package>qemu-kvm</package>
+      <package>virt-manager</package>
+      <package>libvirt</package>
+      <package>libvirt-python</package>
+      <package>bridge-utils</package>
```
yyscamper: Please confirm you have verified these package names are correct.

data/templates/suse-autoinst.xml
anhou: it keep the same with CentOS. it could be refined with centos  in another PR, not be planned in this PR

anhou:
Jekins fail is not related with this PR. Jenkins go to bad state again?
test this please

zyoung51:
@anhou Thanks for adding extra features to align with other os installs.  Jenkins and unittests issues should be resolved now.   Can you rebase to retrigger Travis?

yyscamper:
Still :+1:

PR #175, opened 2016-03-08
updated swagger spec for API 1.1.0

kacole2:
the previous swagger spec was broken because of a basic code generation. this needs to be maintained more carefully and not reliant upon generated outputs. @akutz fixed this part: You must have a  tag for a parameter of type . It was added wherever it was omitted, setting the schema to a variant type instead. It was omitted in a lot of requests as well as one response. Within the lookups, I (@kacole2) changed the lookup query reuirement from FALSE to TRUE. For some reason the code would fail to execute unless it was set to TRUE. There were some memory pointers happening and it will have to be looked at once again. To get a better idea of how this is all generated using go-swagger, checkout the gorackhd repo and follow the directions listed there. https://github.com/emccode/gorackhd

Signed-off-by: Kendrick Coleman <kendrickcoleman@gmail.com>

### DIFF
--- a/static/monorail.yml
+++ b/static/monorail.yml
@@ -677,6 +677,7 @@ paths:
         default:
           description: Unexpected error
           schema:
+            type: object
 
   /workflows/tasks/library:
     get:
@@ -727,7 +728,8 @@ paths:
         - name: body
           in: body
           required: false
-          type: object
+          schema:
+            type: object
       tags:
         - workflow
         - task
@@ -801,7 +803,8 @@ paths:
         - name: body
           in: body
           required: false
-          type: object
+          schema:
+            type: object
       tags:
         - workflow
         - put
@@ -940,7 +943,8 @@ paths:
         - name: body
           in: body
           required: false
-          type: object
+          schema:
+            type: object
       tags:
         - nodes
         - workflow
@@ -977,7 +981,8 @@ paths:
         - name: body
           in: body
           required: false
-          type: object
+          schema:
+            type: object
       tags:
         - nodes
         - dhcp
@@ -1432,7 +1437,7 @@ paths:
         - name: q
           in: query
           description: query object to pass through to waterline.
-          required: false
+          required: true
           type: string
       tags:
         - lookups
@@ -1539,7 +1544,8 @@ paths:
           description: |
             object patches to apply.
           required: true
-          type: object
+          schema:
+            type: object
       tags:
         - lookups
         - get
@@ -1714,7 +1720,8 @@ paths:
         - name: body
           in: body
           required: true
-          type: object
+          schema:
+            type: object
       tags:
         - config
         - patch


jlongever:
ok to test

static/monorail.yml
```diff
@@ -1432,7 +1437,7 @@ paths:
         - name: q
           in: query
           description: query object to pass through to waterline.
-          required: false
+          required: true
```
heckj: @benbp can you take a quick look at the code and verify? The relevant line is https://github.com/RackHD/on-http/blob/master/lib/api/1.1/northbound/lookups.js#L31, so the main branch of this definitely wants to see a `q` there, but I think the intent was that `waterline.lookups.findByTerm(null)` would return all values. If that's correct (and works) then `q` isn't required.

heckj:
Not sure about the required `q` part - but the rest looks great! :+1: with other validation on the `q` required thing in `GET /lookups`

static/monorail.yml
kacole2: i originally tried to make this work with Q set to FALSE but just couldn't. as soon as i switched it to TRUE, then all of sudden the queries would work. Let me try testing again by throwing in `nil`. https://github.com/emccode/docker-machine-rackhd/blob/master/rackhd.go#L155

kacole2: i just tested this and here are the results.

When lookups `q` query is set to TRUE:
1) Pass in q as a string (Node, IP, Mac) and the returned result is the expected array
2) Pass in `nil`, then and error comes back with `cannot use nil as type string in field value`
3) Pass in `""` (completely blank), the returned result is *everything* in an array as expected

When lookups `q` query is set to FALSE:
1) Pass in q as a string (Node, IP, Mac) and the returned result is the error: `cannot use "56dde3441722c192796e3a38" (type string) as type *string in field value`
2) Pass in `nil`,  the returned result is *everything* in an array as expected

zyoung51: The on-http 1.1 API implementation doesn't use swagger in any capacity.  The source of that validation error message is not RackHD.

When the python library is updated to use this change (occurs post commit), this should break our functional tests here: https://github.com/RackHD/RackHD/blob/master/test/tests/api/v1_1/lookups_tests.py#L61.

brianparry:
test this please

kacole2:
@brianparry i have my manual testing notes in the thread. unless that wasn't directed at me.

zyoung51:
@kacole2 Brian's message is a trigger phrase for Jenkins and is only meant for that system.

heckj:
@kacole2 that's the phrase that commiters can invoke to have jenkins do a test build/run against the code - not meant for you specifically

kacole2:
@zyoung51 @heckj sorry. didn't know what wizardry you all had here!

brianparry:
Sorry for the confusion @kacole2.  That comment was for Jenkins not you :)

jlongever:
Jenkins, test this please

heckj:
Mr Jenkins bot, test this please

heckj:
@RackHD/corecommitters looks like earlier testing hung or something, trying to retrigger

jlongever:
bah,  seen this with other checks.  The status gets stuck on running.   I've had to close/reopen the PR before to "dislodge" it.

heckj:
will do that, thanks

static/monorail.yml
jlongever: @zyoung51 comment ^^ still holds true when required is set to true.  The python functional tests run tests without the 'q' parameter, both library and functional tests will need to be updated if we force 'q' parameter.

brianparry: @jlongever The tests and library are correct now.  We shouldn't be setting the required flag for an optional parameter even if it's not actually enforced by the API.

jlongever: fwiw, we're using swagger-codegen and haven't had issues with the optional 'q' parameter.

benbp:
test this please

anhou:
test this please

benbp:
test this please

anhou:
@kacole2 @brianparry @jlongever  is this PR still valid? it's pending for a long time. not sure if it's valid after two months' many codes update, and conflicts need to be resovled.

kacole2:
@anhou close it. will re-open if this isn't fixed in v2.0 API. The gorackhd library is going to keep a manual swagger-spec with this changed to keep it stable and working as expected

PR #178, opened 2016-03-10
[swagger] Add AuthN/AuthZ processing to routes

zyoung51:
- Fittings are created to enable authn (using x-authentication-type) and
  authz (using x-privileges) on a per-method basis defined within the
  swagger file
- ACL service is created as an interface to node_acl.  The ACL service
  contains the mappings of users to roles, and roles to privileges.
- A /users route is added to 2.0 to enable a starting point for a user
  management API.  Initially POST and PUT are supported to create and update
  users.  A local user exception is added to permit a localhost connection to
  add the first user into the database when authentication is enabled and no
  users exist yet.
- The authentication service is refactored to remove the hardcoded default
  username/password and to remove decryptable passwords.  Bcrypt is used
  instead for 1-way password hashing.

### DIFF
--- a/lib/api/2.0/users.js
+++ b/lib/api/2.0/users.js
@@ -1,0 +1,93 @@
+// Copyright 2016, EMC Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var _ = injector.get('_'); // jshint ignore:line
+var configuration = injector.get('Services.Configuration');
+var waterline = injector.get('Services.Waterline');
+var Errors = injector.get('Errors');
+var aclService = injector.get('Acl.Services');
+var localUserException = configuration.get('enableLocalHostException', true);
+
+var addUser = controller({success: 201}, function(req, res) {
+    var userObj = _.pick(req.swagger.params.body.value, ['username', 'password', 'role']);
+    userObj.role = userObj.role || 'ReadOnly';  //TODO: move to Constants when Role code is added
+
+    return waterline.localusers.find({}).then(function(users) {
+        if(!users.length && localUserException && res.locals.ipAddress === '127.0.0.1') {
+            // Only when there are no users, and the remote is a local connection, and we 
+            // permit it, then let them add the first user.
+            return waterline.localusers.create(userObj)
+            .then(function(user) {
+                return [user, aclService.aclMethod('addUserRoles',
+                    userObj.username,
+                    userObj.role )];
+            }).spread(function(user) {
+                return _.pick(user, ['username', 'role']);
+            });
+        }
+
+        if( !_.find(users, function(user) {
+            return user.username === userObj.username;
+        })) {
+            if( req.isAuthenticated && req.isAuthenticated() ) {
+                return waterline.localusers.create(userObj)
+                .then(function(user) {
+                    return [user, aclService.aclMethod('addUserRoles',
+                        userObj.username,
+                        userObj.role )];
+                }).spread(function(user) {
+                    return _.pick(user, ['username', 'role']);
+                });
+            }
+        }
+
+        throw new Errors.UnauthorizedError('Unauthorized');
+    });
+});
+
+var modifyUser = controller(function(req) {
+    var userObj = _.pick(req.swagger.params.body.value, ['username', 'password', 'role']);
+    if(!req.isAuthenticated) {
+        throw new Errors.UnauthorizedError('Unauthorized');
+    }
+
+    if(!req.isAuthenticated()) {
+        throw new Errors.UnauthorizedError('Unauthorized');
+    }
+
+    if( (req.user === userObj.username && req.hasRole('ConfigureSelf')) ||
+        req.hasRole('Administrator') || 
+        req.hasRole('ConfigureUsers') )
+    {
+        return waterline.localusers.findOne({username: userObj.username})
+        .then(function(user) {
+            if(user.role !== userObj.role) {
+                return aclService.aclMethod('removeUserRoles',
+                    user.username,
+                    user.role)
+                .then(function() {
+                    return user;
+                });
+            }
+            return user;
+        }).then(function() {
+            return waterline.localusers.update({username: userObj.username}, userObj);
+        }).then(function(user) {
+            return [user, aclService.aclMethod('addUserRoles',
+                user.username,
+                user.role )];
+        }).spread(function(user) {
+            return _.pick(user, ['username', 'role']);
+        });
+    }
+
+    throw new Errors.ForbiddenError('Forbidden');
+});
+
+module.exports = {
+    addUser: addUser,
+    modifyUser: modifyUser
+};
--- a/lib/fittings/swagger_authz.js
+++ b/lib/fittings/swagger_authz.js
@@ -1,0 +1,67 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var injector = require('../../index.js').injector;
+var _ = injector.get('_');    // jshint ignore:line
+var Promise = injector.get('Promise');    // jshint ignore:line
+var debug = require('debug')('swagger:authz');
+var Error = injector.get('Errors'); // jshint ignore:line
+
+function loadSwagger(authz, swaggerDef) {
+    var promises = [];
+    var methods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch'];
+    _.forOwn(swaggerDef.paths, function(value, resourceName) {
+        _.forEach(methods, function(method) {
+            if(_.has(value, method) && _.has(value[method], 'x-privileges')) {
+                _.forEach(value[method]['x-privileges'], function(role) {
+                    var resource = swaggerDef.basePath + resourceName;
+                    debug('adding ACL: %s:%s:%s', role, resource, method);
+                    promises.push( authz.aclMethod('allow', role, resource, method) );
+                });
+            }
+        });
+    });
+    return Promise.all(promises);
+}
+
+var hasRole = function(role) {
+    return _.includes(this.roles, role );
+};
+
+module.exports = function create(fittingDef, bagpipes) {
+    injector = require('../../index.js').injector;
+    var authz = injector.get('Acl.Services');
+    var swaggerNodeRunner = bagpipes.config.swaggerNodeRunner;
+    var ready = loadSwagger(authz, swaggerNodeRunner.swagger);
+    var basePath = swaggerNodeRunner.swagger.basePath;
+    return function swagger_authn_authz(context, next) {    // jshint ignore:line
+        if(context.request.user) {
+            ready.then(function() {
+                return authz.aclMethod('isAllowed',
+                    context.request.user,
+                    basePath + context.request.swagger.operation.pathObject.path,
+                    context.request.method.toLowerCase());
+            }).then(function(allowed) {
+                if(!allowed) {
+                    return context.response.status(403).json({message: "Forbidden"});
+                } else {
+                    authz.aclMethod('userRoles', context.request.user)
+                    .then(function(roles) {
+                        return authz.aclMethod('_allRoles', roles);
+                    }).then(function(allRoles) {
+                        context.request.roles = allRoles;
+                        context.request.hasRole = hasRole;
+                        next();
+                    });
+                }
+            })
+            .catch(function(err) {
+                next(err);
+            });
+        } else {
+            debug('skipping authz, unknown authn');
+            next();
+        }
+    };
+};
--- a/lib/services/acl-service.js
+++ b/lib/services/acl-service.js
@@ -1,0 +1,75 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di'),
+    nodeAcl = require( 'acl' );
+
+module.exports = aclServiceFactory;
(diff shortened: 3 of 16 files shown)


lib/fittings/swagger_authz.js
```diff
@@ -0,0 +1,67 @@
...
+
+function loadSwagger(authz, swaggerDef) {
+    var promises = [];
+    var methods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch'];
+    _.forOwn(swaggerDef.paths, function(value, resourceName) {
+        _.forEach(methods, function(method) {
+            if(_.has(value, method) && _.has(value[method], 'x-privileges')) {
+                _.forEach(value[method]['x-privileges'], function(role) {
+                    var resource = swaggerDef.basePath + resourceName;
+                    debug('adding ACL: %s:%s:%s', role, resource, method);
```
heckj: dummy question - where is it outputting this data to? Is this basically console only logging? (and curious why you didn't use logger here...)

lib/services/auth-service.js
```diff
@@ -37,23 +39,12 @@ function authServiceFactory(
...
     AuthService.prototype.init = function(){
         var self = this;
-        this.strategies = [];
 
         self.algorithm = 'HS256';
         self.secretOrKey = '';
         self.ignoreTokenExpiration = false;
         self.encoder = 'base64';
 
-        self.hashConfig = {
```
heckj: is removing this reducing the strength of the hashing, or was this the default and simply redundant?

lib/fittings/swagger_authz.js
zyoung51: The debug module is used throughout the swagger code and I wanted to keep that consistent.  This and the authn are in the 'swagger:' debug namespace so someone debugging swagger and using DEBUG=swagger:* will also see these messages.  The output is to the console.

lib/services/auth-service.js
```diff
@@ -86,38 +95,17 @@ function authServiceFactory(
...
-            'zlxkgxjvcFwm0M8sWaGojh25qNYO8tuNWUMN4xKPH93PidwkCAvaX2JItL' +
-            'A3p7BSCWIzkw4GwWuezoMvKf3UXg==');
-        var secret = configuration.get('authTokenSecret',
-            'RackHDRocks!');
-
-        assert.string(username, 'Invalid username in configure file');
-        assert.string(passwordHash, 'Invalid password hash in configure file');
-        assert.string(salt, 'Invalid crypto salt in configure file');
+    AuthService.prototype.loadFromConf = function () {
+        var secret = configuration.get('authTokenSecret', 'RackHDRocks!');
```
heckj: Could you add some detail into http://rackhd.readthedocs.org/en/latest/packaging_and_deployment.html#configuration-parameters with this value, and it's default

lib/services/http-service.js
```diff
@@ -138,18 +139,20 @@ function httpServiceFactory(
...
         var redfishConfig = {
             appRoot: path.resolve(__dirname, '../../'),
             configDir: path.resolve(__dirname, '../../swagger_config'),
             swagger: path.resolve(__dirname, '../../static/redfish.yaml'),
+            authEnabled: this.endpoint.authEnabled
         };
 
         var swaggerCreateAsync = Promise.promisify(swagger.create);
 
         var apiV2 = swaggerCreateAsync(swaggerConfig).then(function(swaggerExpress) {
```
heckj: Do we do any chaining where we need to return the results of `swaggerExpress.register(app)`? any inadvertent implications to this change, or benefits?

lib/fittings/swagger_authz.js
heckj: Ah gotcha - since we're adding something new in terms of how to get to this debugging detail, can you toss a relevant note in the repo's README or something? Somewhere to know how to get to these debugging statements and make them visible since we're using an alternate convention for the swagger pieces?

lib/services/http-service.js
zyoung51: Nice catch.  No chaining, but this was an inadvertent change, will fix it.

spec/lib/services/auth-service-spec.js
```diff
@@ -511,10 +426,10 @@ describe('Auth.Service', function () {
 
             return helper.request('https://localhost:9443')
                 .get('/api/1.1/config?auth_token=' + token)
-                .expect(ERROR_STATUS)
+                .expect(401)
                 .expect(function (res) {
                     expect(res.body.message).to.be.a('string');
-                    expect(res.body.message).to.equal('Internal server error');
+                    expect(res.body.message).to.equal('Unauthorized');
```
heckj: :+1:

static/redfish.yaml
```diff
@@ -1022,6 +1094,8 @@ paths:
   /TaskService/Oem/Tasks/{identifier}:
     x-swagger-router-controller: task-service
     get:
+      x-privileges: [ 'Login' ]
```
heckj: should we be building up a list of privileges to know what we've used (and should re-use) in the developer's guide or something? tying to figure out how to track these pieces to know how/when to re-use them and assert our common conventions for when we need to extend this later.

heckj:
:+1: with some documentation requests

static/redfish.yaml
zyoung51: For now there is a hardcoded set or role to privileges here:https://github.com/RackHD/on-http/pull/178/files#diff-a8be6600959fd2bbeb19f645072425c5R57.  As for where they came from, the Redfish ones are defined in the spec and the 2.0 ones are a stab at something to get it going.

The 'roles' model is added to on-core with the intention that a user will use an API (after we create it) to define their own set of Role/Privilege mappings.  There needs to be a continuing effort to refine the 2.0 privileges to something more granular than Read/Write.

spec/lib/api/2.0/users-spec.js
```diff
@@ -0,0 +1,150 @@
...
+        waterline.localusers.find.reset();
+        waterline.localusers.update.reset();
+        waterline.localusers.findOne.reset();
+    });
+
+    it('should 201 a user post attempt with localexception', function() {
+        waterline.localusers.find.resolves([]);
+        waterline.localusers.create.resolves(userObj);
+        return helper.request().post('/api/2.0/users')
+            .send(userObj)
```
anhou: the test case process is the same with 'should 401 a user post attempt without auth tokens‘  , IMHO, for use case, any access to /api/2.0/users should be authenticated firstly,  so it would not cause the local exception

spec/lib/api/2.0/users-spec.js
```diff
@@ -0,0 +1,150 @@
...
+            .expect('Content-Type', /^application\/json/)
+            .expect(401)
+            .expect(function() {
+                expect(waterline.localusers.find).to.have.been.calledOnce;
+            });
+    });
+
+    it('should 200 a user post attempt with auth tokens', function() {
```
anhou: 'should 201' here

spec/lib/api/2.0/users-spec.js
zyoung51: The local exception is put into place for the case where no one can be authenticated because there are no users in existence yet.  For this test case, find resolves to an empty set and the connection is already over localhost, which allows the 201.

All our connections are over localhost, so to avoid the local exception in the 401 test case below, find resolves to a set containing users.

If it helps, Mongo's usage of local exception is here: https://docs.mongodb.org/manual/core/security-users/#localhost-exception

static/monorail-2.0.yaml
```diff
@@ -1957,7 +2091,60 @@ paths:
...
+      parameters:
+        - name: body
+          in: body
+          description: User information
+          required: true
+          schema:
+            $ref: '#/definitions/user_obj'
+      tags: [ "/api/2.0" ]
+      responses:
+        201:
```
anhou: 200 here?

static/monorail-2.0.yaml
```diff
@@ -1957,7 +2091,60 @@ paths:
...
+          description: |
+            the user has been created
```
anhou: 'created' is 'updated’

static/monorail-2.0.yaml
```diff
@@ -1957,7 +2091,60 @@ paths:
...
+          schema:
+            type: object
+        default:
```
anhou: responses also include '403' and its description?

spec/lib/api/2.0/users-spec.js
anhou: Thanks for explanation, have understand this case is for the first user, :+1: maybe the title is ' **** mongodb Localhost Exception' or ' **** for first user' is more readable for other reviewers

lib/services/auth-service.js
zyoung51: The on-core PR was updated back to using pbkdf2 over bcrypt.  The settings below are copied over into the on-core change.

anhou:
:+1:  just fix some small  pieces in comments, also fix houndci 'Missing semicolon' and Jenkins pass

yyscamper:
Would like @cgx027 and @WangWinson to review this design as they implemented the first version API auth.

lib/api/2.0/users.js
```diff
@@ -0,0 +1,93 @@
...
+var aclService = injector.get('Acl.Services');
+var localUserException = configuration.get('enableLocalHostException', true);
+
+var addUser = controller({success: 201}, function(req, res) {
+    var userObj = _.pick(req.swagger.params.body.value, ['username', 'password', 'role']);
+    userObj.role = userObj.role || 'ReadOnly';  //TODO: move to Constants when Role code is added
+
+    return waterline.localusers.find({}).then(function(users) {
+        if(!users.length && localUserException && res.locals.ipAddress === '127.0.0.1') {
+            // Only when there are no users, and the remote is a local connection, and we
```
brianparry: @zyoung51 I'm thinking we may want to force this user into the admin role or return an error if the posted user isn't admin.

lib/api/2.0/users.js
```diff
@@ -0,0 +1,93 @@
...
+                user.username,
+                user.role )];
+        }).spread(function(user) {
+            return _.pick(user, ['username', 'role']);
+        });
+    }
+
+    throw new Errors.ForbiddenError('Forbidden');
+});
+
```
brianparry: @zyoung51 Might want a getUsers route for Administrator and ConfigureUsers roles

lib/services/auth-service.js
```diff
@@ -67,6 +58,24 @@ function authServiceFactory(
             secretOrKey : self.secretOrKey,
             ignoreExpiration : self.ignoreTokenExpiration
         };
+
+        // used to serialize the user for the session
```
brianparry: @zyoung51 I'm a little confused by this code.  My understanding is that serializeUser produces a unique value that can be used by the deserializeUser function to reference a user object.  This value then gets stored in the session.  In this case, it looks like you're serializing the whole user object to the session, but then expecting a user id in the deserializeUser function.  The id passed to deserializeUser will be the whole user object so I would think you could just return done(null, user).  Maybe this is working differently because we're not using sessions?

The example from the passport documentation looks like this:
```javascript
passport.serializeUser(function(user, done) {
  done(null, user.id);
});

passport.deserializeUser(function(id, done) {
  User.findById(id, function(err, user) {
    done(err, user);
  });
});
```

lib/services/acl-service.js
```diff
@@ -0,0 +1,75 @@
...
+                // Redfish 1.0 fixed privileges
+                acl.addRoleParents('Administrator', ['Login','ConfigureManager','ConfigureUsers']),
+                acl.addRoleParents('Administrator', ['ConfigureComponents','ConfigureSelf']),
+                acl.addRoleParents('Operator', ['Login','ConfigureComponents','ConfigureSelf']),
+                acl.addRoleParents('ReadOnly', ['Login', 'ConfigureSelf'])
+            ]);
+        });
+    };
+
+    AclService.prototype.aclMethod = function(name) {
```
WangWinson: Is there specific reason to encapsulate this method instead of call acl.method directly ?

zyoung51: The main reason why I did it this way is to create a point to log other code's usage of the acl service through a single entry point.  I also found that I was creating a 1:1 mapping between the acl service and the acl library, and this was a way to avoid doing that by creating a 1:N bridge function.

lib/services/auth-service.js
zyoung51: @brianparry Rather than using passport.session() middleware, I'm calling the logIn function directly and passing it the username:https://github.com/RackHD/on-http/pull/178/files#diff-a5da55608b028f454f167e0a83bbfabeR239.  logIn will call serializeUser with the parameter it is passed (the passport middleware normally calls logIn with the session object) .  In the case of my code, serializeUser is really only used to return success, as it needs to return successfully for logIn to succeed.

As for deserializeUser, that isn't used at all and can be deleted.  I'll remove deserializeUser and update the comments around serializeUser.

brianparry: @zyoung51 Makes sense.  Thanks for the explanation.

zyoung51:
on-core dependent patch was merged, can you test this please

zyoung51:
Depends on https://github.com/RackHD/RackHD/pull/136 to resolve failed test.

zyoung51:
test this please

jlongever:
+1

PR #180, opened 2016-03-14
[skupack] Pull workflow name from sku environment

zyoung51:


heckj:
Heh - the inevitable missed `;`,  :+1: for the content

yyscamper:
Hi @zyoung51 , I want to confirm the user case since no description in this PR. Whether it means RackHD will define a list of standard Graph names, user can map them to a platform specified graph in SKU pack, such as Graph.CreatePollers ==> Graph.CreateIpmiPollers or Graph.CreatePollers ==>Graph.CreateRedfishPollers, RackHD can leverage this to get some platform agnostic design, is this correct?

zyoung51:
@yyscamper: The sku specific graph name is assembled and stored in the sku environment here: https://github.com/RackHD/on-http/blob/master/lib/services/sku-pack-service.js#L158.  This code pulls the name set therein back out to be used as the sku specific graph name.  If no sku or sku graph mapping exists, then the default name is used.

For example, if I had a workflow in a skupack of Graph.CreatePollers then the skupack installation would create a graph with a name of Graph.CreatePollers::{skuid} and put an entry in the sku environment of { Graph: { CreatePollers: "GraphCreatePollers::{skuid}" } }  When the workflow api service is called to start a graph with a name of "Graph.CreatePollers", and a sku is present, it will do a environment lookup of "Graph.CreatePollers" and will get back the "Graph.CreatePollers::{skuid}" value.

Of course this is in workflow api service and the pollers do not call the workflow api service to start a graph but instead implement their own manual thing.  The poller code should be refactored to resolve those issues and I believe @jlongever had ideas on how to do that.

yyscamper:
@zyoung51 : gotcha. 
However, I see this only works for SKU workflows not works for SKU tasks. For example, there is a GraphX in SKU pack and it includes a taskX, but taskX is also defined in the SKU pack. So your code change will only find the correct GraphX, but the workflow engine will not be able to find the correct taskX.
So my suggestion is to move both SKU graph and task finding into the on-core (like task-graph-runner, hide the SKU finding in the graph runner and task runner ), so every repo can benefit, because some jobs still need the same functionality, like this one https://github.com/RackHD/on-tasks/blob/master/lib%2Fjobs%2Frun-graph.js#L97 and this one https://github.com/RackHD/on-tasks/blob/master/lib%2Fjobs%2Frun-sku-graph.js#L87 

How do you think about this?

zyoung51:
@yyscamper: The tasks called out in the graph are renamed to the sku specific tasks if they are included in the skupack: https://github.com/RackHD/on-http/blob/master/lib/services/sku-pack-service.js#L162.

yyscamper:
@zyoung51 : thanks, I missed this design. So trigger SKU specified graph (even with SKU specified task) via RackHD API will work well.
Considering some jobs still need this functionality, so if we can provide some helper function in on-core to find the SKU specified graph will be better. But it can be moved to another PR.

So for this PR, it is :+1: from me.

jlongever:
+1

PR #181, opened 2016-03-14
build refactor

ytjohn:
Ready for merge, pending testing.

This does the following:

1. Create extra directory containing make* scripts
2. Modify .travis.yml to utilize make* scripts
3. Modify build-package.bash to anticipate variables passed in.
4. A make-cicd.sh script that will emulate HWIMO-BUILD
5. Add Vagrantfile that will build package.

Travis Build using this method: https://travis-ci.org/ytjohn/on-http/builds/115897749

### DIFF
--- a/build-package.bash
+++ b/build-package.bash
@@ -3,20 +3,43 @@ DEBDIR="./debianstatic/${1}"
 BRANCH=${2}
 set -x
 
-if [ ! -d "${DEBDIR}" ]; then 
+if [ ! -d "${DEBDIR}" ]; then
     echo "no such debian directory ${DEBDIR}"
-    exit 1 
+    exit 1
 fi
 
 rm -rf ${1}_*
 rm -rf packagebuild
-mkdir -p packagebuild/debian 
+mkdir -p packagebuild/debian
 
 git log -n 1 --pretty=format:%h.%ai.%s > commitstring.txt
-export DEBFULLNAME=`git log -n 1 --pretty=format:%an`
-export DEBEMAIL=`git log -n 1 --pretty=format:%ae`
-export DEBBRANCH=`echo "${BRANCH}" | sed 's/[\/\_]/-/g'`  
-export DEBPKGVER=`git log -n 1 --pretty=oneline --abbrev-commit`
+if [ -z "$DEBFULLNAME" ]; then
+	export DEBFULLNAME=`git log -n 1 --pretty=format:%an`
+fi
+
+if [ -z "$DEBEMAIL" ]; then
+	export DEBEMAIL=`git log -n 1 --pretty=format:%ae`
+fi
+
+if [ -z "$DEBBRANCH" ]; then
+	export DEBBRANCH=`echo "${BRANCH}" | sed 's/[\/\_]/-/g'`
+fi
+
+if [ -z "$DEBPKGVER" ]; then
+  export DEBPKGVER=`git log -n 1 --pretty=oneline --abbrev-commit`
+fi
+
+if [ -z "$DCHOPTS" ]; then
+	export DCHOPTS="-l ${DEBBRANCH} -u low ${DEBPKGVER}"
+fi
+
+echo "DEBDIR:       $DEBDIR"
+echo "DEBFULLNAME:  $DEBFULLNAME"
+echo "DEBEMAIL:     $DEBEMAIL"
+echo "DEBBRANCH:    $DEBBRANCH"
+echo "DEBPKGVER:    $DEBPKGVER"
+echo "DCHOPTS:      $DCHOPTS"
+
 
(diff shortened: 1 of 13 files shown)


jlongever:
Should we just create a Makefile with different make rules?  Would probably be easier than managing a bunch of make-* scripts. imho

jlongever:
Example, it would be a single Makefile that does:
`make clean`
`make deb`
`make docs`

ytjohn:
I started doing a Makefile originally for exactly the same reasons. I ran into a lot of issues when it comes to the actual debuild which runs it's own make test and make clean.

Because of how build-package rsync's into a packagebuild directory and puts everything there, `make -j1 clean` was actually interfering. Unless we plan on doing even more directory restructuring and change how we build, the current directory structure is not compatible with Makefile.

ytjohn:
UPDATE: I had to fix a bit with the DCHOPTS in build-package.bash, see [travis-log](https://travis-ci.org/ytjohn/on-http/jobs/115924752) to review output.

Otherwise, I now feel this should be good for merging. If you spot any issues or would like me name something different, let me know and I can refactor further.

yyscamper:
Hi @ytjohn , thanks for this enhancement design, it looks good to me, but if put these bash scripts in the `npm script` section (https://github.com/RackHD/on-http/blob/master/package.json#L6) will be better, and this should also resolve @jlongever 's comment. How do you think?

And if this change is merged, we still need to add these change to other repos to keep them in a consistent view.

Meanwhile, I think @heckj may have further comment on this kind of change.

ytjohn:
@yyscamper  Thanks for the feedback. I admit I'm not familiar with npm packaging, so I didn't dive too far into the package.json. But I don't see any duplication of effort.

I also have a goal of applying this same methodology to all the other on- packages. on-http and on-taskgraph are the most complicated, so I am starting with those first. 

Looking at the scripts section, it looks like there are commands in here that both .travis.yml and HWIMO-* called.  For instance, the "make-docs.sh" runs `npm apidoc`, and make-npmdeps.sh executs `npm install`.

One thing I could see putting in here is the test coveralls.  Today npm test does one thing with mocha, and make-testcoveralls.sh runs that under istanbul. 

`./node_modules/.bin/istanbul cover -x "**/spec/**" ./node_modules/.bin/_mocha --report lcovonly -- $(find spec -name '*-spec.js') -R spec --require spec/helper.js`


```
  "scripts": {
    "doc": "./node_modules/.bin/jsdoc lib/api/1.1/northbound -r -d docs",
    "test": "./node_modules/.bin/mocha $(find spec -name '*-spec.js') -R spec --require spec/helper.js",
    "install": "./install-web-ui.sh && ./install-swagger-ui.sh",
    "apidoc": "./node_modules/.bin/apidoc -i lib/api/1.1/northbound -f '.*\\.js$' -o build/apidoc"
  },
```

build-package.bash
```diff
@@ -13,10 +13,33 @@ rm -rf packagebuild
...
+if [ -z "$DCHOPTS" ]; then
+	export DCHOPTS="-l ${DEBBRANCH} -u low ${DEBPKGVER}"
+fi
+
+echo "DEBDIR:       $DEBDIR"
+echo "DEBFULLNAME:  $DEBFULLNAME"
+echo "DEBEMAIL:     $DEBEMAIL"
+echo "DEBBRANCH:    $DEBBRANCH"
+echo "DEBPKGVER:    $DEBPKGVER"
+echo "DCHOPTS:		  $DCHOPTS"
```
yyscamper: a little obsessive-compulsive personality, this line is not well-aligned with above :)

ytjohn: I can fix that. Actually, I could remove these, but I found it helpful when troubleshooting.

yyscamper:
Hi @ytjohn : Add a command into the scripts section of package.json is very straightforward, no complicated logic like Makefile/Grunt/Gulp. Adding the coverall into the scripts is OK for me.

So it is :+1: for me, if moving some scripts into package.json and further to simplify the travis.yml and HWIMO-xxx is a plus.

build-package.bash
ytjohn: Looks like a real tab snuck in there somehow.

ytjohn:
That test failure is from a prior merge. Prior build fail for comparison. https://travis-ci.org/RackHD/on-http/jobs/116437109

jlongever:
been hitting edge cases with `test_node_workflows_post` test.   jenkins, test this please.

ytjohn:
Is there anything else needed to get this merged in?

yyscamper:
Still lack another +1, @RackHD/corecommitters @heckj , could you please help review this?

anhou:
:+1:

PR #187, opened 2016-03-23
use 'repo' as base_url in profile for CoreOS

heckj:
- partial resolution to https://github.com/RackHD/RackHD/issues/146 and
   include 'current' in profile explicitly so we can use <%repo%> in the
   rendered install template correctly.
 - requires a change to the configuration in RackHD proxies to fully operate (https://github.com/RackHD/RackHD/pull/147)
 - additionally requires https://github.com/RackHD/on-tasks/pull/147 for a default 'version' for CoreOS install task

### DIFF
--- a/data/profiles/install-coreos.ipxe
+++ b/data/profiles/install-coreos.ipxe
@@ -5,6 +5,6 @@ set base-url <%=repo%>
 # JUST BOOT AND RUN COREOS
 # kernel ${base-url}/coreos_production_pxe.vmlinuz coreos.autologin cloud-config-url=http://<%=server%>:<%=port%>/api/common/templates/pxe-cloud-config.yml
 #
-kernel ${base-url}/coreos_production_pxe.vmlinuz console=tty0 console=<%=comport%>,115200n8 coreos.autologin cloud-config-url=http://<%=server%>:<%=port%>/api/common/templates/install-coreos.sh
-initrd ${base-url}/coreos_production_pxe_image.cpio.gz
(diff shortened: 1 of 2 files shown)


heckj:
@RackHD/corecommitters reviews please

yyscamper:
test this please

yyscamper:
If hard-code the `current` in the ipxe script, that means user can only use the latest version of CoreOS. I suggest adding the `version` in the ipxe script, so it is flexible to install any version of CoreOS. By default, the version can be set as `current` in Graph.InstallCoreOS:
```
set base-url <%=repo%>/<%=version%>
kernel ${base-url}/coreos_production_pxe.vmlinuz ...
initrd ${base-url}/coreos_production_pxe_image.cpio.gz
boot || prompt --key 0x197e --timeout 2000 Press F12 to investigate || exit shell
```

benbp:
I think I agree with @yyscamper's comment here.

heckj:
cool - will update

heckj:
@benbp @yyscamper updated. Note - it will break if we update `repo`, since that's passed through to the install script (see the line at https://github.com/RackHD/on-http/blob/master/data/templates/install-coreos.sh#L3). If we put version in there, it'll break the `-b` option on the installation that points to the base level installation scripts that we don't have control over in CoreOS.

This does accommodate letting us pass in a specific version, with a default of `current` defined in the PR https://github.com/RackHD/on-tasks/pull/147.

benbp:
@heckj so you're saying we can't have the repo/base-url equal `<api>/coreos/<version>` because `-b` doesn't support that full URI?

heckj:
@benbp yep, exactly

benbp:
:+1:

yyscamper:
:+1:

PR #188, opened 2016-03-23
Add /tasks route to 2.0 API

geoff-reid:
Added Southbound /tasks rout to the 2.0 API.

Right now all swagger routes, including the new tasks route, appear on both the southbound and northbound interfaces.  I am working on changes that make the routes available only on the correct interfaces.  This will be a different pull request.

### DIFF
--- a/static/monorail-2.0-sb.yaml
+++ b/static/monorail-2.0-sb.yaml
@@ -1,0 +1,152 @@
+swagger: "2.0"
+info:
+  version: "0.0.1"
+  title: RackHD 2.0 Southbound
+# during dev, should point to your local machine
+#host: localhost:10010
+# basePath prefixes all resource paths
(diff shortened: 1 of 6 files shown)


jlongever:
test this please

brianparry:
:+1:

static/monorail-2.0-sb.yaml
```diff
@@ -0,0 +1,152 @@
+swagger: "2.0"
+info:
+  version: "0.0.1"
+  title: RackHD 2.0 Southbound
```
jlongever: Is the plan to stick with the 1.1 paradigm of keeping these routes split over applying roles and access rules?

abildw:
+1

jlongever:
I suppose the answer is no then.

PR #193, opened 2016-03-29
Add northbound sku support

zyoung51:


spec/lib/api/redfish-1.0/system-spec.js
```diff
@@ -233,6 +233,38 @@ describe('Redfish Systems Root', function () {
             });
     });
 
+    it('should return a valid system with sku', function() {
+        waterline.nodes.needByIdentifier.withArgs('1234abcd1234abcd1234abcd')
+        .resolves(Promise.resolve({
```
yyscamper: The nested Promise is not necessary.

yyscamper:
@zyoung51 , could you please add some description in the PR so we can understand what is this PR doing and why does have this design.

spec/lib/api/redfish-1.0/system-spec.js
zyoung51: It needs to return a promise for the promise chain to function.

PR #194, opened 2016-03-29
Add v2.0 workflows API

brianparry:
* GET, POST, DELETE, cancel a workflow.
* GET, PUT, DELETE a graph definition.
* GET, PUT, DELETE a task definition.
* active query param to workflows route

Separate the workflows routes into workflows /workflows/graphs and /workflows/tasks.  The graphs route is used for graph definitions. ant the tasks route is used for task definitions  Some semantics are now inconsistent with /nodes/:identifier/workflows.  This will be cleaned up in another PR.

See https://groups.google.com/forum/#!topic/rackhd/SvggYYKeeaQ for more detailed discussion.

This change has contributions from @dalebremner @abildw and @srinia6 

Depends on https://github.com/RackHD/on-core/pull/107

### DIFF
--- a/lib/api/2.0/workflowGraphs.js
+++ b/lib/api/2.0/workflowGraphs.js
@@ -5,7 +5,7 @@
 var injector = require('../../../index.js').injector;
 var controller = injector.get('Http.Services.Swagger').controller;
 var workflowApiService = injector.get('Http.Services.Api.Workflows');
-var _ = require('lodash');    // jshint ignore:line
+var _ = injector.get('_');    // jshint ignore:line
 
 /**
 * @api {get} /api/2.0/workflows/graphs GET /workflows/graphs
--- a/lib/api/2.0/workflowGraphs.js
+++ b/lib/api/2.0/workflowGraphs.js
@@ -1,0 +1,67 @@
+// Copyright 2016, EMC Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var workflowApiService = injector.get('Http.Services.Api.Workflows');
+var _ = require('lodash');    // jshint ignore:line
+
+/**
+* @api {get} /api/2.0/workflows/graphs GET /workflows/graphs
+* @apiVersion 2.0.0
+* @apiName workflowsGetGraphs
+* @apiDescription Get list of all graphs
+* @apiSuccess {json} List of graphs
+* @apiGroup workflowGraphs
+*/
+var workflowsGetGraphs = controller(function() {
+    return workflowApiService.getGraphDefinitions();
+});
+
+/**
+* @api {get} /api/2.0/workflows/graphs/{injectableName} GET /workflows/graphs/{injectableName}
+* @apiVersion 2.0.0
+* @apiName workflowsGetGraphsByName
+* @apiDescription Get the graph with the specified injectable name
+* @apiParam {String} injectableName
+* @apiSuccess Graph with the specified injectable name
+* @apiGroup workflowGraphs
+*/
+var workflowsGetGraphsByName = controller(function(req) {
+    return workflowApiService.getGraphDefinitions(req.swagger.params.injectableName.value);
+});
+
+/**
+* @api {put} /api/2.0/workflows/graphs PUT /workflows/graphs
+* @apiVersion 2.0.0
+* @apiName workflowsPutGraphs
+* @apiDescription Add a graph to the graph library
+* @apiParam {Object} body
+* @apiSuccess Graph Created
+* @apiGroup workflowGraphs
+*/
+// Can include name in body to modify a specific graph
+var workflowsPutGraphs = controller({success: 201},function(req) {
+    return workflowApiService.defineTaskGraph(req.body);
+});
+
+/**
+* @api {delete} /api/2.0/workflows/graphs DELETE /workflows/graphs/{injectableName}
+* @apiVersion 2.0.0
+* @apiName workflowsDeleteGraphsByName
+* @apiDescription Delete the graph with the specified injectable name
+* @apiParam {String} injectableName
+* @apiSuccess Graph deleted
+* @apiGroup workflowGraphs
+*/
+var workflowsDeleteGraphsByName = controller(function(req) {
+    return workflowApiService.destroyGraphDefinition(req.swagger.params.injectableName.value); 
+});
+
+module.exports = {
+    workflowsGetGraphs: workflowsGetGraphs,
+    workflowsGetGraphsByName: workflowsGetGraphsByName,
+    workflowsPutGraphs: workflowsPutGraphs,
+    workflowsDeleteGraphsByName: workflowsDeleteGraphsByName
+};
--- a/lib/api/2.0/workflows.js
+++ b/lib/api/2.0/workflows.js
@@ -5,6 +5,7 @@
 var injector = require('../../../index.js').injector;
 var controller = injector.get('Http.Services.Swagger').controller;
 var workflowApiService = injector.get('Http.Services.Api.Workflows');
+var Errors = injector.get('Errors');
 var _ = require('lodash');    // jshint ignore:line
 
 /**
@@ -59,18 +60,32 @@ var workflowsGetById = controller(function(req) {
 });
 
 /**
-* @api {put} /api/2.0/workflows/:identifier PUT /workflows/:identifier/cancel
+* @api {put} /api/2.0/workflows/:identifier PUT /workflows/:identifier/action
 * @apiVersion 2.0.0
-* @apiDescription cancel the specified workflow
-* @apiName workflows-cancel
+* @apiDescription perform the specified action on the selected workflow
+* @apiName workflows-action
 * @apiGroup workflows
-* @apiParam {String} instanceId of workflow
+* @apiParam {String} identifier of workflow
 * @apiSuccess {json}  object.
 * @apiError NotFound There is no workflow with <code>instanceId</code>
 */
+var workflowsAction = controller(function(req) {
+    var command = req.body.command;
+    var options = req.body.options || {};
 
-var workflowsCancel = controller(function(req) {
-    return workflowApiService.cancelTaskGraph(req.swagger.params.identifier.value);
+    var actionFunctions = {
+        cancel: function() {
+            return workflowApiService.cancelTaskGraph(req.swagger.params.identifier.value);
+        }
+    };
+
+    if (_(actionFunctions).has(command)) {
+        return actionFunctions[command]();
+    }
+
+    throw new Errors.BadRequestError(
+        command + ' is not a valid workflow action'
+    );
 });
 
 /**
@@ -93,5 +108,5 @@ module.exports = {
     workflowsPost: workflowsPost,
     workflowsGetById: workflowsGetById,
     workflowsDeleteById: workflowsDeleteById,
-    workflowsCancel: workflowsCancel
+    workflowsAction: workflowsAction
 };
--- a/lib/api/2.0/workflows.js
+++ b/lib/api/2.0/workflows.js
@@ -1,0 +1,97 @@
+// Copyright 2016, EMC Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var workflowApiService = injector.get('Http.Services.Api.Workflows');
+var _ = require('lodash');    // jshint ignore:line
+
+/**
+* @api {get} /api/2.0/workflows GET /workflows
+* @apiVersion 2.0.0
+* @apiDescription Get list of active and past run workflow instances
+* @apiName workflows-get
+* @apiGroup workflows
+* @apiSuccess {json} List of all workflows or if there are none an empty object.
+*/
+
+var workflowsGet = controller(function(req) {
+    return workflowApiService.getActiveWorkflows(req.query)
+    .then(function(workflows) {
+        if (req.swagger.query && _(req.swagger.query).has('active')) {
+            return _(workflows)
+            .filter(function(workflow) {
+                return workflow.active() === req.swagger.query.active;
+            });
+        }
+        return workflows;
+    });
+});
+
+/**
+* @api {post} /api/2.0/workflows POST /workflows
+* @apiVersion 2.0.0
+* @apiDescription Run a new workflow
+* @apiName workflows-run
+* @apiGroup workflows
+* @apiError Error problem was encountered, workflow was not run.
+*/
+
+var workflowsPost = controller({success: 201}, function(req) {
+    var configuration = _.defaults(req.query || {}, req.body || {});
+    return workflowApiService.createAndRunGraph(configuration);
+});
+
+/**
+* @api {get} /api/2.0/workflows/:identifier GET /workflows/:identifier
+* @apiVersion 2.0.0
+* @apiDescription get a specific workflow
+* @apiName workflows-getById
+* @apiGroup workflows
+* @apiParam {String} instanceId of workflow
+* @apiSuccess {json} workflows of the particular identifier or if there are none an empty object.
+* @apiError NotFound There is no workflow with <code>instanceId</code>
+*/
+
+var workflowsGetById = controller(function(req) {
+    return workflowApiService.getWorkflowById(req.swagger.params.identifier.value);
+});
+
+/**
+* @api {put} /api/2.0/workflows/:identifier PUT /workflows/:identifier/cancel
+* @apiVersion 2.0.0
+* @apiDescription cancel the specified workflow
+* @apiName workflows-cancel
+* @apiGroup workflows
+* @apiParam {String} instanceId of workflow
+* @apiSuccess {json}  object.
+* @apiError NotFound There is no workflow with <code>instanceId</code>
+*/
+
+var workflowsCancel = controller(function(req) {
+    return workflowApiService.cancelTaskGraph(req.swagger.params.identifier.value);
+});
+
+/**
+* @api {delete} /api/2.0/workflows/:identifier DELETE /workflows/:identifier
+* @apiVersion 2.0.0
+* @apiDescription Cancel currently running workflows for specified node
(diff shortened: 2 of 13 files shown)


abildw:
+1

brianparry:
test this please

brianparry:
test this please

brianparry:
test this please

benbp:
Resolves https://github.com/RackHD/RackHD/issues/95

lib/services/workflow-api-service.js
```diff
@@ -155,7 +237,32 @@ function workflowApiServiceFactory(
     };
 
     WorkflowApiService.prototype.findActiveGraphForTarget = function(target) {
-        return taskGraphStore.findActiveGraphForTarget(target);
+        return waterline.graphobjects.find({ node: target })
+            .then(function(workflows){
+                return _.filter(workflows, function(wf) {
+                    return wf.active();
```
benbp: Why do this work in memory rather than letting the query in `findActiveGraphForTarget` tell the database to do all the processing?

brianparry: @dalebremner Correct me if I'm wrong, but I think the idea was to make this a bit more behavioral.  The code reads as "filter for active workflows" rather than "query for _status equals pending".  
The reason for not using the task graph store is that `taskGraphStore.findActiveGraphForTarget` goes directly to Mongo which results in returned resources that are inconsistent with the /workflows and /workflows/:id

benbp: In this case though, `_status = "pending" and node = target` is the underlying criteria for an active workflow for the node, and it will only ever be zero or one objects, whereas doing a find all has no upper bound and could potentially be very slow. I get that it's easier to understand what's happening from the `return wf.active()` statement, but that was also my intent with the method name of `findActiveGraphForTarget`.

Re your second point, is there perhaps a way to cast the objects coming direct out of mongo to resemble the objects out of the waterline layer (I thought there wasn't a difference, but if there is we should probably be doing this by default)? What are the differences in the resources?

brianparry: @benbp Yeah, I agree a query makes sense here.  I think the thing that makes this a little confusing is that _status == pending is synonomous with "active".  Maybe we should add an active property to Constants.Task.States?

On the second point, I think the difference was the id property.  The raw Mongo document named this property "_id" whereas waterline omits the underscore.  There may be other differences too, but I'd need to look at it again.

benbp: @brianparry I agree, I think the right thing to do might be to change the state to `Constants.Task.States.Running` since it's a pre-existing state? We didn't do it originally because it saved a query operation, but I think it makes sense given how confusing it is.

Ah ok, yeah I remember making some mistakes/inconsistencies with `id` vs. `_id`, so it's possible I didn't resolve all of them.

benbp:
:+1:

brianparry:
https://github.com/RackHD/on-core/pull/107 is merged, so test this please.

brianparry:
Once again, test this please

lib/api/2.0/workflowGraphs.js
```diff
@@ -0,0 +1,67 @@
+// Copyright 2016, EMC Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var workflowApiService = injector.get('Http.Services.Api.Workflows');
+var _ = require('lodash');    // jshint ignore:line
```
yyscamper: Since `lodash` has already in the injectable tables, so how about using the `injector.get` instead of require it directly?

I am thinking about a possible case to reduce the RackHD footprint in future. If the lodash have already in the on-core's injectable tables, so other modules (on-http, on-taskgraph, on-tasks, ...) can always obtain it via on-core, thus the `lodash` can be removed from on-http's dependencies, avoid to install a duplicated lodash in on-http. 

My point is that if the module has been included in on-core, then we should leverage the on-core to load it.

lib/api/2.0/workflows.js
```diff
@@ -0,0 +1,112 @@
...
+        }
+    };
+
+    if (_(actionFunctions).has(command)) {
+        return actionFunctions[command]();
+    }
+
+    throw new Errors.BadRequestError(
+        command + ' is not a valid workflow action'
+    );
```
yyscamper: If only talk for the functionality of this function, I think your code is right and works well. I just have a little concern for this function's logic because it may give user the first impression that this function will eventually throw an error.
My point is usually the error handling is putting into the branching, not in the main logic flow.
```javascript
if (!_(actionFunctions).has(command)) {
    throw new Errors.BadRequestError(...);
}
return actionFunctions[command]();
```
or put the error handling in the `else`:
```javascript
if (_(actionFunctions).has(command)) {
    return actionFunctions[command]();
} else {
    throw new Errors.BadRequestError(...);
}
```

This may be my personal style, so this is not a blocking problem for me.  But I want to hear your comment about this.

lib/services/workflow-api-service.js
```diff
@@ -128,7 +128,56 @@ function workflowApiServiceFactory(
...
+        return waterline.graphobjects.needByIdentifier(graphId)
+        .then(function(workflow) {
+            instanceId = workflow.instanceId;
+
+            if (workflow.active()) {
+                return taskGraphProtocol.cancelTaskGraph(instanceId);
+            }
+            throw new Errors.TaskCancellationError(
+                graphId + ' is not an active workflow'
+            );
```
yyscamper: This is another case for my above concern about putting error handling into branch. But this case is a little nicer than above because it is not in the end of the function.

lib/services/workflow-api-service.js
```diff
@@ -128,7 +128,56 @@ function workflowApiServiceFactory(
...
+            // Since deleteGraph takes instanceId as an argument, save it
+            // here using a closure so we don't need to do another query later.
+            instanceId = workflow.instanceId;
+            if (workflow.active()) {
+                return taskGraphProtocol.cancelTaskGraph(instanceId);
+            }
+            return instanceId;
+        })
+        .then(function(cancelledGraphId) {
+            if (cancelledGraphId !== instanceId) {
```
yyscamper: I think this judgment is redundant, base on your design, the `cancelledGraphId` and `instanceId` is always the equivalent. So the next throwing will never be executed and this is always a coverage hole for unit-testing.

lib/services/workflow-api-service.js
```diff
@@ -155,7 +236,22 @@ function workflowApiServiceFactory(
...
+
+    WorkflowApiService.prototype.getActiveWorkflows = function(query) {
+        return waterline.graphobjects.find(query);
+    };
+
+    WorkflowApiService.prototype.getWorkflowById = function(id) {
+        return waterline.graphobjects.needByIdentifier(id);
+    };
+
+    WorkflowApiService.prototype.destroyGraphDefinition = function(injectableName) {
```
yyscamper: How above to use the uniform words for the same action of task definition and graph definition?
This is your current mapping:
```
destroyGraphDefinition <--> deleteWorkflowsTasksByName 
getGraphDefinitions  <--> getWorkflowsTasksByName 
defineTaskGraph <--> putWorkflowsTasksByName
```
If we use `destroy` for graph, then it should be better to use `destroy` for task as well; If we add `byName` for task, then it should also be better to add the words for graph.

lib/services/workflow-api-service.js
```diff
@@ -155,7 +236,22 @@ function workflowApiServiceFactory(
...
 
     WorkflowApiService.prototype.findActiveGraphForTarget = function(target) {
-        return taskGraphStore.findActiveGraphForTarget(target);
+        return waterline.graphobjects.findOne({
+            node: target,
+            _status: Constants.Task.ActiveStates
+        });
+    };
+
+    WorkflowApiService.prototype.getActiveWorkflows = function(query) {
```
yyscamper: This API name is not well-matched with its implementation. If the API automatically builds the query to lookup active workflow, then I think the name and implementation is matched well.

spec/lib/api/2.0/workflowGraphs-spec.js
```diff
@@ -0,0 +1,89 @@
...
+    });
+
+   describe('workflowsPutGraphs', function () {
+        it('should persist a graph', function () {
+            var graph = { name: 'foobar' };
+            workflowApiService.defineTaskGraph.resolves(graph);
+
+            return helper.request().put('/api/2.0/workflows/graphs')
+            .send(graph)
+            .expect('Content-Type', /^application\/json/)
```
yyscamper: Also a consistent style concern, for some test cases, this `expect` has 4-spaces indent but sometimes not.

spec/lib/services/workflow-api-service-spec.js
```diff
@@ -250,4 +279,85 @@ describe('Http.Services.Api.Workflows', function () {
...
+        return expect(workflowApiService.getWorkflowById())
+               .to.be.rejectedWith(Errors.NotFoundError);
+    });
+
+    it('should return active workflows ', function() {
+        var activeWorkflow = {
+                               id      : 'testgraphid',
+                               _status :"pending"
+
+                           };
```
yyscamper: A little strange alignment style.

spec/lib/services/workflow-api-service-spec.js
```diff
@@ -250,4 +279,85 @@ describe('Http.Services.Api.Workflows', function () {
...
+        waterline.graphobjects.find.resolves(activeWorkflow);
+        return workflowApiService.getWorkflowById().then(function (workflows) {
+            expect(workflows).to.deep.equal(activeWorkflow);
+        });
+    });
+
+
+
```
yyscamper: remove these unwanted lines.

spec/lib/services/workflow-api-service-spec.js
```diff
@@ -250,4 +279,85 @@ describe('Http.Services.Api.Workflows', function () {
...
+        waterline.graphobjects.needByIdentifier.resolves(workflow);
+        return workflowApiService.getWorkflowById().then(function (workflows) {
+            expect(workflows).to.deep.equal(workflow);
+        });
+    });
+
+    it('should return Not Found Error when invalid id is passed', function() {
+        waterline.graphobjects.needByIdentifier.rejects(new Errors.NotFoundError('Not Found'));
+        return expect(workflowApiService.getWorkflowById())
+               .to.be.rejectedWith(Errors.NotFoundError);
```
yyscamper: forgot 'eventually` for promise

spec/lib/services/workflow-api-service-spec.js
```diff
@@ -250,4 +279,85 @@ describe('Http.Services.Api.Workflows', function () {
...
+    it('should return active workflows ', function() {
+        var activeWorkflow = {
+                               id      : 'testgraphid',
+                               _status :"pending"
+
+                           };
+        waterline.graphobjects.find.resolves(activeWorkflow);
+        return workflowApiService.getWorkflowById().then(function (workflows) {
+            expect(workflows).to.deep.equal(activeWorkflow);
+        });
```
yyscamper: Your writing is good, and this can be short to:
```javascript
return workflowApiService.getWorklfowById().to.become(activeWorkflow);
```

spec/lib/services/workflow-api-service-spec.js
```diff
@@ -250,4 +279,85 @@ describe('Http.Services.Api.Workflows', function () {
...
+    });
+
+    it('should throw error, when cancelling an non-active workflow ', function () {
+        var mockWorkflowError = new Errors.TaskCancellationError(
+            "testid is not an active workflow"
+        );
+        waterline.graphobjects.needByIdentifier.rejects(mockWorkflowError);
+        return workflowApiService.cancelTaskGraph()
+            .should.eventually.be.rejectedWith(mockWorkflowError);
```
yyscamper: The `eventually` should be not written together with reject, please see this issue: https://github.com/RackHD/RackHD/issues/54

spec/lib/api/2.0/workflowTasks-spec.js
```diff
@@ -0,0 +1,173 @@
...
+            .expect(200)
+            .expect(function ( res ) {
+                expect(workflowApiService.getWorkflowsTasksByName).to.have.been.calledOnce;
+                expect(workflowApiService.getWorkflowsTasksByName)
+                    .to.have.been.calledWith(workflowTask.injectableName);
+                expect(res.body).to.have.property('friendlyName', 'dummy');
+                expect(res.body).to.have.property('injectableName', 'dummyName');
+                expect(res.body).to.have.property('options').to.be.an('object');
+                expect(res.body).to.have.deep.property('options.oids', 'SNMPv2-MIB::sysDescr');
+                });
```
yyscamper: incorrect indent

yyscamper:
Some of my comment is about the style, this is not  gating for me, but I think at least following two comments need be fixed:
1. Rename the name of workflow service APIs.
2. `eventually` should not be written together with rejected.

lib/api/2.0/workflows.js
brianparry: @yyscamper Yep - usually I structure code like this like your first example.  I prefer this to including error handling in a conditional because I think it obscures your mainline code.  In this case, I structured it more like a default case in a switch statement.  I'll change it to look like your first suggestion.

lib/services/workflow-api-service.js
brianparry: If the cancel fails, it will return null for cancelled graphId.

lib/services/workflow-api-service.js
brianparry: I'm not sure I follow your concern.  Are you suggesting that _graph_ be replaced with _workflow_?  If so, I agree and I think this applies to many of the functions in this service.  I'd suggest adding an issue and cleaning up in the future.

lib/services/workflow-api-service.js
brianparry: I agree with this.  As above, I think we should consider this a separate task that we track with a Github issue.

lib/services/workflow-api-service.js
brianparry: I'll change it to this to be consistent with the first case.
```javascript
          if (!workflow.active()) {
               throw new Errors.TaskCancellationError(
                    graphId + ' is not an active workflow'
               );
          }
          return taskGraphProtocol.cancelTaskGraph(instanceId);

File: spec/lib/services/workflow-api-service-spec.js (hunk 13 above)
Comment (brianparry) [inline spec/lib/services/workflow-api-service-spec.js]: Will this work?
```javascript
return workflowApiService.cancelTaskGraph()
          .should.be.rejectedWith(mockWorkflowError);
```

or do I need to do this:

```javascript
return workflowApiService.cancelTaskGraph().catch(function(err) {
          expect(err).to.deepEqual(mockWorkflowError);
});
```

spec/lib/api/2.0/workflowGraphs-spec.js
brianparry: I think this is because several developers authored this code.  I think the style varies by developer preference.

spec/lib/services/workflow-api-service-spec.js
benbp: Yeah the former should work fine I believe. The second doesn't test the case where an error isn't thrown, so the first block is the better options.

spec/lib/api/2.0/workflowGraphs-spec.js
benbp: Yeah, I believe I may be the culprit for starting the no indent fad >;)

spec/lib/services/workflow-api-service-spec.js
brianparry: Changing to this

```javascript
return workflowApiService.cancelTaskGraph()
          .should.be.rejectedWith(mockWorkflowError);
```

brianparry:
@yyscamper I agree that the function renaming should be done, but disagree that it should gate the PR.  I'd suggest logging an issue in github and cleaning up the naming later.  I addressed the `.eventually` issue and cleaned up the style in a few places.

spec/lib/api/2.0/workflowGraphs-spec.js
brianparry: Haha @benbp!  I could go either way on this.  I think it looks better _with_ an indent if there are no callback functions passed to the chained methods.
```javascript
expect(foo).to.have.property(bar)
    .and.to.equal(baz)
```

I think it looks better _without_ the indent if there are callback functions.

```javascript
    var foo = _(bar)
    .filter(function(baz) {
        return baz === qux;
    });
```
  I'm consistently inconsistent!

lib/services/workflow-api-service.js
yyscamper: ohh, yes, you are right, forget my this comment.

lib/services/workflow-api-service.js
yyscamper: my meaning is that your function is more like "find workflow by query" not "find the active workflow", if user doesn't input the correct query, then your function works normally but not return the active one, this doesn't match with the function name.

spec/lib/api/2.0/workflowGraphs-spec.js
yyscamper: @brianparry @benbp do you think we should define a consistent style for this?so that we write code in same way.

yyscamper:
@brianparry , I'm ok to move the renaming to another PR, thanks for other change, so this PR is +1 for me.

yyscamper:
I just see the coverage reduce a lot, can you improve this?

yyscamper:
test this please

spec/lib/api/2.0/workflowGraphs-spec.js
brianparry: @yyscamper I don't think there should be any requirement for us all to write code in the same style.  As long as the style is clear and expressive (such as the examples above) I'm fine with it.

lib/services/workflow-api-service.js
brianparry: I see what you mean now ().  The problem is that this function returns all workflows that match the provided query, not just active workflows.  Again, I think most of these functions need to be renamed to follow a consistent convention.

brianparry:
test this please

brianparry:
test this please

PR #195, opened 2016-03-29
removing Versions from 2.0

abildw:
removing Versions route from 2.0 since it is not successfully implemented right now.
There is also not currently a functional test for the Versions route so this PR should take care of everything needed to remove support.

@brianparry @geoff-reid @srinia6 @dalebremner @amymullins @RackHD/corecommitters


zyoung51:
+1

brianparry:
:+1:

PR #196, opened 2016-03-29
Add northbound sku support

zyoung51:
- Add parameter handler fittings to perform parameter translation/setup
  functions and add a handler that will extract a node id and setup the
  res.locals.scope to enable northbound sku views.
- Update redfish template scope to use the pre-established res.locals.scope
- Update redfish rendering to add sku/env as done in common-api-presenter

### DIFF
--- a/lib/fittings/swagger_locals.js
+++ b/lib/fittings/swagger_locals.js
@@ -1,0 +1,45 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var injector = require('../../index.js').injector;
+var _ = injector.get('_');    // jshint ignore:line
+var Promise = injector.get('Promise');    // jshint ignore:line
+var debug = require('debug')('swagger:locals'); // jshint ignore:line
+var Error = injector.get('Errors'); // jshint ignore:line
+
+var skuParamHandler = function(context, value) {
+    var injector = require('../../index.js').injector;
+    var nodeServiceApi = injector.get('Http.Services.Api.Nodes');
+    return nodeServiceApi.getNodeById(value)
+    .then(function(node) {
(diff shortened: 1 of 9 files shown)


lib/fittings/swagger_locals.js
```diff
@@ -0,0 +1,45 @@
...
+
+'use strict';
+
+var injector = require('../../index.js').injector;
+var _ = injector.get('_');    // jshint ignore:line
+var Promise = injector.get('Promise');    // jshint ignore:line
+var debug = require('debug')('swagger:locals'); // jshint ignore:line
+var Error = injector.get('Errors'); // jshint ignore:line
+
+var skuParamHandler = function(context, value) {
```
brianparry: I'm thinking it may make sense to use the fittingDef to specify the set of anchor functions.  Maybe add an anchorModulesDir(s) property and an anchorFunctions property?

zyoung51: Something like another layer of dependency injection or something like the router controllers?  Either way, seems like there would be some value in supporting that later.

tannoa2:
+1

lib/fittings/swagger_locals.js
brianparry: @zyoung51 Yeah - that's what I was thinking.  Definitely something that can wait until later (maybe when we need to add another anchor function).

brianparry:
:+1:

brianparry:
test this please

jlongever:
Sorry, flubbed an update to the smoke test while adding slave nodes... all set now.  test this please

brianparry:
Jenkins, test this please

PR #200, opened 2016-03-31
config route refactor

abildw:
* updating config route to use swagger controller 

@dalebremner @geoff-reid @brianparry @srinia6 @RackHD/corecommitters

### DIFF
--- a/lib/api/2.0/config.js
+++ b/lib/api/2.0/config.js
@@ -3,19 +3,19 @@
 'use strict';
 
 var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var config = injector.get('Http.Services.Api.Config');
+
+var configGet = controller(function(req) {
+    return config.configGetAll(req.query);
+});
+
+var configPatch = controller(function(req) {
+    return config.configSet(req.body);
+});
 
 module.exports = {
-    getConfig: getConfig,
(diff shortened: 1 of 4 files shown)


lib/api/2.0/config.js
```diff
@@ -3,19 +3,19 @@
...
 
 var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var config = injector.get('Http.Services.Api.Config');
 
-module.exports = {
-    getConfig: getConfig,
-    patchConfig: patchConfig
-};
+function configGet = controller(function(req) {
```
zyoung51: should be 'var configGet'.  Same comment for the issue below.

lib/services/config-api-service.js
```diff
@@ -39,7 +39,7 @@ function configApiServiceFactory(
 
     ConfigApiService.prototype.configSet = function(req) {
```
brianparry: The service should just take a config object, not a request.

brianparry: Nevermind, this is doing the right thing.  I'd change the name though. `req` implies a request object to me.

abildw:
@brianparry changed function parameter name from "req" to "config" to avoid confusion

srinia6:
+1

brianparry:
:+1:

PR #202, opened 2016-04-01
Master-Delete-Tag option for all nodes with a specified tagName

uppalk1:
* Add swagger API monorail 2.0 support in nodes.js for masterDelTagById
* Add support for masterDelTagById in nodes-api-service.js
* Update monorail-2.0.yaml for master-delete-tag option under /nodes/tags/{tagName}

### DIFF
--- a/lib/api/2.0/nodes.js
+++ b/lib/api/2.0/nodes.js
@@ -85,7 +85,7 @@ var nodesPatchTagById = controller(function(req) {
 
 var nodesMasterDelTagById = controller(function(req) {
     return nodes.masterDelTagById(req.swagger.params.tagName.value);
-})
+});
 
 module.exports = {
     nodesGetAll: nodesGetAll,
--- a/lib/api/2.0/nodes.js
+++ b/lib/api/2.0/nodes.js
@@ -80,7 +80,7 @@ var nodesDelTagById = controller(function(req) {
 
 var nodesPatchTagById = controller(function(req) {
     return nodes.addTagsById(req.swagger.params.identifier.value,
-                             req.swagger.params.body.value.tags);
+                             req.body.tags);
 });
 
 var nodesMasterDelTagById = controller(function(req) {
--- a/lib/api/2.0/nodes.js
+++ b/lib/api/2.0/nodes.js
@@ -79,9 +79,13 @@ var nodesDelTagById = controller(function(req) {
 });
 
 var nodesPatchTagById = controller(function(req) {
-    return nodes.addTagsById(req.swagger.params.identifier.value, 
+    return nodes.addTagsById(req.swagger.params.identifier.value,
                              req.swagger.params.body.value.tags);
 });
+
+var nodesMasterDelTagById = controller(function(req) {
+    return nodes.masterDelTagById(req.swagger.params.tagName.value);
+})
 
 module.exports = {
     nodesGetAll: nodesGetAll,
@@ -101,5 +105,6 @@ module.exports = {
     nodesDelActiveWorkflowById: nodesDelActiveWorkflowById,
     nodesGetTagsById: nodesGetTagsById,
     nodesDelTagById: nodesDelTagById,
-    nodesPatchTagById: nodesPatchTagById
+    nodesPatchTagById: nodesPatchTagById,
+    nodesMasterDelTagById: nodesMasterDelTagById
 };
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -520,18 +520,11 @@ function nodeApiServiceFactory(
         .then(function() {
             return waterline.nodes.findByTag(tagName);
         })
-        .then(function(nodes) {
-            nodes.map(function (node) {
-                return waterline.nodes.remTags(node.id, tagName);
-            })
-            return nodes;
-        })
-        .then(function(nodes) {
-            var nodeIdlist = nodes.map(function(node){
-                return node.id;
-            })
-            return nodeIdlist;
-        })
+        .map(function(node) {
+            return waterline.nodes.remTags(node.id, tagName)
+                .then(function()
+                {return node.id; });
+        });
     };
 
 
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -515,7 +515,7 @@ function nodeApiServiceFactory(
      */
     NodeApiService.prototype.masterDelTagById = function(tagName) {
         return Promise.resolve().then(function() {
-                assert.string(tagName, 'tag must be a string');
+            assert.string(tagName, 'tag must be a string');
         })
         .then(function() {
             return waterline.nodes.findByTag(tagName);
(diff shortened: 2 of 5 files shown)


uppalk1:
@amymullins @jlongever @tannoa2 @zyoung51 @keedya

lib/services/nodes-api-service.js
```diff
@@ -509,6 +509,33 @@ function nodeApiServiceFactory(
...
+     * @return {Promise}
+     */
+    NodeApiService.prototype.masterDelTagById = function(tagName) {
+        return Promise.resolve().then(function() {
+                assert.string(tagName, 'tag must be a string');
+        })
+        .then(function() {
+            return waterline.nodes.findByTag(tagName);
+        })
+        .then(function(nodes) {
```
zyoung51: This could also be written as:
.map(function(node) {
  return node.remTags(node.id, tagName).then(function() { return node.id; });
});

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -736,8 +737,16 @@ describe("Http.Services.Api.Nodes", function () {
...
+
         var node = {
-            id: '1234abcd1234abcd1234abce'
+            id: '1234abcd1234abcd1234abcd',
+            name: 'name1',
+            type: 'compute',
+        };
+        var node1 = {
+            id: '5678efgh5678efgh5678efgh',
+            name: 'name1',
```
zyoung51: findByTags is stubbed such that the tag is not actually needed, but for what its worth, the tags are in the 'tags' property, but the test references the value of the 'name' property, which may be confusing.  Can you change this to be "tags: [ 'name1' ]  " ?

lib/api/2.0/nodes.js
```diff
@@ -79,10 +79,14 @@ var nodesDelTagById = controller(function(req) {
 });
 
 var nodesPatchTagById = controller(function(req) {
-    return nodes.addTagsById(req.swagger.params.identifier.value, 
+    return nodes.addTagsById(req.swagger.params.identifier.value,
                              req.swagger.params.body.value.tags);
```
brianparry: @uppalk1 You can simplify this to req.body.tags since the body parameter has generic object schema.

lib/services/nodes-api-service.js
```diff
@@ -509,6 +509,26 @@ function nodeApiServiceFactory(
     };
 
     /**
+     * Remove the tag from a list of nodes
+     * @param  {String}     tagName
+     * @return {Promise}
+     */
+    NodeApiService.prototype.masterDelTagById = function(tagName) {
+        return Promise.resolve().then(function() {
+                assert.string(tagName, 'tag must be a string');
```
brianparry: @uppalk1 This assert doesn't need to be in the Promise chain.  You could simplify this code a bit by moving the assert out of the promise chain and beginning the chain with `return waterline.nodes.findByTag(tagName).map(...)`

zyoung51: @uppalk1 Be careful making that change.  An assert outside of the promise chain will generate a synchronous error and will be missed by a .catch().  Promise chains can be started with Promise.try() to handle that case.

brianparry: @zyoung51: This function is called in a controller callback which is always called inside a promise chain so this error would be caught by the .catch in the controller wrapper function.  That said, I think you're right.  We should probably just leave this function as is since in case it is ever used outside a controller.

spec/lib/api/2.0/nodes-spec.js
```diff
@@ -722,5 +724,14 @@ describe('Http.Api.Nodes', function () {
                 });
         });
 
+        it('should call masterDelTagById', function() {
+            return helper.request().delete('/api/2.0/nodes/tags/name')
```
brianparry: I'd suggest having nodesApi.masterDelTagById resolve to an array of mock Ids and then expect delete response body to match that array.

brianparry:
:+1: after missing semi-colon fix.

brianparry:
test this please

lib/services/nodes-api-service.js
```diff
@@ -509,6 +509,26 @@ function nodeApiServiceFactory(
...
+        return Promise.resolve().then(function() {
+            assert.string(tagName, 'tag must be a string');
+        })
+        .then(function() {
+            return waterline.nodes.findByTag(tagName);
+        })
+        .map(function(node) {
+            return waterline.nodes.remTags(node.id, tagName)
+                .then(function()
+                {return node.id; });
```
yyscamper: A style suggestion, Could you please rewrite this code though it has only one line, just like what you did in line#543.

yyscamper:
Still :+1: , if my minor style suggestion is resolved, that will be better.

yyscamper:
Meanwhile, could you please teach me why the API name needs the word `master` (masterDelTagById)? why not use the `delTagById`or `delTagByName`?

lib/services/nodes-api-service.js
```diff
@@ -509,6 +509,26 @@ function nodeApiServiceFactory(
     };
 
     /**
+     * Remove the tag from a list of nodes
+     * @param  {String}     tagName
+     * @return {Promise}
+     */
+    NodeApiService.prototype.masterDelTagById = function(tagName) {
```
yyscamper: maybe need a rename for the API name, this is to delete tag by the name, not by id

uppalk1: The term Master is used to imply the fact that its a "master delete" of the specified tag from all nodes that have it. This is an add-on to already existing "delete a tag from a specific node",

Agree that it is by tagName and not Id.

jlongever:
Jenkins, test this please

lib/services/nodes-api-service.js
yyscamper: Thanks @uppalk1 , my understanding for tag is that it will be assigned a mongo unique identifier, but the tag name is not unique, because in  the database schema, there is not an unique flag for tag name like what we did for the injectableName of [taskDefinition] (https://github.com/RackHD/on-core/blob/master/lib%2Fmodels%2Ftask-definition.js#L24) , 
Do you mean the tag name should also be unique? If so we need an unique flag for tag schema.

Or the API is just to delete tag by identifier, but you named the argument with "tagName"? 

Anyway, my meaning is that keeping the API name and argument be consistent, both use `id` or both use `name`, then it will not confuse other developers.

Add @zyoung51

zyoung51: The tagName is enforced as unique in the code.  The model should be updated to add the unique tag.

The term id is used to refer interchangeably to nodeid, graphid, taskid, etc throughout the code regardless of them being mongo id, uuid, etc.  In this case, 'id' is being used to refer correctly to the tag's primary key (which happens to be a stringid)

yyscamper: Thanks @zyoung51 , make sense, please add a follow-up to update the tag model.

PR #205, opened 2016-04-05
fix unit-test: eventually should not be together with reject

yyscamper:
I had found an odd problem, if putting the `eventually` together with `reject` in an unit-testing assertion, both positive and negative judgement will pass. Please see this issue for detail: https://github.com/RackHD/RackHD/issues/54.

I tend to think `Chai-as-Promise` need to fix this, as people is difficult to realize this problem. I will submit an issue for `Chai-as-Promise`. However, at present, we should avoid to write `eventually` with `reject`.

@RackHD/corecommitters @srinia6 @iceiilin @WangWinson @pengz1 @sunnyqianzhang @zyoung51

### DIFF
--- a/spec/lib/services/pollers-api-service-spec.js
+++ b/spec/lib/services/pollers-api-service-spec.js
@@ -52,7 +52,7 @@ describe("Http.Services.Api.Pollers", function () {
         it('should return error if poller informations is not found', function () {
             var mockPollerError = new Errors.NotFoundError("Could not find workitem with identifier");
             waterline.workitems.find.rejects(mockPollerError);
-            return pollerService.getPollers().should.eventually.be.rejectedWith(mockPollerError);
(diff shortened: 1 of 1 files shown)


yyscamper:
I have created an issue for Chai-as-Promised: https://github.com/domenic/chai-as-promised/issues/146

brianparry:
:+1:

heckj:
:+1: (and thank you for opening the bug in Chai-as-Promised as well)

yyscamper:
The author of Chai-as-Promised just replied me, he said this was expected behavior, we should not comprehend the word `eventually` as its literal meaning :

    eventually means: apply the following assertion to the fulfillment value of the promise.

    rejectedWith means: assert that this promise rejects with the given value.

    Combined, you have: assert that this promise fulfills with a promise that rejects with the given value.

    Since promises can never fulfill with promises, this will of course never fire.

    Working as intended.

PR #206, opened 2016-04-06
API 2.0 /views route

brianparry:
Views are a collection of ejs templates used for rendering resources into views.  The intent of this change is to logically separate templates used for workflows from templates used for presentation of resources.  Additional PRs to integrate views with the render fitting and to add views for various resources are forthcoming.

Depends on https://github.com/RackHD/on-core/pull/109

### DIFF
--- a/lib/api/2.0/views.js
+++ b/lib/api/2.0/views.js
@@ -1,0 +1,99 @@
+// Copyright 2016, EMC Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var views = injector.get('Views');
+var Promise = injector.get('Promise');
+var Errors = injector.get('Errors');
+var _ = injector.get('_');
+var bodyParser = require('body-parser');
+
+/**
+* @api {get} /api/2.0/views GET /views
+* @apiVersion 2.0.0
+* @apiDescription Get list of views
+* @apiName views-get
+* @apiGroup views
+* @apiSuccess {json} List of all views or if there are none an empty collection.
+*/
+var viewsGet = controller(function() {
+    return views.getAll();
+});
+
+/**
+* @api {get} /api/2.0/views/:identifier GET /views/:identifier
+* @apiVersion 2.0.0
+* @apiDescription get a specific view
+* @apiName views-getById
+* @apiGroup views
+* @apiParam {String} name of view
+* @apiSuccess {json} view with specified identifier.
+* @apiError NotFound There is no view with specified name
+*/
+var viewsGetById = controller(function(req) {
+    return views.get(req.swagger.params.identifier.value)
+    .then(function(view) {
+        if (_.isEmpty(view)) {
+            throw new Errors.NotFoundError(
+                'Could not find ' + req.swagger.params.identifier.value
+            );
+        }
+        return view;
+    });
+});
+
+/**
+* @api {put} /api/2.0/views/:identifier PUT /views/:identifer
+* @apiVersion 2.0.0
+* @apiDescription create or update a view
+* @apiName views-put
+* @apiGroup views
+* @apiParam {String} name of view
+* @apiSuccess {json} created or updated view.
+* @apiError Error was encountered, view is unchanged.
+*/
+var viewsPut = controller(function(req, res) {
+    var contentType = req.get('Content-Type');
+
+    var parsers = {
+        'application/octet-stream': Promise.promisify(bodyParser.raw({type: '*/*'})),
+        'text/plain': Promise.promisify(bodyParser.text({type: '*/*'}))
+    };
+
+    if (!_(parsers).has(contentType)) {
+        throw Errors.BadRequestError(contentType + ' is not a valid data type');
+    }
+
+    return parsers[contentType](req, res).then(function() {
+        return views.put(req.swagger.params.identifier.value, req.body.toString());
+    });
+});
+
+/**
+* @api {delete} /api/2.0/views/:identifier DELETE /views/:identifier
+* @apiVersion 2.0.0
+* @apiDescription Unlink specified view
+* @apiName views-delete
+* @apiGroup views
+* @apiParam {String} name of view
+* @apiError NotFound There is no view with specified name
+*/
+var viewsDelete = controller({send204OnEmpty: true}, function(req, res) {
+    return views.unlink(req.swagger.params.identifier.value)
+    .then(function(view) {
+        if (_.isEmpty(view)) {
+            throw new Errors.NotFoundError(
+                'Could not delete ' + req.swagger.params.identifier.value
+            );
+        }
+    });
+});
+
+module.exports = {
+    viewsGet: viewsGet,
+    viewsGetById: viewsGetById,
+    viewsPut: viewsPut,
+    viewsDelete: viewsDelete
+};
--- a/lib/api/2.0/views.js
+++ b/lib/api/2.0/views.js
@@ -80,7 +80,7 @@ var viewsPut = controller(function(req, res) {
 * @apiParam {String} name of view
 * @apiError NotFound There is no view with specified name
 */
-var viewsDelete = controller({send204OnEmpty: true}, function(req, res) {
+var viewsDelete = controller({send204OnEmpty: true}, function(req) {
     return views.unlink(req.swagger.params.identifier.value)
     .then(function(view) {
         if (_.isEmpty(view)) {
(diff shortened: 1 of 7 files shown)


brianparry:
@RackHD/corecommitters @srinia6 @geoff-reid @dalebremner @abildw

brianparry:
Test failures should be resolved when https://github.com/RackHD/on-core/pull/109 is merged

lib/api/2.0/views.js
```diff
@@ -0,0 +1,99 @@
...
+* @api {get} /api/2.0/views/:identifier GET /views/:identifier
+* @apiVersion 2.0.0
+* @apiDescription get a specific view
+* @apiName views-getById
+* @apiGroup views
+* @apiParam {String} name of view
+* @apiSuccess {json} view with specified identifier.
+* @apiError NotFound There is no view with specified name
+*/
+var viewsGetById = controller(function(req) {
```
zyoung51: The 'name & scope' are the primary key for the model, so the name is not strictly the identifier.  Should this be called 'viewsGetByName'?  Do you want another method that also consider scope?

lib/api/2.0/views.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+* @api {delete} /api/2.0/views/:identifier DELETE /views/:identifier
+* @apiVersion 2.0.0
+* @apiDescription Unlink specified view
+* @apiName views-delete
+* @apiGroup views
+* @apiParam {String} name of view
+* @apiError NotFound There is no view with specified name
+*/
+var viewsDelete = controller({send204OnEmpty: true}, function(req, res) {
```
zyoung51: Do you want to delete the file in the view directory as well?  The view will be reloaded into the database when the service restarts.

lib/api/2.0/views.js
brianparry: Right now, I'm only using global scope.  I'm considering using scope to link resources to the set of views that can be used to present them (e.g. all views with scope 'node' could be used to render node resources).  To do this, I'd probably add views/<scope>/ and views/<scope>/<name> and GetById would become GetScopeByName or something along those lines.

lib/api/2.0/views.js
brianparry: This is actually a problem for the put method as well.  When the service is restarted, any modifications to a view are replaced by the content of the view file.  I'd like to use the files in /data/views to seed the database the first time the service starts and rely solely on the database thereafter.  That should limit confusion about the 'source of truth' and eliminate the need to manipulate files from the views routes.  Haven't looked into whether this approach would actually be feasible though.

brianparry:
test this please

brianparry:
test this please

lib/api/2.0/views.js
```diff
@@ -0,0 +1,99 @@
+// Copyright 2016, EMC Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var views = injector.get('Views');
+var Promise = injector.get('Promise');
```
yyscamper: This warning can be ignored, since we have already pre-defined the `Promise`, `_` and some other variables in jshint config file.

spec/lib/api/2.0/views-spec.js
```diff
@@ -0,0 +1,147 @@
...
+
+        return helper.startServer([
+            dihelper.simpleWrapper(viewsProtocol, 'Views')
+        ]).then(function() {
+            _ = helper.injector.get('_');
+        });
+    });
+
+    beforeEach('reset stubs', function () {
+        _(viewsProtocol).methods().forEach(function (method) {
```
yyscamper: you can simplify this design by putting all these stubs into a sinon.sandbox.

spec/lib/api/2.0/views-spec.js
```diff
@@ -0,0 +1,147 @@
...
+        });
+
+        it('should create an application/octet-stream view', function() {
+            var view = {name: 'foo', content: 'foo', scope: 'global'};
+            var req = helper.request().put('/api/2.0/views/foo');
+
+            viewsProtocol.put.resolves(view);
+            req.set('Content-Type', 'application/octet-stream');
+            req.write(Buffer('{ "message": "hello" }', 'ascii'));
+            req.end(function(err, res) {
```
yyscamper: Since the request is not returned, the mocha may not catch up the assertion error.

spec/lib/api/2.0/views-spec.js
brianparry: @yyscamper Personally, I prefer to do it this way because I think this technique can be understood by most developers without needing to reference sinon documentation.

spec/lib/api/2.0/views-spec.js
brianparry: @yyscamper Returning the request results in a double callback.  I verified that a false assertion within the `end` callback causes the test to fail.

lib/api/2.0/views.js
brianparry: @zyoung51 I'll address this in a seperate PR as we discussed yesterday.

brianparry:
test this please

spec/lib/api/2.0/views-spec.js
benbp: @brianparry I think doing it the way you're doing it is valid, but let me add another argument in favor of the sandbox approach (as a general approach, not asking you to change anything):

The `methods().forEach((method) => { method.reset() })` approach is vulnerable to someone doing: `method.restore()` in their test, which is something I commonly do when I need the real version of a method so I can test its code path. Also I think that it ends up being easier to reason about what is being stubbed and what is not in a unit test when the test starts with something like:

```
this.sandbox.stub(myClass, 'myMethod1');
this.sandbox.stub(myClass, 'myMethod2');
```

Because of this, it's defined in the unit test what is stubbed, rather than having to refer to the one or more before blocks that meta-program a bunch of stubs (and usually not directly by method name, but by iteration). With the `methods().forEach()` approach, I think it's actually much harder to track down what is a stub and what is not, because `viewsProtocol[method]` doesn't tell me at first glance that `viewsProtocol.myMethod` is stubbed. This can get even harder when there are multiple describe and beforeEach blocks doing things like this to different objects.

brianparry: @benbp In the case where stubs are monkey patched into the actual module, I totally agree.  In this case though, I'm replacing the whole dependency with a mock, so there is nothing to restore.  There also shouldn't be any need to refer back to before or beforeEach blocks since it should be clear that all viewsProtocol methods are mocked.

srinia6:
+1

spec/lib/api/2.0/views-spec.js
benbp: Ah yep, makes sense.

spec/lib/api/2.0/views-spec.js
```diff
@@ -0,0 +1,147 @@
...
+            var views = [
+                {name: 'foo', content: 'foo', scope: 'global'},
+                {name: 'bar', content: 'bar', scope: 'global'}
+            ];
+            viewsProtocol.getAll.resolves(views);
+            return helper.request().get('/api/2.0/views')
+                .expect('Content-Type', /^application\/json/)
+                .expect(200)
+                .expect(function(res) {
+                    _(res.body).forEach(function(view, i) {
```
benbp: Don't you need to call `.value()` here, otherwise the expect statement/function body won't get executed?

spec/lib/api/2.0/views-spec.js
```diff
@@ -0,0 +1,147 @@
...
+        });
+
+        it('should create an application/octet-stream view', function() {
+            var view = {name: 'foo', content: 'foo', scope: 'global'};
+            var req = helper.request().put('/api/2.0/views/foo');
+
+            viewsProtocol.put.resolves(view);
+            req.set('Content-Type', 'application/octet-stream');
+            req.write(new Buffer('{ "message": "hello" }', 'ascii'));
+            req.end(function(err, res) {
```
benbp: I know less about this behavior, but this looks like it gets called asynchronously? In which case the test won't actually catch these errors?

brianparry: @benbp Felix had the same comment which was somehow deleted..  I was originally returning req but I was getting double callback warnings.  I confirmed that an assertion in the end callback will cause the test to fail.

benbp:
A couple questions re: some tests and whether they work, but if I'm incorrect or they're resolved, 👍

spec/lib/api/2.0/views-spec.js
brianparry: @benbp Yep - those asserts will never fire.  In fact, deepEqual, should be deep.equal, so there are a couple errors in here.  I'll push a new patch momentarily.

spec/lib/api/2.0/views-spec.js
benbp: Cool.

PR #208, opened 2016-04-07
Support additional params for Ubuntu OS install

WangWinson:
The update is to support additional parameters same as CentOS install 
related issue : https://github.com/RackHD/RackHD/issues/50
Dependency PR: https://github.com/RackHD/on-tasks/pull/156

- static network configurations
- multi user creation
- root password
- remote repository location
- OS version for multiple versions of Ubuntu

@RackHD/corecommitters @pengz1 @cgx027 

Note: the spec/doc will be update in another PR, since it's common for all OS install


anhou:
@WangWinson could you also add PR dependency, spec link, github issue link to PR message?

anhou:
:+1:

PR #210, opened 2016-04-08
Enable network switch active discovery

benbp:
This PR includes changes made to enable active discovery of network switches (Cisco, Arista, etc.).

The bulk of the change is adding a new route to the southbound profiles api (/profiles/switch/:vendor), which enables us to use much of our existing discovery code but tailored for switches.

The taskrunner.py profile that has been added behaves in much the same way as the bootstrap.js template that we use for existing interaction with the discovery microkernel. This way we can re-use all the pre-existing linux-command job and task option data structures (via `downloadUrl`). The biggest difference is it only supports downloading other python scripts and running them, rather than shelling out bash commands (which isn't a very reliable or consistent approach across different switch environments).

In my experience pretty much all scripts support running python scripts within these bootstrap environments, so I figure that is a good common denominator to use across all switch vendors for discovery.

FYI @dpasupathi

Requires https://github.com/RackHD/on-tasks/pull/158, https://github.com/RackHD/on-taskgraph/pull/84 and https://github.com/RackHD/on-tftp/pull/34. 
https://github.com/RackHD/on-dhcp-proxy/pull/32 and https://github.com/RackHD/on-skupack/pull/10 are relevant but not blocking.

@RackHD/corecommitters @johren @pscharla @heckj @VulpesArtificem @zyoung51

### DIFF
--- a/data/profiles/taskrunner.py
+++ b/data/profiles/taskrunner.py
@@ -59,13 +59,24 @@ while True:
             task['error'] = str(error)
             break
 
-    if "exit" in task_data.keys():
-        print "Task execution complete"
-        sys.exit(int(task_data["exit"]))
+    try:
+        if "exit" in task_data.keys():
+            print "Task execution complete"
+            sys.exit(int(task_data["exit"]))
+        task_data = json.dumps(task_data)
+        req = urllib2.Request(TASKS_URI, task_data, json_content_type)
+    except Exception as error:
+        task_data = [{'error': str(error)}]
+        task_data = json.dumps(task_data)
+        req = urllib2.Request(TASKS_URI, task_data, json_content_type)
 
-    print "Posting task data\n {}".format(task_data)
-    task_data = json.dumps(task_data)
-    req = urllib2.Request(TASKS_URI, task_data, json_content_type)
-    urllib2.urlopen(req)
+    for _ in range(3):
+        try:
+            print "Posting task data\n {}".format(task_data)
+            urllib2.urlopen(req)
+        except urllib2.URLError:
+            sleep(TASK_REQUEST_PERIOD)
+            continue
+        break
 
     sleep(TASK_REQUEST_PERIOD)
--- a/data/profiles/taskrunner.py
+++ b/data/profiles/taskrunner.py
@@ -6,6 +6,7 @@ Download and execute python scripts from the RackHD tasks api in a loop
 
 import urllib2
 import json
+import sys
 from time import sleep
 
 TASK_REQUEST_PERIOD = 5
@@ -57,6 +58,10 @@ while True:
             print "Failure running task {}".format(error)
             task['error'] = str(error)
             break
+
+    if "exit" in task_data.keys():
+        print "Task execution complete"
+        sys.exit(int(task_data["exit"]))
 
     print "Posting task data\n {}".format(task_data)
     task_data = json.dumps(task_data)
--- a/lib/services/profiles-api-service.js
+++ b/lib/services/profiles-api-service.js
@@ -78,7 +78,7 @@ function profileApiServiceFactory(
         return Promise.resolve();
     };
 
-    ProfileApiService.prototype.getNode = function(macAddresses) {
+    ProfileApiService.prototype.getNode = function(macAddresses, type) {
         var self = this;
         return waterline.nodes.findByIdentifier(macAddresses)
         .then(function (node) {
@@ -91,7 +91,7 @@ function profileApiServiceFactory(
                             return node;
                         })
                         .catch(function() {
-                            return self.runDiscovery(node);
+                            return self.runDiscovery(node, type);
                         });
                     } else {
                         // We only count a node as having been discovered if
@@ -102,22 +102,38 @@ function profileApiServiceFactory(
 
                 });
             } else {
-                return self.createNodeAndRunDiscovery(macAddresses);
+                return self.createNodeAndRunDiscovery(macAddresses, type);
             }
         });
     };
 
-    ProfileApiService.prototype.runDiscovery = function(node) {
+    ProfileApiService.prototype.runDiscovery = function(node, type) {
         var self = this;
 
-        var configuration = {
-            name: 'Graph.SKU.Discovery',
-            options: {
-                defaults: {
-                    graphOptions: {
-                        target: node.id
-                    },
-                    nodeId: node.id
+        if (type === 'switch') {
+            logger.info("====== PAUL: Running switch discovery graph here ======");
+            // TODO: Run ZTP graph and Graph.Switch.Discovery here!!!
+            var configuration = {
+                name: 'Graph.Switch.Discovery',
+                options: {
+                    defaults: {
+                        graphOptions: {
+                            target: node.id
+                        },
+                        nodeId: node.id
+                    }
+                }
+            }
+        } else {
+            var configuration = {
+                name: 'Graph.SKU.Discovery',
+                options: {
+                    defaults: {
+                        graphOptions: {
+                            target: node.id
+                        },
+                        nodeId: node.id
+                    }
                 }
             }
         };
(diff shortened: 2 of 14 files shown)


jlongever:
Hi @benbp,  Looks like node discovery is timing out.  Looks consistent between PR builds:
http://rackhdci.lss.emc.com/job/Templates/job/smoke-test/1125/console
http://rackhdci.lss.emc.com/job/Templates/job/smoke-test/1130/console

benbp:
Thanks @jlongever I'll take a look

benbp:
@jlongever whoops, I was marking compute nodes as switch nodes :)

benbp:
test this please

jlongever:
Thanks @benbp.   ^^ Random workflow delete failure... test this please

lib/services/profiles-api-service.js
```diff
@@ -102,41 +102,80 @@ function profileApiServiceFactory(
...
-        return workflowApiService.createAndRunGraph(configuration)
-        .then(function() {
-            return self.waitForDiscoveryStart(node.id);
-        })
-        .then(function() {
-            return node;
-        });
+        vendor = vendor.toLowerCase();
+
+        if (vendor === 'cisco') {
```
jlongever: Would it be better to call vendor specific workflows over hard coding vendor types in support of a generic workflow definition?  In other words, could 'Graph.Switch.Discovery.Cisco.Poap' be invoked directly based on the profile set for specific vendors (perhaps based on MAC prefix or vendor ID)?

benbp: @jlongever yeah I think this is the weakest part of the PR right here, though I also figured we might want to start with something really basic and then evolve the abstraction as our needs grow? 

We do sort vendor type based on mac prefix right now, it's just done in the dhcp configuration file rather than here (since we have to check the mac at that point anyways to know whether it's a switch or a compute node).

The overall workflow that's run **is** a generic one, while there is a nested cataloging workflow that is run as a vendor specific one. Another way to do this that I thought about would be to run a completely generic cataloging workflow, and enforce being vendor agnostic in the python scripts, though I think that has the potential to get pretty messy very quickly.

But as far as a better place for putting vendor type mappings, I haven't come up with a solution yet.

Thoughts?

jlongever: Could there be a generic script that is executed on the switch and pulls down it's relevant vendor specific script? (via downloadUrl maybe?).  That way all the vendor checks are contained to the switch and we still avoid munging scripts?

benbp: @jlongever Like, move the script declarations out of the workflow and into a big python script? If I'm understanding correctly, I think inverting the control like this would probably mean we'd have to reinvent a few wheels that we already have?

This makes me think of another approach though, which could avoid the issue of hardcoding vendor types. What if we enforce a naming convention for cataloging scripts (e.g. `catalog_<vendor>_switch_<catalog-type>.py`), and then have a single cataloging workflow (e.g. catalog vendor task, catalog snmp task, etc.) that takes a vendor type as an option, and updates the downloadUrl, like (rough graph pseudo-definition):

```
// graph definition
{
    <stuff>,
    options: {
        defaults: {
            switchvendor: <set dynamically based on API params>
        }
    },
    tasks: [
        {
            label: "catalog-version",
            <stuff>,
            options: {
                switchvendor: null,
                downloadUrl: "{{ api.server }}/templates/catalog_{{ options.switchvendor }}_switch_version.py"
        },
        {
            label: "catalog-snmp",
            <stuff>,
            options: {
                switchvendor: null,
                downloadUrl: "{{ api.server }}/templates/catalog_{{ options.switchvendor }}_switch_snmp.py"
        }
    ]
}
```

This way, vendor types aren't hardcoded, just passed through dynamically, and we have one cataloging workflow definition to rule them all, rather than one per vendor.

jlongever: I didn't exactly mean move everything into a big python script, but a common script that pulled down the right script based on the downloadUrl input?  Sounds like you're suggesting that but using a single workflow definition that takes different options.  Sounds good to me,  but why put a restriction on the script naming and instead pass the filename in explicitly?

pscharla: That totally makes sense to me.  If I understand correctly, you are talking about doing something like this:

if (vendor === 'cisco') {
            configuration.options.switchvendor = 'cisco';
        } else if (vendor === 'arista') {
            configuration.options.switchvendor = 'arista'';
        } else {

Then we can collapse Graph.Switch.Discovery.Cisco.Poap, Graph.Switch.Discovery.Arista.Ztp, etc down to one common graph.  I think that means that we could collapse the deploy switch startup config graphs down too (i.e. deploy-cisco-nexus3000-switch-startup-config-graph.js).  The trick there may be if we need different startup config scripts or different catalog scripts between different flavors of switches from one vendor.  In that case we may need to get model number too and pass it down.

pscharla: If I understand correctly, you are talking about doing something like this:

if (vendor === 'cisco') {
configuration.options.switchvendor = 'cisco';
} else if (vendor === 'arista') {
configuration.options.switchvendor = 'arista'';
} else {

Then we can collapse Graph.Switch.Discovery.Cisco.Poap, Graph.Switch.Discovery.Arista.Ztp, etc down to one common graph. I think that means that we could collapse the deploy switch startup config graphs down too (i.e. deploy-cisco-nexus3000-switch-startup-config-graph.js). The trick there may be if we need different startup config scripts or different catalog scripts between different flavors of switches from one vendor. In that case we may need to get model number too and pass it down.

benbp: @jlongever If I understand you correctly, passing the filename in directly is what my current proposal is doing (cisco-specific cataloging graph: https://github.com/benbp/on-taskgraph/blob/cisco-poap/lib/graphs/cisco-switch-discovery-graph.js#L18). Do you mean keeping a mapping somewhere of vendor type to a collection of script names?

@pscharla yeah, that's what I was talking about. With the deploy-cisco graph though, I put that in as an example, but I imagine we'll probably want to do less standardization on the dynamic SKU graphs, since some users may want to do different things (only deploy startup config, also deploy boot images, etc.).

jlongever: @benbp you're hard coding the filename in that definition,  I was just suggesting to make it an input option. So a common definition (as you suggested) that takes a filename option, instead of vendor options that strings the filename together. 

Also - would it make sense to move these graphs into SKU packs for each switch?

benbp: @jlongever The downside of passing in the filename as a vendor option happens when the cataloging workflow gets more complicated. If we have three cataloging tasks, now we have to pass in three filenames for each of the cataloging scripts, but the workflow still has to label those tasks separately as what they are, mark the catalog format and source, etc. I think decoupling those values in this case would just make things harder, because now they are defined in two places instead of one.

Yep, the example graphs are probably better placed in on-skupack. I'll make a PR for that, maybe as a cisco-nexus-example SKU?

data/profiles/taskrunner.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Download a python script from the tasks API
+    """
+    url = urllib2.urlparse.urljoin('http://<%=server%>:<%=port%>', downloadUrl)
+    print "Downloading script at {}".format(url)
+    script = urllib2.urlopen(url).read()
+    with open('script.py', 'w') as rackhd_script:
+        rackhd_script.write(script)
+
+while True:
```
yyscamper: I think we need a big `try`/`catch` block for the endless `while` loop, otherwise any exception of a particular command will cause the task runner failed and impact the following commands. I can see following cases may raise an exception:
1 - int(task_data["exit"])
2 - urllib2.Request
3 - urllib2.urlopen

benbp: @yyscamper in the Cisco case, we do wrap this in a try/catch, but from the outer script we downloaded via tftp: https://github.com/benbp/on-tftp/blob/cisco-poap/data/templates/cisco-poap.py#L16-L27

However, I'm realizing in the Arista case, we won't be wrapping this. That said, I think if the taskrunner execution fails at this level, my initial thinking is we *don't* want to recover from that, because staying in an infinite loop regardless of what happens is worse than letting the switch continue with its auto-install process (and likely fail that as well since the script will exit with a non-zero code). Perhaps a better approach is to wrap the urllib requests with retry handling, and add a smaller catch block for the `int(task_data["exit"])` statement. What do you think?

benbp:
@yyscamper I wrapped the rest of the commands in try/except blocks, and added some retry logic for urllib2.URLError cases. In other cases, we expect/want the script to exit.

heckj:
👍

PR #214, opened 2016-04-11
Modify template/profile rendering to add options that are normally

zyoung51:
exposed to task templates:

- api.server: set to full API server URI
- api.base: set to API base URI
- api.files: set to File API base URI
- api.nodes: set to Nodes API base URI
- context: set to the graph context
- task.nodeId: set to the nodeId of the current graph

### DIFF
--- a/lib/services/common-api-presenter.js
+++ b/lib/services/common-api-presenter.js
@@ -203,7 +203,7 @@ function CommonApiPresenterFactory(
          * the EJS renderer.
          * @param  {Integer} [status]  An optional HTTP status code.
          */
-        CommonApiPresenter.prototype.renderTemplate = function (name, options, status) {
+        CommonApiPresenter.prototype.renderTemplate = function (name, options, status, graphContext) {
             var self = this;
 
             options = options || {};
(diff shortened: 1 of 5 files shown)


lib/services/common-api-presenter.js
```diff
@@ -316,6 +299,36 @@ function CommonApiPresenterFactory(
             });
         };
 
+        CommonApiPresenter.prototype._buildContext = function(graphContext) {
+            var self = this;
+            var scope = self.res.locals.scope;
+            var config = configuration.getAll();
```
benbp: @zyoung51 It may make sense to cache this value as a class property on startup, as I found it can be quite an expensive call at runtime (blocking CPU for ~40ms, see my comment here: https://github.com/RackHD/on-tasks/blob/master/lib/task.js#L387-L390). Alone, it's not so bad, but when we have a bunch of pxe clients triggering property rendering operations all at the same time, it causes the system to be pretty laggy exactly when we don't want it to be.

lib/services/profiles-api-service.js
```diff
@@ -178,55 +178,37 @@ function profileApiServiceFactory(
...
-                else {
-                    return {
-                        profile: 'error.ipxe',
-                        options: {
-                            error: 'Unable to locate active workflow or there is no bootSettings.'
-                        }
-                    };
-                }
+                return Promise.resolve({
+                    profile: _.get(node, 'bootSettings.profile', 'error.ipxe'),
```
benbp: Ah, this is much nicer.

tannoa2:
+1

PR #224, opened 2016-04-13
Update tags-api-service and monorail-2.0.yaml 

uppalk1:
Updates to tags-api-service to use "generate-tag" key for options
Update monorail-2.0.yaml for POST/tags response code to be 201


tannoa2:
+1

uppalk1:
resolve https://github.com/RackHD/RackHD/issues/169

PR #225, opened 2016-04-13
Clean up error handling in Swagger renderer

brianparry:
Depends on https://github.com/RackHD/on-core/pull/112

### DIFF
--- a/lib/services/swagger-api-service.js
+++ b/lib/services/swagger-api-service.js
@@ -136,8 +136,8 @@ function swaggerFactory(
 
     function swaggerRenderer(req, res, viewName, next) {
         var status = req.swagger.options.success || 200;
-        return Promise.resolve().then(function() {
-            assert.ok(!res.headersSent);
+        return Promise.try(function() {
+            assert.ok(!res.headersSent, 'headers have already been sent');
             if (_.isEmpty(res.body) && req.swagger.options.send204OnEmpty) {
                 status = 204;
             }
@@ -163,15 +163,16 @@ function swaggerFactory(
             }
 
             return render.then(function(data) {
-                if (!data || typeof(data) !== 'string') {
-                    var err = new Errors.TemplateRenderError(
-                        viewName + ' could not be rendered.'
-                    );
-                    err.status = 500;
-                    throw(err);
+                // Rendered data should be JSON string.
+                try {
+                    JSON.parse(data);
+                } catch(err) {
+                    throw new Errors.ViewRenderError(err.message);
                 }
-                // Remove whitespace
                 return data.replace(/\s/g, '');
+            })
+            .catch(function(err) {
+                throw new Errors.ViewRenderError(err.message);
             });
         })
         .then(function(data) {
--- a/spec/lib/services/swagger-api-service-spec.js
+++ b/spec/lib/services/swagger-api-service-spec.js
@@ -502,6 +502,17 @@ describe('Services.Http.Swagger', function() {
             });
         });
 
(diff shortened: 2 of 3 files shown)


lib/services/swagger-api-service.js
```diff
@@ -163,15 +163,16 @@ function swaggerFactory(
...
+                // Rendered data should be JSON string.
+                try {
+                    JSON.parse(data);
+                } catch(err) {
+                    throw new Errors.ViewRenderError(err.message);
                 }
-                // Remove whitespace
                 return data.replace(/\s/g, '');
+            })
+            .catch(function(err) {
```
yyscamper: Seems this `catch` is not nessary

spec/lib/services/swagger-api-service-spec.js
```diff
@@ -502,6 +502,17 @@ describe('Services.Http.Swagger', function() {
             });
         });
 
+        it('should throw 500 on invalid template', function() {
+            res.body = { message: "foo" };
+            return renderer(req, res, 'foo', mockNext)
+            .then(function() {
+                expect(mockNext).to.be.calledWithMatch({status: 500});
+            },
+            function(err) {
```
yyscamper: I think this error callback is not necessary. If remove this error callback, the unit-test will still be marked as failed since the promise is rejected.

yyscamper:
Still :+1: with some minor suggestion.

lib/services/swagger-api-service.js
brianparry: Strictly speaking, you're right.  The reason I added it is to convert an AssertionError to a ViewRenderError so that the status property is set properly for the swagger render fitting.  Since both AssertionError and ViewRenderError result in 500 status, it ultimately doesn't matter but there could be problems if other unexpected errors are thrown or if we decide that we want to change the status code for ViewRenderError.

spec/lib/services/swagger-api-service-spec.js
brianparry: I think you're right.  I was afraid that the assertion wouldn't fire if the promise were rejected, but returning a rejected promise should cause the test to fail.

brianparry:
test this please

brianparry:
https://github.com/RackHD/on-core/pull/112 is merged, so test this please

spec/lib/services/swagger-api-service-spec.js
yyscamper: Hi @brianparry:

Below is a very simple test, I let the test case returns a rejected promise, it is surely caught by mocha, and other test cases are not impacted.

```javascript
it('return a rejected promise', function() {
    return Promise.reject();
});

it('return a fulfilled promise', function() {
    return Promise.resolve();
});
```

Below is the test output:
```
  test
    1) return a rejected promise
    ✓ return a fulfilled promise


  1 passing (7ms)
  1 failing

  1) test return a rejected promise:
     Error: Promise rejected with no or falsy reason
      at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
      at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:507:31)
      at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:581:18)
      at Promise._settlePromiseAtPostResolution (node_modules/bluebird/js/main/promise.js:245:10)
      at Async._drainQueue (node_modules/bluebird/js/main/async.js:128:12)
      at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
      at Async.drainQueues (node_modules/bluebird/js/main/async.js:15:14)
```

PR #227, opened 2016-04-13
Update centos-ks

keedya:
CentOS 7.1 needs efibootmgr to be installed in UEFI

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -47,7 +47,6 @@ reboot
 -fcoe*
 -b43-openfwwf
 kernel-firmware
--efibootmgr
 wget
 sudo
 perl


keedya:
@benbp @zyoung51 @RackHD/corecommitters 

I tested the ks script in legacy and uefi

heckj:
@keedya not quite getting this - PR description says it needs to be installed, but the diff removes the line from the kickstart. Is that intentional?

keedya:
@heckj  a leading dash is used to exclude this particular package

yyscamper:
:+1:

anhou:
:+1:

jlongever:
@keedya if the package needs to be installed shouldn't the '--' just be removed?

keedya:
@jlongever, must be installed as a dependency to other packages. I don't mind adding it , your call :smile:

jlongever:
Fine with me +1

PR #228, opened 2016-04-14
Tag bugfixes

zyoung51:
- Handle special characters in tag name
- Return 404 on GET/DELETE when tag resource does not exist
- Return 409 on POST when tag resource already exists

Resolves https://github.com/RackHD/RackHD/issues/179, resolves https://github.com/RackHD/RackHD/issues/180

### DIFF
--- a/lib/api/1.1/northbound/tags.js
+++ b/lib/api/1.1/northbound/tags.js
@@ -14,7 +14,8 @@ di.annotate(tagsRouterFactory, new di.Inject(
         'Http.Services.Api.Nodes',
         'Http.Services.RestApi',
         '_',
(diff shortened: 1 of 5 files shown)


yyscamper:
@zyoung51: could you please add a unit-testing to test the special characters (see https://github.com/RackHD/RackHD/issues/179) are well-handled?

:+1: after adding above proposed test case.

nortonluo:
If the tag exist in the system. Is there any way to modify/update the tag ? Or I have to delete the original tag first?

zyoung51:
@nortonluo: There are no PATCH/PUT verbs implemented for tags.

PR #233, opened 2016-04-18
Change body parser to allow post files upto 2MB

iceiilin:
Partial fix of ODR-620 about failure to update BIOS FW. In flash-quanta-bios-graph.js, after finishing FW update, the task POST the result (similar as stdout, >1M) to RackHD. In previous code, on-http refused this request, while the task is waiting for the response from nodes and never ends.

In current skupack implementation, there is no this problem since the output didn't return to RackHD. It is a fix for OnRack 1.2.0 who doesn't apply skupack and keeps using  flash-quanta-bios-graph.js

Removing the return value in on-tasks/ is also another solution, but enlarging the data size here can provide more flexibility for future usage.

### DIFF
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -196,8 +196,8 @@ function httpServiceFactory(
             skuService.static(req, res, next);
         });
 
-        // Parse request body. Limit set to 1MB
-        app.use(bodyParser.json({limit: '1mb'}));
+        // Parse request body. Limit set to 50MB
+        app.use(bodyParser.json({limit: '50mb'}));
 
         // Default Static Directory
         app.use(express.static('./static/http'));


iceiilin:
test this please

iceiilin:
test this please

iceiilin:
test this please

lib/services/http-service.js
```diff
@@ -197,7 +197,7 @@ function httpServiceFactory(
         });
 
         // Parse request body. Limit set to 1MB
-        app.use(bodyParser.json({limit: '1mb'}));
+        app.use(bodyParser.json({limit: '2mb'}));
```
stuart-stanley: please update comment to match.

iceiilin: done

stuart-stanley:
thank you! +1

benbp:
@iceiilin while you're making changes to this code, can we just bump it up to something like 10mb or even 50mb? We've come across this issue so many times throughout the history of the project (unpredictable catalog sizes have usually been the culprit) that I think it might make sense to just set a large maximum. I think it used to be set to 10mb a year ago, then got reverted back to 1mb for some reason.

@heckj do you remember whether we ended up deciding we wanted a smaller max size again for any reason, or was this just a regression?

heckj:
Regression - we set it to 10mb quite some time ago - default from express's `body-parser` plugin was 100K, which is WAY too small for this. I specifically recall if we set it to 1mb or 10mb, but no qualms about kicking it up higher for some of the more complex and complete payloads we're trying to capture here

anhou:
test this please

iceiilin:
@benbp @heckj Change the body parser limit to 50mb

anhou:
:+1:

Issue #211, opened 2016-04-08
Missing explicit IPMI channel numbers in set_bmc_credentials.sh


psalmia:
Some ipmitool commands in set_bmc_credentials.sh do not specify a channel number so the "currently used" channel is utilized. This causes problems in our hardware:

```
root@sut1-server2-pub:~# ipmitool user list
Get User Access command failed (channel 14, user 1): Invalid data field in request
root@sut1-server2-pub:~#
```
This causes the whole BMC credentials setting to fail.

The commands without a channel number are these:
```
cmdReturn=$(ipmitool user list)
...
cmdReturn1=$(ipmitool user summary)
```
The rest of the code suggests that channel 1 is to be used (judging from for example "ipmitool channel setaccess 1 $userNumber callin=on ipmi=on link=on privilege=4"), so the fix could be as simple as adding the channel number 1 to the commands listed above (e.g. "ipmitool user list 1" and "ipmitool user summary 1"). In our hw, using channel 1 works as expected.

Another option would be to figure out which channel works with something like this:
```
...
channel=''
function set_channel()
{
    for i in {1..15}; do
        ipmitool user list $i &>/dev/null
        status=$?
        if [ "$status" -eq "0" ] ; then
            channel=$i
            break
        fi
    done
}
set_channel
echo " Getting the user list from $channel"
cmdReturn=$(ipmitool user list $channel)
...
```

heckj:
fyi @keedya @RackHD/corecommitters

keedya:
Adding @uppalk1, @tannoa2

uppalk1:
Agree with that. I will fix it and update.

heckj:
@psalmia just out of curiosity, which hardware is this causing problems with? Is that something you can share?

amymullins:
For reference, the effort to cover resolution/update for this issue is being tracked in PT : https://www.pivotaltracker.com/projects/1492890/stories/117244649

psalmia:
The HW is Quanta STRATOS S210-X22RQ rackmount server.
On a related note, we have found a similar issue also in an HP ProLiant DL380p Generation8 server:
```
[root@S12 ~]# ipmitool user list
Get User Access command failed (channel 14, user 1): Parameter out of range
[root@S12 ~]# ipmitool user list 1
Get User Access command failed (channel 1, user 1): Parameter out of range
[root@S12 ~]# ipmitool user list 2
[...]

PR #234, opened 2016-04-18
Script changes to data/templates/set_bmc_credentials.sh

uppalk1:
* check the ipmi channel that's available
* use it to issue the ipmi commands
* Linked to gitHub issue :
Resolves https://github.com/RackHD/on-http/issues/211

### DIFF
--- a/data/templates/set_bmc_credentials.sh
+++ b/data/templates/set_bmc_credentials.sh
@@ -4,7 +4,7 @@ channel=''
 function set_channel()
 {
     for i in {1..15}; do
-        ipmitool -I lanplus -H 172.31.128.2 -U admin -P admin user list $i &>/dev/null
+        ipmitool user list $i &>/dev/null
         status=$?
         if [ "$status" -eq "0" ] ; then
             channel=$i
@@ -13,7 +13,11 @@ function set_channel()
     done
 }
 set_channel
-
+echo "channel number is" $channel
+if [ -z "${channel}" ]; then
+ echo "Channel number was not set correctly, exiting script"
+exit
+fi
 echo " Getting the user list"
 cmdReturn=$(ipmitool user list $channel)
 myarray=(${cmdReturn//$'\n'/ })
--- a/data/templates/set_bmc_credentials.sh
+++ b/data/templates/set_bmc_credentials.sh
@@ -1,9 +1,23 @@
 #!/usr/bin/env bash
 
+channel=''
+function set_channel()
+{
+    for i in {1..15}; do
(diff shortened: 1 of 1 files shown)


uppalk1:
@amymullins @jlongever @zyoung51 @tannoa2 @keedya

data/templates/set_bmc_credentials.sh
```diff
@@ -1,9 +1,23 @@
...
 
+channel=''
+function set_channel()
+{
+    for i in {1..15}; do
+        ipmitool -I lanplus -H 172.31.128.2 -U admin -P admin user list $i &>/dev/null
+        status=$?
+        if [ "$status" -eq "0" ] ; then
+            channel=$i
+            break
```
benbp: It's doubtful we'll hit this scenario, but it seems like this won't handle the case where $status is non-zero for all channel's 1-15?

benbp: Perhaps there should be a check after line 15 (`set_channel`) that errors out of the script if channel is still not set?

data/templates/set_bmc_credentials.sh
```diff
@@ -1,9 +1,23 @@
 #!/usr/bin/env bash
 
+channel=''
+function set_channel()
+{
+    for i in {1..15}; do
+        ipmitool -I lanplus -H 172.31.128.2 -U admin -P admin user list $i &>/dev/null
```
yyscamper: why hard-code the BMC IP?

yyscamper:
test this please

psalmia:
Hardcoding the address doesn't work and the command should be run locally anyway (i.e. just "ipmitool user list $i").

data/templates/set_bmc_credentials.sh
uppalk1: Of course hardcoding won't help :) I missed editing this line from my test script. Will fix it.

uppalk1:
test this please

data/templates/set_bmc_credentials.sh
```diff
@@ -1,9 +1,27 @@
...
+            channel=$i
+            break
+        fi
+    done
+}
+set_channel
+echo "channel number is" $channel
+if [ -z "${channel}" ]; then
+ echo "Channel number was not set correctly, exiting script"
+exit
```
jlongever: You may want to do a `exit 1` if this is an error case.  In case the caller wants to evaluate the script exit status $?.

fyi: http://tldp.org/LDP/abs/html/exitcodes.html

jlongever:
minor feedback, +1 either way.

uppalk1:
test this please !

PR #235, opened 2016-04-19
Add "sku" property for the node view

tannoa2:


tannoa2:
@geoff-reid @dalebremner @abildw @srinia6 @brianparry @amymullins

data/views/node.2.0.json
```diff
@@ -3,21 +3,28 @@
...
 <% } %>
-  "type": "<%=type%>",
-  "updatedAt": "<%=updatedAt%>"
-}
+"type": "<%=type%>",
+<% if (hasOwnProperty("sku")) { %>
+<% if (sku === null) { %>
+"sku" : null,
+<% } else  {%>
+"sku" : "<%=basepath %>/skus/<%=sku%>",
```
yyscamper: I'm not against this change, I just want to confirm that do you have get some agreement on changing the `sku` property to link instead of the raw sku identifier? Because this is different with API 1.1

anhou: This link style is not consistent with other parts of RackHD APIs like node's poller, this change need to be we discussed @RackHD/corecommitters

tannoa2:
@anhou and @yyscamper 
The link implementation was discussed with the Stryker team and that is how they wanted to be implemented.

data/views/node.2.0.json
yyscamper: yes, I prefer the link because it is more restful, but as @anhou mentioned, we need make it be consistent for all APIs (with same version).

yyscamper:
@tannoa2, could you please hold this PR until we discussed it in today's CC meeting?

lib/services/swagger-api-service.js
```diff
@@ -153,12 +153,14 @@ function swaggerFactory(
             res.set('Content-Type', 'application/json');
             if (_.isArray(res.body)) {
                 render = Promise.map(res.body, function(element) {
+                    element.basepath = req.swagger.operation.api.basePath;
```
brianparry: @tannoa2 Not a major concern, but you probably shouldn't be adding properties to the waterline record.  I'd suggest something like:
```javascript
var renderElement = _.merge({ basebase: req.swagger.operation.api.basePath }, element.toJSON());
```

data/views/node.2.0.json
brianparry: @yyscamper: Consistency with the 1.1 API has never been a goal of the 2.0 design.  By semver rules (and our own) we have the flexibility to develop new conventions if we feel they are appropriate.  Feedback on this is welcome but I don't want to operate in a mode where every change needs core committer pre-approval.  Our approach has been to implement the new API based on our best judgement and collect feedback on our work in Sprint and PR reviews.  IMO - comments like this should include a suggested alternative with an explanation of how that alternative improves things.
.
@anhou You are right that this is inconsistent with other parts of the 2.0 API where views have not been implemented yet.  There are stories in our backlog to implement views for other routes (workflows, pollers, etc).  In doing so, we'll make sure that we use a consistent set of conventions unless we have a reason to do otherwise.  For example, there are inconsistencies between /api/2.0/workflows and /api/2.0/nodes/{identifier}/workflows that we plan to change.

lib/services/swagger-api-service.js
tannoa2: Discussed with  @brianparry  again and it is fine to keep as it is

brianparry: Nevermind.  It looks like `element` has already been stripped of its instance methods, so this is fine as is.

brianparry:
:+1:

dalebremner:
+1

brianparry:
@yyscamper I'm going to go ahead and merge this so that @tannoa2 can make submit his functional test PR.  We can discuss further changes in the CC meeting tonight.

PR #236, opened 2016-04-19
[redfish] Add account service

zyoung51:
- Add /AccountService to Redfish:
    GET /AccountService
    GET /AccountService/Accounts
    POST /AccountService/Accounts
    GET /AccountService/Accounts/:name
    PATCH /AccountService/Accounts/:name
    DELETE /AccountService/Accounts/:name

- Update 2.0 /users routes
    GET /users
    POST /users
    GET /users/:name
    PATCH /users/:name
    DELETE /users/:name

- Fold acl-service into account-service
- Moved /Roles to /AccountService/Roles

@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @brianparry


lib/api/2.0/users.js
```diff
@@ -49,45 +48,48 @@ var addUser = controller({success: 201}, function(req, res) {
...
-        throw new Errors.UnauthorizedError('Unauthorized');
-    }
+    var name = req.swagger.params.name.value;
+    var userObj = _.pick(req.swagger.params.body.value, ['password', 'role']);
 
-    if(!req.isAuthenticated()) {
-        throw new Errors.UnauthorizedError('Unauthorized');
-    }
+    return Promise.try(function() {
+        if(!req.isAuthenticated || !req.isAuthenticated()) {
```
benbp: What if `isAuthenticated` is falsey but not a function? Won't we throw a `TypeError` here rather than the intended `UnauthorizedError`?

lib/api/redfish-1.0/account-service.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+});
+
+var getAccount = controller(function(req, res) {
+    var options = redfish.makeOptions(req, res);
+    var name = req.swagger.params.name.value;
+    return accountService.getUserByName(name)
+    .then(function(user) {
+        if(!user) {
+            throw new Error.NotFoundError();
```
benbp: Should be `Errors.NotFoundError()`

lib/api/redfish-1.0/account-service.js
```diff
@@ -0,0 +1,168 @@
...
+            // permit it, then let them add the first user.
+            return accountService.createUser({ 
+                username: payload.UserName,
+                password: payload.Password,
+                role: payload.RoleId
+            });
+        }
+
+        if( !_.find(users, function(user) {
+            return user.username === payload.UserName;
```
benbp: You can use the `_.matches` lodash shorthand here `_.find(users, { username: payload.UserName })`:

```
> var l = require('lodash')
> var user = {username: 'test'}
> var payload = {UserName: 'test'}
> l.find([user], {username: payload.UserName})
{ username: 'test' }
> l.find([user], {username: 'notfound'})
undefined
```

lib/api/redfish-1.0/account-service.js
```diff
@@ -0,0 +1,168 @@
...
+            }
+        }
+        throw new Errors.UnauthorizedError('Unauthorized');
+    })
+    .then(function() {
+        return accountService.getUserByName(payload.UserName);
+    })
+    .then(function(user) {
+        if(!user) {
+            throw new Error.NotFoundError();
```
benbp: Same as above, needs to be `Errors`

lib/api/redfish-1.0/account-service.js
```diff
@@ -0,0 +1,168 @@
...
+
+    return Promise.try(function() {
+        if(!req.isAuthenticated || !req.isAuthenticated()) {
+            throw new Errors.UnauthorizedError('Unauthorized');
+        }
+        return accountService.getUserByName(name);
+    })
+    .then(function(entry) {
+        if(!entry) {
+            throw new Error.NotFoundError();
```
benbp: Same as above.

benbp:
A couple of potential bugs I think I spotted, +1 after fixes.

lib/api/2.0/users.js
zyoung51: isAuthenticated is setup as a function in passport when a user is authenticated and logged in or absent otherwise.  I suppose something could tamper with that between the autn and here, but then passport wouldn't work and much more than this check would blow up.  How much less would you want me to trust passport?

jlongever:
test this please

lib/api/2.0/users.js
```diff
@@ -49,45 +48,48 @@ var addUser = controller({success: 201}, function(req, res) {
...
+            throw new Errors.UnauthorizedError('Unauthorized');
+        }
+        return accountService.getUserByName(name);
+    })
+    .then(function(entry) {
+        if(!entry) {
+            throw new Error.NotFoundError();
```
yyscamper: It's better to add some error message when throws an error

lib/services/account-api-service.js
```diff
@@ -0,0 +1,193 @@
...
+
+    function AccountApiService() {
+        this.acl = new nodeAcl(new nodeAcl.memoryBackend());  // jshint ignore:line
+    }
+
+    AccountApiService.prototype.start = function() {
+        var acl = this.acl;
+        var self = this;
+        var endpoints = configuration.get('httpEndpoints', []);
+        var authEnabled = _(endpoints).filter(function(endpoint) {
```
yyscamper: use `every` rather than `filter` can save the following `_.isEmpty`

lib/services/account-api-service.js
```diff
@@ -0,0 +1,193 @@
...
+                role: user.role,
+                privileges: self.aclMethod('_rolesParents', [user.role])
+            });
+        });
+    };
+
+    AccountApiService.prototype.createUser = function(user) {
+        var self = this;
+        return Promise.try(function() {
+            assert.string(user.username);
```
yyscamper: Seems need another two assertions for `password` and `role`

zyoung51: The swagger model validation has already validated these parameters.  The assert here is only of a pedantic nature and could be removed.

lib/services/account-api-service.js
```diff
@@ -0,0 +1,193 @@
...
+            return self.aclMethod('addUserRoles', entry.username, entry.role);
+        })
+        .then(function() {
+            return self.getUserByName(user.username);
+        });
+    };
+
+    AccountApiService.prototype.modifyUserByName = function(name, info) {
+        var self = this;
+        return Promise.try(function() {
```
yyscamper: I see some duplicated code for these APIs, following code are shared by most APIs, I suggest wrap the code to a function:
```
        return Promise.try(function() {
            assert.string(username);
            return waterline.localusers.findOne({username: username});
        })
        .then(function(user) {
            if(!user) {
                throw new Errors.NotFoundError();
            }
```

spec/lib/api/redfish-1.0/account-service-spec.js
```diff
@@ -0,0 +1,236 @@
...
+    var redfish;
+    var waterline;
+    var Promise;
+    var template;
+    var fs;
+    var accountService;
+
+    // Skip reading the entry from Mongo and return the entry directly
+    function redirectGet(entry) {
+        return fs.readFileAsync(__dirname + '/../../../../data/templates/' + entry, 'utf-8')
```
yyscamper: Using this constant about template path will be better: https://github.com/RackHD/on-core/blob/master/lib%2Fcommon%2Fconstants.js#L160

spec/lib/services/account-api-service-spec.js
```diff
@@ -0,0 +1,154 @@
...
+        waterline.localusers.find.resolves([user1, user2]);
+        return accountService.listUsers()
+        .then(function(users) {
+            expect(waterline.localusers.find).to.have.been.called;
+            expect(users.length).to.equal(2);
+            return Promise.map(users, function(user) {
+                expect(user.username).to.be.a('string');
+                expect(user.password).to.be.an('undefined');
+                expect(user.role).to.be.a('string');
+                expect(user.privileges).to.be.instanceof(Array);
```
yyscamper: It will be better to expect the user strictly equals to the info of `user1` and `user2` rather than only judging the returned types,

spec/lib/services/account-api-service-spec.js
```diff
@@ -0,0 +1,154 @@
...
+            expect(waterline.localusers.create).to.have.been.called;
+            expect(user.username).to.equal('admin');
+            expect(user.password).to.be.an('undefined');
+            expect(user.role).to.equal('Administrator');
+            expect(user.privileges).to.be.instanceof(Array);
+            expect(user.privileges[0]).to.equal('ConfigureUsers');
+        });
+    });
+
+    it('should recreate a user', function() {
```
yyscamper: The test case title doesn't match with the implementation, `should not recreate a duplicated user`?

spec/lib/services/account-api-service-spec.js
```diff
@@ -0,0 +1,154 @@
...
+        waterline.localusers.findOne
+            .onFirstCall().resolves(user1)
+            .onSecondCall().resolves(user1);
+        waterline.localusers.create.resolves(user1);
+        return accountService.createUser(user1)
+        .then(function() {
+            expect(false);
```
yyscamper: This line doesn't have any assertion logic, so it always passes. Need throw error here.

spec/lib/services/account-api-service-spec.js
```diff
@@ -0,0 +1,154 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+require('../../helper');
+
+describe("Account API Service", function() {
```
yyscamper: It lacks following two kinds of negative unit-test:
(1) If the `username` is not passed to some APIs (like getUser, deleteUser, modifyUser, .etc), these APIs should return a rejected promise.
(2) If the `username` is specified, but fail to find the user info from document, these API should also return a rejected promise.

lib/services/account-api-service.js
yyscamper: Can you think of is there any possibility that the Account-Api-Service will be used by other module except the `swagger` ? If so, the assertion logic is required, otherwise, remove the assertion is acceptable to me.

yyscamper:
Invite @cgx027 and @WangWinson to review this PR, since they contributed a lot for API 1.1 auth code.

lib/services/account-api-service.js
zyoung51: 2.0 and redfish both use swagger validation.  Your feedback below about adding negative unittests would make these assertions a requirement.  So while this code is unnecessary for production or functional use, it would strictly be required to satisfy your need to test for conditions that cannot occur.

yyscamper: If this code is unnecessary, then the unit-test requirement is still unnecessary.

lib/services/account-api-service.js
brianparry: `every` wouldn't work, but `some` would.  `every` would only initialize the service if _all_ endpoints have auth enabled.

yyscamper: Yes, you are right, `some` is correct.

PR #239, opened 2016-04-20
Add skus PUT to upsert sku definitions

zyoung51:
- Modify POST /skus to 409 on name conflict
- Create PUT /skus to upsert a sku definition
- Modify POST /skus/pack to 409 on name conflict

@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @brianparry 

Resolves https://github.com/RackHD/RackHD/issues/113

### DIFF
--- a/spec/lib/services/sku-pack-service-spec.js
+++ b/spec/lib/services/sku-pack-service-spec.js
@@ -32,7 +32,8 @@ describe("SKU Pack Service", function() {
         waterline.skus = {
             needByIdentifier: sinon.stub(),
             find: sinon.stub(),
-            create: sinon.stub()
+            create: sinon.stub(),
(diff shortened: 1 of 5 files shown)


spec/lib/services/sku-pack-service-spec.js
```diff
@@ -151,7 +152,7 @@ describe("SKU Pack Service", function() {
         });
     });
 
-    describe('hould post a sku', function(){
+    describe('should post a sku', function(){
```
yyscamper: How about to add a new test case to cover the new code? verify that the installPack will be rejected if the sku has already existed?

yyscamper:
:+1: it will be better to fix my comment.

PR #245, opened 2016-04-21
Added API 2.0 Files Routes

dalebremner:
@RackHD/corecommitters @abildw @zyoung51 

Contributions made by Billy Abildgaard, Brian Parry, Anitha Srinivasan, Dale Bremner


brianparry:
test this please

lib/api/2.0/files.js
```diff
@@ -0,0 +1,120 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var fileService = injector.get('fileService');
+
+/**
```
zyoung51: Same comment for this and all the apidoc comments below: The values that you put here are different than the values you put in the swagger file.  If we are going to document the same thing twice then both places should contain the same documentation.  As an example, here you state the apiName is files-get but in swagger the API is named filesGet.  Here you say 'get the specified file' and do not document ':fileid' but in swagger you say 'get file based on uuid or file name' and do document 'fileidentifier'

dalebremner: @zyoung51 As we discussed, removed the apidoc comments from the controller.

PR #247, opened 2016-04-21
[redfish] Add Redfish EventService resource

jlongever:
- Added POST /EventService/Subscription route to subscribe to events
- Added GET /EventService to get the root EventService resource
- Added GET /EventService/Subscription/:id route to get subscription resource collection or specific resource
- Added DELETE /EventService/Subscription/:id to delete an existing subscription resource
- Added POST /EventService/Actions/EventService.SubmitTestEvent
- Added poller alert listener, publish EventService types StatusChange (discrete) and Alert (threshold)
- Added unit-tests

@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2 @geoff-reid @keedya

### DIFF
--- a/data/templates/redfish.1.0.0.eventdestination.json
+++ b/data/templates/redfish.1.0.0.eventdestination.json
@@ -1,0 +1,11 @@
+{
+    "@odata.context": "<%= basepath %>/$metadata#EventService/Members/Subscriptions/Members/$entity",
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#EventService.1.0.0.EventDestination",
+    "Id": "<%= subscription.Id %>",
+    "Name": "Event Subscription Destination",
+    "Destination": "<%= subscription.Destination %>",
+    "EventTypes": <%- JSON.stringify(subscription.EventTypes) %>,
+    "Context": "<%= subscription.Context %>",
+    "Protocol": "<%= subscription.Protocol %>"
+}
--- a/data/templates/redfish.1.0.0.eventdestinationcollection.json
+++ b/data/templates/redfish.1.0.0.eventdestinationcollection.json
@@ -1,0 +1,16 @@
+{
+    "@odata.context" : "<%= basepath %>/$metadata#EventService/Members/Events/$entity",
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#EventDestinationCollection.EventDestinationCollection",
+    "Name": "Event Subscription Collection",
+    "Members@odata.count": <%= subscriptions.length %>,
+    "Members": [
+        <% subscriptions.forEach(function(subscription, i, arr) { %>
+            {
+                "@odata.id": "<%= basepath %>/EventService/Subscriptions/<%=subscription.Id%>"
+            }
+            <%= ( arr.length > 0 && i < arr.length-1 ) ? ',': '' %>
+        <% }); %>
+    ]
+    
+}
--- a/lib/api/redfish-1.0/event-service.js
+++ b/lib/api/redfish-1.0/event-service.js
@@ -1,0 +1,266 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+var urlParse = require('url-parse');
+var injector = require('../../../index.js').injector;
+var redfish = injector.get('Http.Api.Services.Redfish');
+var schemaApi = injector.get('Http.Api.Services.Schema');
+var Promise = injector.get('Promise'); // jshint ignore:line
+var _ = injector.get('_');  // jshint ignore:line
+var Constants = injector.get('Constants');
+var RedfishTool = injector.get('JobUtils.RedfishTool');
+var controller = injector.get('Http.Services.Swagger').controller;
+var Errors = injector.get('Errors');
+var messenger = injector.get('Services.Messenger');
+var Logger = injector.get('Logger');
+var uuid = injector.get('uuid');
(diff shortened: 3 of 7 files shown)


data/templates/redfish.1.0.0.eventdestination.json
```diff
@@ -0,0 +1,11 @@
+{
+    "@odata.context": "<%= basepath %>/$metadata#EventService/Members/Subscriptions/Members/$entity",
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#EventService.1.0.0.EventDestination",
+    "Id": "<%= subscription.Id %>",
+    "Name": "Event Subscription Destination",
+    "Destination": "<%= subscription.Destination %>",
+    "EventTypes": <%- JSON.stringify(subscription.EventTypes) %>,
```
benbp: Have we marked `<%-` as a valid delimiter (or is it another default?), or is this just a typo from `<%=` ?

data/templates/redfish.1.0.0.eventdestinationcollection.json
```diff
@@ -0,0 +1,16 @@
...
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#EventDestinationCollection.EventDestinationCollection",
+    "Name": "Event Subscription Collection",
+    "Members@odata.count": <%= subscriptions.length %>,
+    "Members": [
+        <% subscriptions.forEach(function(subscription, i, arr) { %>
+            {
+                "@odata.id": "<%= basepath %>/EventService/Subscriptions/<%=subscription.Id%>"
+            }
+            <%= ( arr.length > 0 && i < arr.length-1 ) ? ',': '' %>
```
benbp: Haha, I like this.

data/templates/redfish.1.0.0.eventdestination.json
jlongever: It's a valid delimiter for rendering variables.

zyoung51: <%- will not do HTML escaping.  <%= will do escaping.

lib/api/redfish-1.0/event-service.js
```diff
@@ -0,0 +1,266 @@
...
+var sendClient = function(client, retry, data) {
+    return client.clientRequest(client.settings.root, 'POST', data)
+    .catch(function(error) {
+        if(0 !== retry) {
+            logger.warning('Send Error', {
+                error:error,
+                settings:client.settings,
+                retry:retry
+            });
+            sendClient(client, retry - 1);
```
benbp: Need to return this as well otherwise the original caller won't wait on any recursive promises.

lib/api/redfish-1.0/event-service.js
```diff
@@ -0,0 +1,266 @@
...
+var messenger = injector.get('Services.Messenger');
+var Logger = injector.get('Logger');
+var uuid = injector.get('uuid');
+var timeoutInSeconds = 10;
+var retryAttempts = 3;
+var eventSubscriptions = {};
+var subscription;
+var logger = Logger.initialize('event-service');
+
+var eventListener = function(callback) {
```
benbp: You could also just represent this as `function eventListener` rather than `var eventListener = function`. Same for other functions below. If you do it that way, then the location of the definition in the file doesn't matter for having access to the function name in this scope.

data/templates/redfish.1.0.0.eventdestination.json
benbp: Oh, that's handy to know.

lib/api/redfish-1.0/event-service.js
```diff
@@ -0,0 +1,266 @@
...
+            var err = new Errors.BaseError('EventDestination Exists');
+            err.status = 409;
+            throw err;
+        }
+        event.Id = uuid('v4');
+        _.set(eventSubscriptions, event.Id, event);
+        return event;
+    })
+    .then(function(output) {
+        eventListener(eventCallback);
```
benbp: So `createSubscription` can get called when someone wants to start retrieval a collection of events? A potential issue I'm seeing is that we don't have any way to clean up these AMQP subscriptions, as `messenger.subscribe` creates a persistent object/subscription to AMQP until `.dispose()` is called, rather than a one-time use subscription.

benbp: Also, I think there are some gaps in error handling, since we don't return any of the promises triggered off the `eventListener(eventCallback)` call, and additionally the promises within `eventListener` aren't really collected in any way, or have `.catch` calls on them. Just worried about potential hard to debug/silently caught errors in the future.

jlongever: `.dispose()` is called when there are no client subscriptions.  See https://github.com/RackHD/on-http/pull/247/files#diff-4c22a32287806e96d657bda177f19b1dR245

jlongever: Also, only a single subscription is made on the first subscription request.

jlongever: @benbp I tested the typical error scenarios from the eventListener such as sendClient() and schemaApi.validate() exceptions.  Most of the code there is synchronous and not much to handle in terms of errors.   Outside of failing to send the event record to the client or error validation what should we be handling?

benbp:
👍  with a follow-on to add more logging/handling around errors.

PR #249, opened 2016-04-22
Fix ubuntu templates issues

WangWinson:
- Fix error when rootSshKey not defined in post install script
- Add grub-installer/bootdev section in preseed to prevent installer interrupt question such as below

   > [ Install the GRUB boot loader on a hard disk ]

   > You need to make the newly installed system bootable, by installing the GRUB boot loader on a       bootable device. The usual way to do this is to install GRUB on the master boot record of your first hard drive. If you prefer, you can install GRUB elsewhere on the drive, or to another drive, or even to a floppy.

   > The device should be specified as a device in /dev. Below are some examples:

   > "/dev/sda" will install GRUB to the master boot record of your first hard drive;
   > "/dev/sda2" will use the second partition of your first hard drive;

@RackHD/corecommitters @pengz1 @sunnyqianzhang

### DIFF
--- a/data/templates/install-ubuntu/post-install-ubuntu.sh
+++ b/data/templates/install-ubuntu/post-install-ubuntu.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # create SSH key for root
-<% if (undefined !== rootSshKey) { -%>
+<% if ('undefined' !== typeof rootSshKey && null !== rootSshKey) { -%>
 mkdir /root/.ssh
 echo <%=rootSshKey%> > /root/.ssh/authorized_keys
 chown -R root:root /root/.ssh
--- a/data/templates/install-ubuntu/ubuntu-preseed
+++ b/data/templates/install-ubuntu/ubuntu-preseed
@@ -144,6 +144,8 @@ d-i grub-installer/only_debian boolean true
 # OS, which is less safe as it might not be able to boot that other OS.
 d-i grub-installer/with_other_os boolean true
 
+d-i grub-installer/bootdev string <%=installDisk%>
+
 # Avoid that last message about the install being complete.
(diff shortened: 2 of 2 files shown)


pengz1:
+1

anhou:
test this please

yyscamper:
:+1:

data/templates/install-ubuntu/post-install-ubuntu.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # create SSH key for root
-<% if (undefined !== rootSshKey) { -%>
+<% if ('undefined' !== typeof rootSshKey && null !== rootSshKey) { -%>
```
benbp: You can shorten this if you want to either:

`hasOwnProperty("rootSshKey")` or `'undefined' != typeof rootSshKey` (!= non-strict comparison matches both undefined and null).

But this is also more explicit so it works as well :)

benbp:
👍

heckj:
:+1:

heckj:
test this please

data/templates/install-ubuntu/post-install-ubuntu.sh
WangWinson: Thanks @benbp. I know this piece looks very clumsy. 
Using `!=` looks good, but it breaking our convention, doesn't it ? 
I think you mean `if (this.hasOwnProperty("rootSshKey"))`, right ? This should work but I still need to check `null` value.
The clear way maybe using `_.isEmpty(this.rootSshKey)`, but unfortunately, we can not using lodash directly in ejs , right ?

WangWinson:
test this please

WangWinson:
test this please

PR #253, opened 2016-04-27
remove setup_iso.py as it has been moved to on-tools

yyscamper:
I just see the setup_iso.py has been moved to on-tools repo (https://github.com/RackHD/on-tools/blob/master/scripts%2Fsetup_iso.py), so it doesn't need in on-http repository, meanwhile the on-tools is the better place to host such utility script  than on-http.

@RackHD/corecommitters @zyoung51 @pengz1 @WangWinson @iceiilin


anhou:
:+1:

zyoung51:
FYI @DavidjohnBlodgett as you mentioned this file in Slack: 
`sudo python ~/src/on-http/data/templates/setup_iso.py /tmp/CentOS-7-x86_64*.iso /var/mirrors --link=/home/vagrant/src`

Can you update your reference to on-tools?

DavidjohnBlodgett:
@zyoung51 no problem, in fact I've already updated docs on my branch... however I haven't pushed the changes because I hit a lot of other problems with those instructions that are forcing me to rework them.

PR #255, opened 2016-04-27
Refactor config.json into sku model and cleanup error messages

zyoung51:
- Change InternalError errors to BadRequest when the issue arises due to
  bad data in the skupack file that is posted.
- Move the config.json properties into the model

### DIFF
--- a/lib/services/sku-pack-service.js
+++ b/lib/services/sku-pack-service.js
@@ -52,6 +52,14 @@ function skuPackServiceFactory(
     var logger = Logger.initialize(skuPackServiceFactory);
     var fs = Promise.promisifyAll(nodeFs);
     var rimrafAsync = Promise.promisify(rimraf);
+    var confProperties = [
+        'httpStaticRoot',
+        'httpTemplateRoot',
+        'httpProfileRoot',
+        'workflowRoot',
+        'taskRoot'
+    ];
+    var validConfProperties = confProperties.concat('skuConfig', 'version', 'description');
 
     function SkuPackService() {
         this.loader = new FileLoader();
@@ -126,7 +134,6 @@ function skuPackServiceFactory(
         var self = this;
         var name = uuid('v4');
         var tmpDir = tmp();
-        var skuName = '';
 
         return Promise.try(function() {
             assert.ok(req.headers['content-type']);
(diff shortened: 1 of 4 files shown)


zyoung51:
test this please

lib/services/sku-pack-service.js
```diff
@@ -444,28 +462,30 @@ function skuPackServiceFactory(
      * @param  {String}     fromRoot The path to the configuration file
      * @return {Promise}
      */
-    SkuPackService.prototype.validatePack = function(contents, fromRoot) {
-        try {
+    SkuPackService.prototype.validatePack = function(contents, fromRoot, options) {
+        options = options || {};
+
+        return Promise.try(function() {
             var conf = JSON.parse(contents);
```
yyscamper: Could you please add a try/catch for JSON.parse? so the invalid JSON format error can be explicitly threw out to user,  I had said this was an enhance opportunity in this issue https://github.com/RackHD/RackHD/issues/190

zyoung51: Promise.try() converts synchronous throws into rejects to be caught by .catch().  The error message, along with the notice about the invalid json, occurs at line 487.

lib/services/sku-pack-service.js
```diff
@@ -477,63 +497,69 @@ function skuPackServiceFactory(
...
-                        throw new Error('rules or name is missing');
+        .then(function(fileContents) {
+            contents = fileContents;
+            return self.validatePack(fileContents, fromRoot, {
+                validateSku: !_.isUndefined(skuid)
+            });
+        })
+        .then(function() {
+            if(skuid === undefined)  {
+                var conf = JSON.parse(contents);
```
yyscamper: Also to add a try/catch for this JSON.parse

zyoung51: Pre-validated in validatePack above.

lib/services/sku-pack-service.js
```diff
@@ -477,63 +497,69 @@ function skuPackServiceFactory(
...
-                return [skuID, fs.readdirAsync(fromRoot), fs.mkdirAsync(dst) ];
-            })
-            .spread(function(sku, entries) {
-                return Promise.map(entries, function(entry) {
-                    var src = fromRoot + '/' + entry;
-                    if( entry === 'config.json' ) {
-                        return fs.renameAsync(src, dst + '.json');
-                    } else {
-                        return fs.renameAsync(src, dst + '/' + entry);
+                    return waterline.skus.create(_.pick(conf, ['name', 'rules', 'discoveryGraphName', 'discoveryGraphOptions'] ));
```
yyscamper: Why only pick these keys? From your SKU model definition in on-core (https://github.com/RackHD/on-core/pull/123/files) , other keys such as "workflowRoot", "taskRoot" should also be inserted.

So either PR need a fix.

zyoung51: Line 542...

lib/services/sku-pack-service.js
```diff
@@ -477,63 +497,69 @@ function skuPackServiceFactory(
...
-                    if(!conf.hasOwnProperty('rules') || !conf.hasOwnProperty('name')) {
-                        throw new Error('rules or name is missing');
+        .then(function(fileContents) {
+            contents = fileContents;
+            return self.validatePack(fileContents, fromRoot, {
+                validateSku: !_.isUndefined(skuid)
+            });
+        })
+        .then(function() {
+            if(skuid === undefined)  {
```
yyscamper: `undefined` seems not enough, other values like `null` need be judged as well.

zyoung51: The controller code validates that the skuid exists if it is defined.  It is already validated.

lib/services/sku-pack-service.js
yyscamper: Oh, yes, I see it. Thanks.

lib/services/sku-pack-service.js
yyscamper: Why split to two db operations? Any reason to design like this?

zyoung51: In the case where we are not creating the sku (skuid is defined), we are only modifying the sku pack directories and not replacing the sku rules/names

jlongever:
test this please

lib/services/sku-pack-service.js
yyscamper: Sorry, still confused, not creating the SKU? but this is `waterline.skus.create`, it is sure to create a new SKU.

zyoung51: It creates a new sku if this was called with the sku undefined as in POST /skus/pack.  It does not create or update the sku related model settings if it was called with the sku defined when replacing the pack binaries as in PUT /skus/:id/pack.  In both cases the .update is called to create/replace the skupack related model settings.

PR #256, opened 2016-04-28
API 2.0 Schemas Routes

dalebremner:
@RackHD/corecommitters @zyoung51 @srinia6 @abildw @geoff-reid


abildw:
+1 

@dalebremner should there be error codes expected for responses in the yaml?

Also curious why you named the controller "schemas2"

PR #257, opened 2016-04-28
Adding arista catalog, config scripts

pscharla:
This PR contains startup script for Arista as well as catalog scripts.  The Arista CLI does not support return of JSON format for most of the commands so needed to do parsing of the data.

@RackHD/corecommitters @benbp @heckj @VulpesArtificem @johren

### DIFF
--- a/data/templates/arista-catalog-config.py
+++ b/data/templates/arista-catalog-config.py
@@ -1,0 +1,19 @@
+#!/usr/bin/python
+
+def main():
+    import subprocess
+    data ={}
+
+    try:
+	cmd = "Cli -p2 -c 'show startup-config'"
+        data['startup-config'] = subprocess.check_output(cmd, shell=True)
+    except:
+        pass
+
+    try:
+	cmd = "Cli -p2 -c 'show running-config'"
+        data['running-config'] = subprocess.check_output(cmd, shell=True)
+    except:
+        pass
+
+    return data
--- a/data/templates/arista-catalog-config.py
+++ b/data/templates/arista-catalog-config.py
@@ -5,13 +5,13 @@ def main():
     data ={}
 
     try:
-	cmd = "Cli -p2 -c 'show startup-config'"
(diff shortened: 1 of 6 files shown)


benbp:
👍

data/templates/arista-catalog-config.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/python
+
+def main():
+    import subprocess
+    data ={}
+
+    try:
+	cmd = "Cli -p2 -c 'show startup-config'"
+        data['startup-config'] = subprocess.check_output(cmd, shell=True)
```
brianparry: @pscharla Looks like this will raise an `IndentationError`.

data/templates/arista-catalog-config.py
```diff
@@ -0,0 +1,19 @@
...
+    except:
+        pass
+
+    try:
+	cmd = "Cli -p2 -c 'show running-config'"
+        data['running-config'] = subprocess.check_output(cmd, shell=True)
```
brianparry: @pscharla Another indentation error.

data/templates/arista-catalog-version.py
```diff
@@ -0,0 +1,64 @@
...
+    for line in versionString.splitlines():
+        emptyValue = False
+        # Add quotes around key
+        line = '"' + line
+        line = line.replace(':', '":')
+
+        # Add key for the Arista switch model
+        if (line.find(':') == -1):
+            emptyValue = True
+	    line = '"Hardwaremodel":' + line + '"'
```
brianparry: Indentation error

brianparry:
There are a few indentation errors which I think will result in `IndentationError` exceptions.  :+1: Once those are resolved.

data/templates/arista-catalog-config.py
pscharla: @brianparry Thanks for catching that.  Had to fix my .vimrc file.

PR #258, opened 2016-04-28
Added lookups route to 2.0 API

geoff-reid:
@brianparry @dalebremner @abildw @srinia6


static/schemas/2.0/Lookups.2.0.json
```diff
@@ -0,0 +1,25 @@
...
+        "Lookup": {
+            "description": "A lookup record",
+            "type": "object",
+            "properties": {
+                "node": {
+                    "type": "string"
+                },
+                "ipAddress": {
+                    "type": "string",
+                    "format": "ipv4"
```
zyoung51: The ipv4 format validator needs to be added to tv4.  It can be added to tv4 by calling addFormat: "tv4.addFormat(format, validationFunction)".

lib/api/2.0/lookups.js
```diff
@@ -0,0 +1,101 @@
...
+ * @apiParam {String} id lookup id
+ * @apiSuccess {json} lookup the lookup that has the <code>id</code> that was patched
+ * @apiError NotFound The lookup with the <code>id</code> was not found.
+ * @apiErrorExample {json} Error-Response:
+ *     HTTP/1.1 404 Not Found
+ *     {
+ *       "error": "Not Found"
+ *      }
+ */
+var lookupsPatchById = controller(function(req) {
```
brianparry: Because the schema requires the macAddress property, patch will only work if macAddress is one of the properties provided with the request.  For the pollers route, I added a sub-schema with no required properties for use with patch: https://github.com/brianparry/on-http/blob/feature/pollers-refactor/static/schemas/2.0/poller.2.0.json#L27

lib/api/2.0/lookups.js
```diff
@@ -0,0 +1,101 @@
...
+ * @apiName lookups-post
+ * @apiParam {String} macAddress MAC Address for the Lookup Record
+ * @apiParam {String} ipAddress IP Address associated with the MAC Address
+ * @apiParam {String} node Node associated with the Lookup Record
+ * @apiGroup lookups
+ * @apiSuccess {json} lookup the lookup that was created
+ */
+
+var lookupsPost = controller({success: 201}, function (req) {
+    return waterline.lookups.create(req.swagger.params.body.value);
```
yyscamper: How about if the lookups pool already have an identical one?

zyoung51: Mongo will reject it.  We setup the unique index in on-core to prevent the duplicates from being created.

spec/lib/api/2.0/lookups-spec.js
```diff
@@ -0,0 +1,130 @@
...
+
+'use strict';
+
+describe('Http.Api.Lookup 2.0', function () {
+    var waterline, stub;
+
+    var data = [
+        {
+            id: '123',
+            createdAt: 'createdAt',
```
yyscamper: Even this is unit-test and this is correct to meet the test logic, but I still suggest using a real datetime string. Because the unit-test is a kind of code document, developer may lookup the unit-test to find how to use the API.

spec/lib/api/2.0/lookups-spec.js
```diff
@@ -0,0 +1,130 @@
...
+        }
+    });
+
+    after('stop HTTP server', function () {
+        return helper.stopServer();
+    });
+
+    describe('/api/2.0/lookups', function () {
+        describe('GET', function () {
+            it('should should call waterline.lookups.findByTerm', function() {
```
yyscamper: duplicated word "should"

spec/lib/api/2.0/lookups-spec.js
```diff
@@ -0,0 +1,130 @@
...
+                stub = sinon.stub(waterline.lookups, 'findByTerm').resolves(data);
+
+                return helper.request().get('/api/2.0/lookups')
+                    .expect('Content-Type', /^application\/json/)
+                    .expect(200)
+                    .expect(function () {
+                        expect(stub).to.have.been.calledWith(undefined);
```
yyscamper: Miss the assertion for the response data

lib/api/2.0/lookups.js
geoff-reid: I have updated the schema so PATCH will no longer require a macAddress.

zyoung51:
Thank you for adding the format validator. Some cleanup to remove console.log's and then looks good to me, +1

package.json
```diff
@@ -26,30 +26,31 @@
     "express": "^4.10.7",
     "glob": "^4.3.2",
     "http-proxy-middleware": "~0.11.0",
+    "ip": "~1.1.2",
```
brianparry: @geoff-reid It looks like there are some changes in here that aren't part of your patch.

geoff-reid: this was added by Joel, see commit:  https://github.com/RackHD/on-http/commit/138f6f265d836020af4020feb70f3dab83efea3a, I just put it in alphabetical order.

geoff-reid:
fixed hound errors and removed debug console.log messages.

brianparry:
expected failure, test this please

geoff-reid:
rebases and squashed commits.

PR #265, opened 2016-04-29
Bump mocha timeout to 10000 to mitigate potential jenkins issues

benbp:
Proposing a potential way to address what could be jenkins timeout issues when loading modules in mocha "before" blocks.

@RackHD/corecommitters @tldavies @davidzuhn

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "doc": "./node_modules/.bin/jsdoc lib/api/1.1/northbound -r -d docs",
-    "test": "./node_modules/.bin/mocha $(find spec -name '*-spec.js') -R spec --require spec/helper.js",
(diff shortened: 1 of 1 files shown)


brianparry:
@benbp One of the issues here is that each test calls startServer in a before block and endServer in an after.  We need to investigate why, but swagger takes several seconds to initialize which means that tests require long timeouts.  I think this is a reasonable workaround, so :+1: but we should still investigate if there is a way to reduce swagger startup time or reduce its impact on test time.

tldavies:
👎

benbp:
@brianparry yeah I was wondering if perhaps there was some way to cache whatever loading swagger does so we could avoid the test lag.

Closing this for now since @tldavies is suggesting the root problem with exceeding timeouts (rather than just taking a while) could be in the provisioning of the build environment.

Issue #261, opened 2016-04-29
The SKU pack post fails in vagrant setup


tannoa2:
The vagrant setup runs off an nfs mount which is causing an issue in the skupack handler when trying to create a /tmp directory.

tannoa2:
@amymullins @zyoung51

PR #268, opened 2016-05-02
Change skupack to use fs-extra's move()

zyoung51:
- This resolves an issue where using node's rename() across volumes would cause
  EXDEV errors.

Resolves https://github.com/RackHD/on-http/issues/261


tannoa2:
+1

PR #270, opened 2016-05-04
Fix for Issue 172

dalebremner:
@RackHD/corecommitters @BillyAbildgaard @geoff-reid @srinia6 @zyoung51 

resolves https://github.com/RackHD/RackHD/issues/172


BillyAbildgaard:
+1

lib/services/workflow-api-service.js
```diff
@@ -143,11 +143,13 @@ function workflowApiServiceFactory(
...
-                // Cancelled graph ID doesn't match the requested ID.
-                throw new Errors.TaskCancellationError('Failed to cancel ' + graphId);
-            }
-            return waterline.graphobjects.needByIdentifier(graphId);
+            return waterline.graphobjects.needByIdentifier(graphId)
+            .then (function(workflow) {
+                if (workflow.active()) {
+                    throw new Errors.TaskCancellationError('Failed to cancel ' + graphId);
+                }
+                return waterline.graphobjects.needByIdentifier(graphId);
```
benbp: Graph cancellation is not instantaneous, it is an asynchronous action that gets added to the task scheduler queue. This seems likely to produce false negatives.

benbp: This API call isn't really meant to be deterministic either, but rather something that gets kicked off into the background. An upstream caller, if they wanted to verify cancellation, would probably be better off polling the workflow object by ID after issuing the cancellation action to make sure the status changes within a certain time window.

benbp:
I know this is likely a high priority issue, but I don't think this is the right approach to solving it given the behavior of the underlying system (see above comments). -1

zyoung51:
@benbp Should this be returning a 202 and maybe a Location header where the caller can query the real final state when it is reached?

jlongever:
Agree with @benbp,  -1.  Probably want to do what @zyoung51 suggested too.

lib/services/workflow-api-service.js
jlongever: I think similar to the random 1.1 functional test errors,  for some reason the underlining cancelledGraphId doesn't always get set by cancelTaskGraph, it ends up undefined.

jlongever: Some more relevant info: https://github.com/RackHD/RackHD/issues/151

dalebremner:
@benbp @jlongever @zyoung51 @brianparry It seems as though graph cancellation doesn't lend itself well to functional testing. Do you think we should eliminate the functional test for the short term?

jlongever:
@dalebremner I think the test is exposing a legitimate problem that shouldn't be ignored.

yyscamper:
Agree @zyoung51

lib/services/workflow-api-service.js
brianparry: @dalebremner You don't need another query here; you can just return `workflow`.  Also TaskCancellationError will default to a 400, it should actually be a 500.

brianparry:
@zyoung51 @benbp When I originally implemented this function, I considered returning 202 or implementing an additional message to indicate when cancel had completed.  However, I discovered that cancel uses `messenger.request` which returns a promise that resolves only after receiving a response.  So, while graph cancellation is asynchronous, the operation should have completed when the promise returned from `taskGraphProtocol.cancelTaskGraph` is resolved.

The root cause of the workflow cancel errors we've been seeing in functional tests is that the graph is failing before we can cancel it.  The intent of this change is to return an error only in cases where we attempted to cancel a running graph and found the graph to be still running after the cancel operation completed.  Effectively we're defining cancel sementics as "make graph inactive".  So in the cases where a graph completes or fails prior to cancellation, we'd return 200.

jlongever:
@brianparry I don't think it's a good idea to force the client to wait on the promise to be resolved.  Ideally the client should get the graphId to poll the cancel status on.  Seems like running a workflow that randomly fails (discovery graph in this case) is exposing unexpected state values,  since we wait for the state to transition to pending/running?

brianparry:
@zyoung51 @jlongever @benbp That sounds reasonable.  The more I think about it, the more I like the idea of returning 202; it think it will eliminate a few edge cases and simplify the code.  This change will also impact the DELETE method because it currently attempts to cancel an active workflow before deleting it.  We probably just want to reject the DELETE with a 4xx if the workflow is active and require that the client either wait for it to finish or explicitly cancel it.

I think the functional test should also change to check the Location header and validate that the workflow actually get cancelled.

benbp:
202 approach sounds good to me as well. @brianparry the race condition where a graph fails during the cancel request was something I tried to anticipate with the code after it initially was brought up in review. I think the intended behavior is to return a 400 or something in those cases. If this is also causing functional test failures then it seems to me like part of the problem is that the workflow under test is failing when it shouldn't be? Since that's not the state that we are trying to test?

lib/api/2.0/nodes.js
```diff
@@ -73,7 +73,7 @@ var nodesGetActiveWorkflowById = controller({ send204OnEmpty: true }, function(r
     return nodes.getActiveNodeWorkflowById(req.swagger.params.identifier.value);
 });
 
-var nodesDelActiveWorkflowById = controller({success: 204}, function(req) {
+var nodesDelActiveWorkflowById = controller({success: 202}, function(req) {
     return nodes.delActiveWorkflowById(req.swagger.params.identifier.value);
 });
```
zyoung51: Might be even cooler to:
return nodes.delActiveWorkflowById(req.swagger.params.identifier.value)
.then(function(graph) {
    res.setHeader('Location', '/api/2.0/workflows/' + graph);  
});

(assuming that is the right uri to get the ultimate status of the graph?)

dalebremner: Thanks @zyoung51 ... I added the Location header...

zyoung51:
TravisCI had a timeout error in the lookups 2.0 api, restarted/retrying it.

srinia6:
+1

PR #272, opened 2016-05-05
Remove decodeURIComponent and add sway 1.0.0 fix

zyoung51:
The previous decodeURI, which was replaced by decodeURIComponent, was
masking an issue in sway.  This change removes the decodeURIComponent,
which is already being done by express and (attempted by) sway and
changes the 2.0 code to use the sway 1.0.0 fix.

### DIFF
--- a/lib/api/2.0/tags.js
+++ b/lib/api/2.0/tags.js
@@ -8,6 +8,19 @@ var nodes = injector.get('Http.Services.Api.Nodes');
 var tags = injector.get('Http.Services.Api.Tags');
 var _ = injector.get('_'); // jshint ignore:line
 var Errors = injector.get('Errors');
+var parseUrl = require('url').parse;
+
+function getTagName(req) {
+    //TODO: sway 1.0.0 implements a fix for this, but it is lacking in the ^0.6.0 versions
+    //      so we duplicate the 1.0.0 implementation here. 
+    // Ref: https://github.com/apigee-127/sway/commit/1b7acd9743aae2f2f94057d3f8911e2af72614f5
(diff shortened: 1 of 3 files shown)


benbp:
👍

anhou:
:+1:

lib/api/2.0/tags.js
```diff
@@ -8,6 +8,19 @@ var nodes = injector.get('Http.Services.Api.Nodes');
 var tags = injector.get('Http.Services.Api.Tags');
 var _ = injector.get('_'); // jshint ignore:line
 var Errors = injector.get('Errors');
+var parseUrl = require('url').parse;
+
+function getTagName(req) {
+    //TODO: sway 1.0.0 implements a fix for this, but it is lacking in the ^0.6.0 versions
+    //      so we duplicate the 1.0.0 implementation here. 
+    // Ref: https://github.com/apigee-127/sway/commit/1b7acd9743aae2f2f94057d3f8911e2af72614f5
```
yyscamper: Could you please clarify why not just upgrade the sway to 1.0.0? Any known issue?

yyscamper: I see `sway` is not in the dependency list, so this is a fix you found from other code and copy it here? This line confused me `but it is lacking in the ^0.6.0 versions`.

yyscamper:
+1

lib/api/2.0/tags.js
zyoung51: @yyscamper: swagger-express-mw includes swagger-node-runner includes sway.  We do not control the dependency list of swagger-node-runner.

PR #274, opened 2016-05-09
add support for Api proxy

shanemsully:
In order to support running the RackHD on-taskgraph and on-http services on a host that does not have Layer-2 network access to the PXE network we need a host that does have access to the PXE network and can proxy requests to the southbound API endpoint over a management network.

The proxy server, nginx in my POC, adds headers to the requests coming from the PXE client that contain headers that will be used during profile, template, and task rendering.

The headers are
X-Real-IP - the actual IP address of the PXE client
X-RackHD-api-proxy-ip - The IP address of the API proxy server
X-RackHD-api-proxy-port - The port of the API proxy server

The proxy server URL is added to the lookup record for the PXE client so that it can be used during task rendering since the original HTTP request is not available at that time.

This is one of three pull requests. There will also be subsequent changes to on-core  and on-tasks to support this.

on-core PR: https://github.com/RackHD/on-core/pull/131
on-tasks PR: https://github.com/RackHD/on-tasks/pull/185

[test-coverage.tar.gz](https://github.com/RackHD/on-http/files/255454/test-coverage.tar.gz)

### DIFF
--- a/lib/services/profiles-api-service.js
+++ b/lib/services/profiles-api-service.js
@@ -22,7 +22,8 @@ di.annotate(profileApiServiceFactory,
         '_',
         'Profiles',
         'Services.Environment',
-        'Http.Services.Swagger'
+        'Http.Services.Swagger',
+        'Constants'
     )
 );
 function profileApiServiceFactory(
@@ -38,7 +39,8 @@ function profileApiServiceFactory(
     _,
     profiles,
     Env,
-    swaggerService
+    swaggerService,
+    Constants,
 ) {
 
     var logger = Logger.initialize(profileApiServiceFactory);
@@ -71,7 +73,8 @@ function profileApiServiceFactory(
         return _.flattenDeep([macs]);
     };
 
-    ProfileApiService.prototype.setLookup = function(query) {
+    ProfileApiService.prototype.setLookup = function(req) {
+        var query = req.query;
         if (query.mac && query.ip) {
             return waterline.nodes.findByIdentifier(this.getMacs(query.mac))
             .then(function (node) {
@@ -80,6 +83,14 @@ function profileApiServiceFactory(
                         query.ip,
                         query.mac
                     );
+                }
+            }).then(function () {
+                if ( req.get(Constants.HttpHeaders.ApiProxyIp) ) {
+                    var proxy =  'http://%s:%s'.format(
+                        req.get(Constants.HttpHeaders.ApiProxyIp),
+                        req.get(Constants.HttpHeaders.ApiProxyPort)
+                    );
+                    return waterline.lookups.upsertProxyToMacAddress(proxy, query.mac);
                 }
             });
         }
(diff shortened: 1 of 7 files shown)


lib/services/workflow-api-service.js
```diff
@@ -67,6 +69,9 @@ function workflowApiServiceFactory(
             var context = configuration.context || {};
             if (node) {
                 context = _.defaults(context, { target: node.id });
+                lookupService.nodeIdToProxy(node.id).then(function(proxy) {
+                    context.proxy = proxy;
```
benbp: I don't think this will work, because `self.createActiveGraph` below will get called before this promise resolves and the callback is executed.

benbp: It would need to be something like:

```
if (node) {
    context = _.defaults(context, { target: node.id });
    return lookupService.nodeIdToProxy(node.id).then(function(proxy) {
        context.proxy = proxy;
        return self.createActiveGraph(
                    definition, configuration.options, context, configuration.domain, true)
    });
}
return self.createActiveGraph(
            definition, configuration.options, context, configuration.domain, true);
```

benbp:
@shanemsully sorry for the merge delays here, looks like all these are ready to go, just need another rebase here.

yyscamper:
@shanemsully : Could you please fix the unit-test problem? I checked the log it is caused by your code.

shanemsully:
@yyscamper, The failures are due to dependencies with code changes in my on-core review (https://github.com/RackHD/on-core/pull/131)  I changed package.json locally to pull on-core from my forked repo end the unit tests pass.

yyscamper:
@shanemsully : OK, thanks

yyscamper:
:+1:

brianparry:
Dependent PRs have been merged, so Jenkins test this please

lib/services/profiles-api-service.js
```diff
@@ -81,6 +84,14 @@ function profileApiServiceFactory(
...
                     );
                 }
+            }).then(function () {
+                if ( req.get(Constants.HttpHeaders.ApiProxyIp) ) {
+                    var proxy =  'http://%s:%s'.format(
+                        req.get(Constants.HttpHeaders.ApiProxyIp),
+                        req.get(Constants.HttpHeaders.ApiProxyPort)
+                    );
+                    return waterline.lookups.upsertProxyToMacAddress(proxy, query.mac);
+                }
```
brianparry: @shanemsully Do we need to return the lookup record if proxy IP header isn't set in the request?  It looks like the returned Promise will resolve to undefined if the header is missing.  Also, it looks like we may want to pass the node returned by the previous query into the .then callback and return immediately if it is undefined.

shanemsully: @brianparry Yeah I see what you mean.  So should setLookup return the node only if it added it to the database and return undefined if it did not add anything?  It looks like that is what the code was doing previously.

brianparry: @shanemsully Yeah - I think that's the behavior we want.  It turns out that the profiles API doesn't look at the fulfillment value of the returned promise, so it's not an important detail at the moment but I think we should keep things consistent with the previous implementation.

brianparry:
:+1: Looks good.  Thanks for the contribution!  FYI - on-http unit tests occasionally fail with timeouts which is what happened during the last Travis run.  I've restarted Travis, I expect it to pass this time.

brianparry:
ok to test

PR #276, opened 2016-05-12
Fix incorrect wget argument in esxi kickstart file

zyoung51:
/usr/bin/wget --spider results in an "unrecognized option '--spider'"
error in ESXi 5.5 and ESXi 6.0.  Change the command option to the correct
"-s" argument


anhou:
:+1:  just curious about previous behavior, does this issue exist before, and still could work with this error?

zyoung51:
@anhou The previous command never worked but nothing was dependent on it working, so the error went unnoticed.

anhou:
@zyoung51 Gotta, Thanks!

zyoung51:
@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @jlongever

PR #279, opened 2016-05-13
Add callback API service to centos-ks installer

benbp:
This addresses an issue where we end up doing SSH validation *before* a reboot during a CentOS install. These changes add a "run-once" service that hits an API route to verify the OS is up and running post installation reboot.

Requires https://github.com/RackHD/on-taskgraph/pull/99 and RackHD/on-tasks#201

@RackHD/corecommitters @heckj @zyoung51 @VulpesArtificem @johren @stuart-stanley

### DIFF
--- a/data/templates/centos.rackhdcallback
+++ b/data/templates/centos.rackhdcallback
@@ -1,0 +1,31 @@
+#! /bin/bash
+#
+# centos.rackhdcallback       callback to rackhd post installation API hook
+#
+# description: calls back to rackhd post installation API hook
+#
+### BEGIN INIT INFO
+# Provides: centos.rackhdcallback
+# Required-Start:    $network
+# Default-Start:     3 4 5
+# Short-Description: Callback to rackhd post installation API hook
+# Description: Callback to rackhd post installation API hook
+### END INIT INFO
+
+
+# We can't really know when networking is actually up and running from
+# a simple "Required-Start: $network", so instead just use curl with
+# a bunch of retries and hope it works. We could use more sophisticated
+# dependency mechanisms provided by systemd, but ideally we can just use
+# a single script to service both CentOS 6.5 and CentOS 7 installs, hence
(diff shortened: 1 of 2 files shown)


data/templates/centos.rackhdcallback
```diff
@@ -0,0 +1,31 @@
...
+# a single script to service both CentOS 6.5 and CentOS 7 installs, hence
+# reasoning for using `wget --retry-connrefused` below. Using `curl --retry`
+# fails immediately because it only handles connection timeouts, not connection
+# refused cases.
+# See https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/
+
+# Nothing wrong with set -e here since we're not doing anything complex
+set -e
+echo "Attempting to call back to RackHD CentOS installer"
+wget --retry-connrefused --waitretry=1 -t 30 http://<%=server%>:<%=port%>/api/common/templates/<%=completionUri%>
```
jlongever: Is the file used or removed? Assuming this is the ssh pub key.

jlongever: Also.. `completionUri` seems to infer it's the complete URI path and not a file.

yyscamper: Yes, this is just used to send a signal to RackHD, sometime this downloaded file is not even used.
I had think about to add some new APIs in southbound to do such particular job, to signify the status or report progress. Recently, the secure erase task that we are working will definitely need such new API, because the secure erase often takes several hours even a whole days, the progress is so important to user in such case.

yyscamper:
+1

data/templates/centos.rackhdcallback
benbp: @jlongever it's a meaningless file in this specific case. I was just following the convention of `"completionUri": "renasar-ansible.pub"` that's used everywhere as a short-term step, though I agree it's probably better in the long run to have some sort of API for this type of stuff.

heckj:
:+1:

PR #282, opened 2016-05-17
Add API callbacks to ESXi installer

benbp:
This addresses an issue where we end up doing SSH validation before we're ready during an ESXi install. These changes add a script that hits an API route to verify the OS is up and running during the `%firstboot` phase of installation and on post installation reboot.

ESXi is much, much more annoying when it comes to creating custom services, or making persistent changes of any kind, so this isn't quite as clean as the solution for the CentOS installer (https://github.com/RackHD/on-http/pull/279), but it's the same pattern. The mechanisms used here assume ESXi 5.1 or greater (for older versions, we'd need to write to `/etc/rc.local` instead of `/etc/rc.local.d/local.sh`).

Requires https://github.com/RackHD/on-taskgraph/pull/102 and https://github.com/RackHD/on-tasks/pull/209

@RackHD/corecommitters @heckj @zyoung51 @VulpesArtificem @johren @stuart-stanley @richav1 @amymullins

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -78,6 +78,28 @@ server 1.vmware.pool.ntp.org
 __NTP_CONFIG__
 /sbin/chkconfig ntpd on
 
+# Download the service to callback to RackHD after OS installation/reboot completion
+# %firstboot ends with a reboot, this script will run afterwards to signify completion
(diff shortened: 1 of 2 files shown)


anhou:
:+1:

yyscamper:
+1

PR #283, opened 2016-05-17
[1.2] Add callback mechanisms to ESXi installations

benbp:
This addresses an issue where we end up doing SSH validation before we're ready during an ESXi install. These changes add a script that hits an API route to verify the OS is up and running during the %firstboot phase of installation and on post installation reboot.

ESXi is much, much more annoying when it comes to creating custom services, or making persistent changes of any kind, so this isn't quite as clean as the solution for the CentOS installer (#279), but it's the same pattern. The mechanisms used here assume ESXi 5.1 or greater (for older versions, we'd need to write to /etc/rc.local instead of /etc/rc.local.d/local.sh).

Requires RackHD/on-taskgraph#103 and RackHD/on-tasks#210

@RackHD/corecommitters @heckj @zyoung51 @VulpesArtificem @johren @stuart-stanley @richav1 @amymullins

### DIFF
--- a/data/templates/esx.rackhdcallback
+++ b/data/templates/esx.rackhdcallback
@@ -1,0 +1,22 @@
+#!/bin/sh
+# esx       callback to rackhd post installation API hook
+# description: calls back to rackhd post installation API hook
+
+echo "Attempting to call back to RackHD ESX installer"
+
+# *sigh*, busybox shell does not support {1..30}. Retry 30 times with 1 second
+# sleep in between.
+for retry in $(awk 'BEGIN { for ( i=0; i<30; i++ ) { print i; } }');
+do
+    wget -s http://<%=server%>:<%=port%>/api/common/templates/<%=completionUri%>
+    if [ "$?" -ne 0 ];
+    then
+        echo "Failed to connect to RackHD API callback, retrying"
+        sleep 1
+    else
+        exit 0
+    fi
+done;
+
+echo "Exceeded retries connecting to RackHD API callback. Exiting with failure code 1"
+exit 1
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -78,6 +78,28 @@ server 1.vmware.pool.ntp.org
 __NTP_CONFIG__
 /sbin/chkconfig ntpd on
 
+# Download the service to callback to RackHD after OS installation/reboot completion
+# %firstboot ends with a reboot, this script will run afterwards to signify completion
+# of the installer and all reboot steps.
+#
+# The approved method for adding startup commands is to write to /etc/rc.local.d/local.sh, 
+# which is a pre-existing file with a sticky bit set by VisorFS. You can't just create new
+# files and expect them to stick around, even if you set a sticky bit yourself.
+# The /sbin/auto-backup.sh script will ensure the changes are persisted across reboots and
+# MUST be executed after making any changes.
+#
+# See these links for more information:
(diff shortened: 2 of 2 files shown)


data/templates/esx.rackhdcallback
```diff
@@ -0,0 +1,22 @@
...
+# esx       callback to rackhd post installation API hook
+# description: calls back to rackhd post installation API hook
+
+echo "Attempting to call back to RackHD ESX installer"
+
+# *sigh*, busybox shell does not support {1..30}. Retry 30 times with 1 second
+# sleep in between.
+for retry in $(awk 'BEGIN { for ( i=0; i<30; i++ ) { print i; } }');
+do
+    wget -s http://<%=server%>:<%=port%>/api/common/templates/<%=completionUri%>
```
yyscamper: @benbp : have you verified this design on 1.2.0 branch. I ask this question because I found this URl will not success.
It will always fail due to this code: https://github.com/RackHD/on-http/blob/master/lib%2Fapi%2F1.1%2Fsouthbound%2Ftemplates.js#L46

The root cause is that the static IP that assigned during OS installation has not been inserted into the lookup table. The ARP poller may solve this problem, but the 1.2.0 is still monitoring the isc-dhcp-leased result. Apparently the static IP is not assigned by isc-dhcp-server, so the lookup entry will not contain such entry.

So we may have to change to another URL to ensure such pure notification works, I tried the static file that in the repository, but the static URL will not publish to the message queue of HTTP response.

Maybe it's the good time to design some new API to do such notification work, but seems this is not a quick fix for 1.2.0 release.

Any better thought?

yyscamper: I made a mistake, the static file request will also publish to the AMQP, so if we change the completionUri to a file in the OS repository, it will work.

The disadvantage is that the completionUri will vary with OS, even sometime with OS version.

zyoung51: @yyscamper Resolved with https://github.com/RackHD/on-http/pull/289 which uses lookups to add ip/mac/nodeid relationships on the newly created vmk's and static IPs.

zyoung51:
+1.  Jenkins failure is expected.  Jenkins uses master's version of integration tests which is testing functionality that is not present in the 1.2 branch.

data/templates/esx.rackhdcallback
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# esx       callback to rackhd post installation API hook
+# description: calls back to rackhd post installation API hook
+
+echo "Attempting to call back to RackHD ESX installer"
+
+# *sigh*, busybox shell does not support {1..30}. Retry 30 times with 1 second
+# sleep in between.
+for retry in $(awk 'BEGIN { for ( i=0; i<30; i++ ) { print i; } }');
```
zyoung51: But busybox does support arithmetic:
```
retry=0
while true; do
    wget -s http://<%=server%>:<%=port%>/api/common/templates/<%=completionUri%>
    if [ $? -eq 0 ]; then
        exit 0
    fi;
    retry=$((retry+1));
    if [ $retry -eq 30 ]; then 
        echo "exceeded retries"
        exit 1 
    fi;
    sleep 1
done;
```

yyscamper:
+1

data/templates/esx.rackhdcallback
benbp: Yeah, that would have been cleaner :)

PR #284, opened 2016-05-17
change uid to optional

PengTian0:
change uid to optional  for centos and Ubuntu;
and SUSE has supported uid optional, 
EXS don't support uid input.

@yyscamper @iceiilin @panpan0000


yyscamper:
:+1:

yyscamper:
Jenkins: ok to test

anhou:
:+1:  just let others know that SUSE has support uid optional, and esx don't support uid input

yyscamper:
@anhou :I had the same concern before, @PengTian0 had changed the doc accordingly.

PR #288, opened 2016-05-17
Removes WebSocket server

rolandpoulter:
Been working on removing the WebSocket functionality from `on-http`. This PR deletes all the WebSocket related code so it can be moved to a new module.

I've tested these changes locally using docker.

I've already create a person repo that can be transfered to RackHD, however this module requires a small change to on-core so that it doesn't need to include our DI.js dependency. See https://github.com/RackHD/on-core/pull/141


rolandpoulter:
Here is a link to the `on-wss` module that contains the WebSocket server for RackHD. https://github.com/rolandpoulter/on-wss

rolandpoulter:
@RackHD/corecommitters

yyscamper:
@rolandpoulter : so no existing functionality depends on the web-socket and it's safe to completely remove it right now? If future requirement depends on this, we can then includes your `on-wss` as new micro-service for RackHD?

rolandpoulter:
@yyscamper only the UI use WebSocket server, it doesn't require it and will function without websocket but the UI won't update in real time.

yyscamper:
@rolandpoulter thx.

:+1: for this PR.

benbp:
👍

PR #289, opened 2016-05-18
Post to lookups with static ip and newly created MAC

zyoung51:
To resolve an issue where the template cannot be rendered because the information
is not in the lookups table.  This change will post to /lookups as the entries
are created, setting the static ip, mac, nodeid relationship in the database, so
the entry can be found when the template is requested to signal the completion of
the kickstart.

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -34,6 +34,22 @@ rootpw <%=rootPlainPassword%>
 reboot
 
 %firstboot --interpreter=busybox
+postLookup () {
+  echo "Attempting postLookup operation on $1" >> /vmfs/volumes/datastore1/firstboot.log
+  mac=`esxcli --debug --formatter=csv network ip interface list | grep $1 | awk -F, '{print $3}'`
+  if [ "<%=version%>" == "5.5" ]; then
+    mac=`esxcli --debug --formatter=csv network ip interface list | grep $1 | awk -F, '{print $2}'`
+  fi
+  BODY="{"
+  BODY=$BODY"\"macAddress\": \"$mac\","
+  BODY=$BODY"\"ipAddress\": \"$2\","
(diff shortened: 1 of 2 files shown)


data/templates/esx-ks
```diff
@@ -166,8 +184,12 @@ cp /var/log/esxi_install.log "/vmfs/volumes/datastore1/firstboot-esxi_install.lo
   <% }); %>
 <% } %>
 
-#signify ORA the installation completed
+# signify ORA the installation completed, the 60s sleep is to wait for the LRU to timeout
+# the stale entries that it latched prior to the lookups loading above being set.  Without
+# the sleep, the completionUri hit below will not publish to the correct AMQP channel
+sleep 60
```
yyscamper: Good finding

yyscamper:
:+1:

data/templates/esx-ks
zyoung51: This is a temporary workaround for 1.2 branch to avoid changing the POST /lookups route to use something else in on-core that would invalidate the LRU entries when the entry is created directly in waterline.

yyscamper: OK， could you please add some story to track this problem? so we can improve this in master branch.

lib/api/1.1/southbound/router.js
```diff
@@ -25,6 +25,7 @@ function internalApiFactory (injector) {
         router.use(injector.get(require('./files')));
```
anhou: file mode is changing 644 to 755

anhou:
:+1: with keep file mode 644 and 'missing semicolon' fix

data/templates/esx-ks
zyoung51: @yyscamper I thought your work for the event sink api will be the master branch fix

yyscamper: OK, make sense.

tannoa2:
+1

PR #290, opened 2016-05-18
change all 'vlanId' to 'vlanIds'

changev:
1. old 'vlanId' should not be used anymore, 'vlanIds' is recommanded now. but 'vlanId' is still supported with  deprecated message.

2. this PR is integrated with https://github.com/RackHD/on-tasks/pull/215

@RackHD/corecommitters @yyscamper @anhou   @panpan0000 please review.

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -100,8 +100,8 @@ echo -e "NETWORKING=yes\nHOSTNAME=<%=hostname%>.<%=domain%>" > /etc/sysconfig/ne
   <% ipv6 = 0 %>
   <% networkDevices.forEach(function(n) { %>
     <% if( undefined != n.ipv4 ) { %>
-      <% if( undefined != n.ipv4.vlanId ) { %>
-        <% n.ipv4.vlanId.forEach(function(vid) { %>
+      <% if( undefined != n.ipv4.vlanIds ) { %>
+        <% n.ipv4.vlanIds.forEach(function(vid) { %>
           echo "Configuring vlan <%=vid%> on <%=n.device%>"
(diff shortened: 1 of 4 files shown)


yyscamper:
:+1:

anhou:
@changev Related documents  also need to be updated. http://rackhd.readthedocs.io/en/latest/rackhd/index.html#os-installation-workflow-support

zyoung51:
There was an open ask for this variable to be changed to a single integer instead of an array of strings. What was the result of that discussion?

yyscamper:
@zyoung51 : Sorry I don't remember there is such discussion or maybe I am not in the CC list.  Could you please forward that discussion to me?

zyoung51:
@yyscamper It was part of the ODR-666, ODR-669 discussions.  The point was made that vlanID should not be an array because assigning the same IP on all VLANs is an unrealistic use case.  I will forward the internal links to you.

yyscamper:
Thanks @zyoung51 , I didn't follow the discussion for both issues. So let's hold the PR until we have conclusion.

yyscamper:
@zyoung51 : After checking the discussion you forward to me,  the conclusion is that the VLANs should also be an array. so I think this PR is still valid and we can go ahead with merging.

yyscamper:
Meanwhile, now in RackHD code, `vlanIds` can be either array of string or array of number (Number is our recommendation), and both `vlanId` and `vlanIds` will be supported, so existing RackHD users don't need to change their code for this feature. see this PR: https://github.com/RackHD/on-tasks/pull/215

iceiilin:
+1

PR #291, opened 2016-05-18
Enhance get_driveid.js to get eUSB SSD catalog

lanchongyizu:
This addresses RackHD/RackHD#145 to enhance get_driveid.js to get eUSB SSD catalog, but it could not work well when there are two or more eUSB SSDs or there are other non-SCSI Adapters such as SATA.

@RackHD/corecommitters @yyscamper @anhou @panpan0000

### DIFF
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -28,7 +28,7 @@ function parseDriveWwid(idList) {
 
     //According to SCSI-3 spec, vendor specified logic unit name string is 60
 	//Only IDE and SCSI disk will be retrieved
-    var scsiLines = [], sataLines = [], wwnLines = [], requiredStrLen = 60;
+    var scsiLines = [], sataLines = [], wwnLines = [], usbLines = [], requiredStrLen = 60;
     lines.forEach(function(line){
         if ( line && !(line.match('part')) && line.match(/sd[a-z]$|hd[a-z]$/i)){
             var nameIndex = line.lastIndexOf('/'), idIndex = line.lastIndexOf('->');
@@ -40,6 +40,9 @@ function parseDriveWwid(idList) {
             }
             else if (line.indexOf('wwn') === 0) {
                 wwnLines.push([line.slice(nameIndex + 1), line.slice(0, idIndex)]);
+            }
+            else if (line.indexOf('usb') === 0) {
+                usbLines.push([line.slice(nameIndex + 1), line.slice(0, idIndex)]);
             }
         }
     });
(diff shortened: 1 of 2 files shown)


data/templates/get_driveid.js
```diff
@@ -185,6 +192,18 @@ function buildDriveMap(wwidData, vdData, scsiData) {
...
             }
         });
+        //vmhbaAdapter:CChannel:TTarget:LLUN
+        //Adapter is set to default value 32 for only one eUsb
+        if (esxiWwid[1] === "mpx.") {
+            esxiWwid[1] = '';
+            if (scsiId !== '') {
+                var hctl = scsiId.split(":");
+                if (hctl.length === 4) {
+                    esxiWwid[1] = "mpx.vmhba32:C".concat(hctl[1], ":T", hctl[2], ":L", hctl[3]);
```
yyscamper: What does  the value `32` mean?

yyscamper:
Why not support the multiple eUSB case? Is it not a normal hardware configuration?
@pengz1 @cgx027

data/templates/get_driveid.js
pengz1: 32 is the USB controller id. Servers use USB controller ID integrated in PCH. It is OK to only include one id here.

data/templates/get_driveid.js
```diff
@@ -101,13 +104,17 @@ function parseDriveWwid(idList) {
         return [esxiLine[0], 'naa.' + split[1].slice(1)];
     });
```
pengz1: Add an example here for eUSB

data/templates/get_driveid.js
yyscamper: Make sense

yyscamper:
:+1:

pengz1:
Please try to install ESXi w/ this new eUSB ID.
After that, +1

data/templates/get_driveid.js
```diff
@@ -185,6 +192,18 @@ function buildDriveMap(wwidData, vdData, scsiData) {
                 }
             }
         });
+        //vmhbaAdapter:CChannel:TTarget:LLUN
+        //Adapter is set to default value 32 for only one eUsb
+        if (esxiWwid[1] === "mpx.") {
+            esxiWwid[1] = '';
+            if (scsiId !== '') {
+                var hctl = scsiId.split(":");
+                if (hctl.length === 4) {
```
panpan0000: [minor]

@pengz1 @lanchongyizu 
in what kind of case will the scsiId  not be 4 digit ?( ex "8:0:0:0") ?
I meant , if that really happen, what consequence will the code turns out to be ?

pengz1: leave it to try catch...

yyscamper:
Tested on the real system, the parsed eUSB driveId can be recognized by ESXi installer, so it it a valid value. But the ESXi refuses the following installation and reports don't supporting vmfs on eUSB, this problem doesn't belong to our scope.

So this is still a valid PR to fix the driveId catalog on discovery the eUSB disk.

PR #292, opened 2016-05-18
Monorail API return arrays

codenrhoden:
Two commits in this one.

The main one stems from working with generated swagger code, and finding that most of the GET APIs actually return lists/arrays of objects, but the swagger definition is just `type: object` for a schema.  however, the `GET` for `/nodes` is defined correctly.  Fix this by putting in the correct schema for GET calls that return lists.

Conversely, there were a couple of schemas that said they return lists when they actually return a single object.

The second commit removes extraneous whitespace throughout the YAML file. It was getting hard to work with the YAML files in various editors because of the extra spaces throughout the file. Split that out into a second commit in case you don't accept whitespace fixes.

I have more PRs coming to fix the YAML file here -- depending on the order they are merged, re-basing on my side will be necessary.

### DIFF
--- a/static/monorail.yml
+++ b/static/monorail.yml
@@ -9,7 +9,7 @@ schemes:
 basePath: /api/1.1
 produces:
   - application/json
-  # x-gzip explicitly for files once the bugs listed 
+  # x-gzip explicitly for files once the bugs listed
   #   in comments below are fixed by swagger team.
   - application/x-gzip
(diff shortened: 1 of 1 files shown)


yyscamper:
Thanks @codenrhoden , :+1:

yyscamper:
Jenkins: ok to test

benbp:
👍

PR #293, opened 2016-05-18
API and SKU pack changes for new template loader.

brianparry:
* Change PUT /api/1.1/templates and /profiles put to use new loader semantics.
* Change PUT /api/2.0/views to use new loader semantics.
* Changes to skupack service to use templates/profiles.loadFile rather than put.

This change is dependent on https://github.com/RackHD/on-core/pull/143.  Tests will not pass until both changes are merged.

### DIFF
--- a/lib/api/1.1/northbound/profiles.js
+++ b/lib/api/1.1/northbound/profiles.js
@@ -75,21 +75,15 @@ function profilesRouterFactory (
      *     }
      */
 
-    router.put('/profiles/library/:identifier',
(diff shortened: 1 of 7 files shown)


brianparry:
@RackHD/corecommitters @zyoung51 Verified functional tests passed when https://github.com/RackHD/on-core/pull/143 is combined with this PR.

zyoung51:
+1

anhou:
+1

PR #294, opened 2016-05-18
Fix swagger spec 11

codenrhoden:
Add misising schema reference for default case that was left out in f708e0e9084a51e6ca02876f2b6f3ee39273d310

Fix swagger spec for body params:
According to the swagger spec, when defining parameters if `in` is `body`
then the `schema` field is required. Using `type` in the current way is
invalid, and actually crashes go-swagger.

### DIFF
--- a/static/monorail.yml
+++ b/static/monorail.yml
@@ -728,7 +728,8 @@ paths:
         - name: body
           in: body
           required: false
-          type: object
+          schema:
+            type: object
       tags:
         - workflow
         - task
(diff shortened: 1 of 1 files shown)


jlongever:
Jenkins, ok to test

codenrhoden:
Are these false positives? Nothing changed re: Redfish, and I also don't think the `monorail.yml` file actually serves as input to anything...

jlongever:
Looks like it.  Jenkin's test this please.

brianparry:
:+1:

benbp:
👍

PR #295, opened 2016-05-19
Update node definition for 1.1 swagger spec

codenrhoden:
There are several missing fields for the `node` definition, and some that have the wrong type.  Fix those by doing the following:

- Add missing fields
- sku comes back as the ID/string, not as an object.
- obmSettings is an array not an object

Here is an example node object I get back from the API, to see that it matches:

```
    {
        "autoDiscover": false,
        "catalogs": [],
        "createdAt": "2016-05-18T18:40:30.298Z",
        "id": "573cb71ed11f55c1075ec878",
        "identifiers": [
            "08:00:27:2e:69:a6"
        ],
        "name": "08:00:27:2e:69:a6",
        "obmSettings": [
            {
                "config": {},
                "service": "noop-obm-service"
            }
        ],
        "relations": [
            {
                "relationType": "enclosedBy",
                "targets": [
                    "572434ae23947dd00a5d8533"
                ]
            }
        ],
        "sku": "572516a873b83ec515d32719",
        "tags": [],
        "type": "compute",
        "updatedAt": "2016-05-18T18:41:44.183Z",
        "workflows": []
    }
```

FWIW, this is the last change I've had floating around to submit for the 1.1 swagger spec.  Thanks for taking a look!

### DIFF
--- a/static/monorail.yml
+++ b/static/monorail.yml
@@ -1895,14 +1895,12 @@ definitions:
     properties:
       id:
         type: string
-      profile:
-        type: string
       name:
         type: string
       obmSettings:
-        type: object
-      ipAddresses:
-        type: object
+        type: array
+      type:
+        type: string
       workflows:
         type: array
         items:
@@ -1912,9 +1910,20 @@ definitions:
         items:
           $ref: '#/definitions/catalog'
       sku:
-        $ref: '#/definitions/sku'
+        type: string
+      snmpSettings:
+        type: object
+      bootSettings:
+        type: object
+      sshSettings:
+        type: object
+      autoDiscover:
+        type: boolean
+      relations:
+        type: array
+      tags:
+        type: array
     required:
-    - profile
     - name
   graphobject:
     properties:


anhou:
+1

yyscamper:
By comparing the node db model (https://github.com/RackHD/on-core/blob/master/lib/models/node.js), I still see some inconsistent. For example, the yaml has an extra "ipAddress", and the yaml doesn't have "snmpSetting", "bootSetting", is there any auto-way to do this? @brianparry

yyscamper:
jenkins: ok to test

codenrhoden:
@yyscamper Thanks for the comments.  From what I've learned via Slack discussions, what I get back from the API call is a combination of the DB model, and some extra fields added by the ORM (fields like id, createAt, updatedAt, etc.).

Perhaps I should update this to just cover what is listed in the model?

I can also remove `ipAddress`, as it doesn't show up anywhere.

I'm being selfish with this PR - I'm updating the `node` definition because I am using it, but not going through other definitions to see if they are right are wrong. I went down this path because `tags` was missing and I wanted to use it.

As for an auto-way to do any of this -- my understanding is that the swagger spec for the 1.1 API is really just an add-on, and not used at all by the RackHD code. This is changed with the 2.0 API, which is a good thing. But I do not know if there is any automatic connection between the 2.0 YAML and the DB models.

yyscamper:
@codenrhoden:
I'm OK with only update the `node` first, that's better than nothing. So could you please at least keep the `node` definition is correct (all necessary properties are there and non-necessary properties are removed)?

As for the auto-way, I think it has something with our RackHD code arch, that's beyond the scope of your PR, I just want to raise this question, and maybe @brianparry had some thought about this before.

brianparry:
@yyscamper I don't know of any way to auto-generate Swagger definitions from waterline models.  Even if we could, I'm not sure we'd want to because not everything in the waterline model is necessarily included in the response output.  For example, some models may choose to delete properties in their toJSON methods.  That said, swagger uses a subset of json-schema for its model definitions.  For resources such as nodes that have a defined schema (see https://github.com/RackHD/on-http/blob/master/lib/serializables/v1/node.js#L39) we can probably just convert this to yaml and drop it into the swagger file.  Note that this only applies to the 1.1 API since schemas are done differently in the 2.0 API.

brianparry:
@codenrhoden I think we need to combine your swagger file changes together and verify that our Jenkins functional tests still work with the changes to the generated client library.

codenrhoden:
@yyscamper @brianparry I updated my commit here, but am not sure if it's 100% accurate yet.

Going from https://github.com/RackHD/on-http/blob/master/lib/serializables/v1/node.js#L39-L74 and https://github.com/RackHD/on-core/blob/master/lib/models/node.js#L20-L72, I added `snmpSettings`, `bootSettings`, `sshSettings`, `autoDiscover`, `relations`, and `tags`. I removed `profile` and `ipAddress`. and I modified `obmSettings` to be an array, `sku` to be a string.

The reason I say this might not be 100% is that there is still at least one field I cannot account for that is in the returns I see -- things like `updatedAt` come from the ORM, I get that, but where does `identifiers` come from? I guess it has to come from the ORM as well?  I left it out at this time.

yyscamper:
@codenrhoden : I am :+1: with your current code.

@brianparry : I think we can open an enhancement issue, since these schema data are placed in different files and even in different repos,  they will tend to diverge again if there is further schema change in future.

benbp:
@codenrhoden looks like identifiers comes from the ORM as well, but in the definition for our base model, rather than in the node model definition (https://github.com/RackHD/on-core/blob/fea46c9e5c7b3f665abe47cd6ec68f27a0c8b3cc/lib/common/model.js#L51-L54).

benbp:
👍

codenrhoden:
@benbp Thanks for pointing out where `identifiers` comes from!

PR #296, opened 2016-05-19
2.0 API templates route and tests

brianparry:
GET /api/2.0/templates/metadata -> Get metadata for all templates
GET /api/2.0/templates/metadata/:name -> Get metadata for all templates with name.
GET /api/2.0/templates/metadata/:name?scope=:scope -> Get a single template with specified name in specified scope
GET/PUT /api/2.0/templates/library/:name -> Get/put raw template with specified name in global scope
GET/PUT /api/2.0/templates/library/:name?scope=:scope -> Get/put raw template with specified name in specified scope.
GET /api/2.0/templates/:name -> Southbound templates route.

### DIFF
--- a/lib/api/2.0/templates.js
+++ b/lib/api/2.0/templates.js
@@ -1,0 +1,92 @@
+// Copyright 2016, EMC Inc.
+
+'use strict';
+
+var injector = require('../../../index').injector;
+var swagger = injector.get('Http.Services.Swagger');
+var controller = swagger.controller;
+var templates = injector.get('Templates');
+var taskProtocol = injector.get('Protocol.Task');
+var Errors = injector.get('Errors');
+var lookups = injector.get('Services.Lookup');
+var waterline = injector.get('Services.Waterline');
+var workflowApiService = injector.get('Http.Services.Api.Workflows');
+var _ = injector.get('_');
+
+// GET /templates/metadata
+var templatesMetaGet = controller(function() {
+    return templates.getAll();
+});
+
+// GET /templates/metadata/:name
+var templatesMetaGetByName = controller(function(req) {
+    return templates.getName(req.swagger.params.name.value,
+                             req.swagger.query.scope);
+});
+
+// GET /templates/:name
+var templatesGetByName = controller(function(req, res) {
+    return lookups.ipAddressToMacAddress(res.locals.ipAddress)
+    .then(function(mac) {
+        console.log(mac)
+        return waterline.nodes.needByIdentifier(mac);
+    })
+    .then(function(node) {
+        return Promise.all([
+            workflowApiService.findActiveGraphForTarget(node.id),
+            node
+        ]);
+    })
+    .spread(function(workflow, node) {
+        if (!workflow) {
+            throw new Errors.NotFoundError('no active workflow');
+        }
+        return Promise.all([
+            swagger.makeRenderableOptions(req, res, workflow.context),
+            taskProtocol.requestProperties(node.id)
+        ]);
+    })
+    .spread(function(options, properties) {
+        console.log(options);
+        console.log(properties);
+        return templates.render(
+            req.swagger.params.name.value,
+            _.merge({}, options, properties),
+           res.locals.scope
+        ) ;
+    });
+});
+
+// GET /templates/library/:name
+var templatesLibGet = controller(function(req, res) {
+    return templates.get(req.swagger.params.name.value,
+                         req.swagger.query.scope)
+    .then(function(template) {
+        if (!template || !template.contents) {
+            throw new Errors.NotFoundError('template not found');
+        }
+        return template.contents;
+    })
+});
+
+// PUT /templates/library/:name
+var templatesLibPut = controller(function(req, res) {
+    return templates.put(req.swagger.params.name.value, 
+                         req,
+                         req.swagger.query.scope);
+});
+
+// DELETE /templates/library/:name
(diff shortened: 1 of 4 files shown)


brianparry:
@RackHD/corecommitters @srinia6 @dalebremner @BillyAbildgaard

lib/api/2.0/templates.js
```diff
@@ -0,0 +1,89 @@
...
+// GET /templates/metadata/:name
+var templatesMetaGetByName = controller(function(req) {
+    return templates.getName(req.swagger.params.name.value,
+                             req.swagger.query.scope);
+});
+
+// GET /templates/:name
+var templatesGetByName = controller(function(req, res) {
+    return lookups.ipAddressToMacAddress(res.locals.ipAddress)
+    .then(function(mac) {
```
yyscamper: Need throw an error if mac is not found, see the API 1.1 (https://github.com/RackHD/on-http/blob/master/lib/api/1.1/southbound/templates.js#L44) and the discussion of ODR-682

spec/lib/api/2.0/templates-spec.js
```diff
@@ -0,0 +1,225 @@
...
+            });
+        });
+    });
+
+    describe('GET /templates/metadata/:name', function () {
+        it('should return a single template', function () {
+            templates.getName.resolves(template);
+            return helper.request().get('/api/2.0/templates/metadata/123')
+            .expect('Content-Type', /^application\/json/)
+            .expect(200)
```
yyscamper: Miss to expect the response body?

lib/api/2.0/templates.js
brianparry: @yyscamper `ipAddressToMacAddress` calls `lookups.findOneByTerm` which will throw a `Errors.NotFoundError` if the lookup record is not found

spec/lib/api/2.0/templates-spec.js
brianparry: @yyscamper Fixed

yyscamper:
:+1:

yyscamper:
@brianparry  : one more comment, please keep your unit-code in consistent style, I see sometimes you write following code:
```javascript
return helper.request('xxx')
    .expect(xxx)
    .expect(yyy)
```
But sometimes no indent for each `expect`.
```javascript
return helper.request('xxx')
.expect(xxx)
.expect(yyy)
```

Either style is OK, but need be in a consistent  style at least in the same file.

spec/lib/api/2.0/templates-spec.js
```diff
@@ -0,0 +1,225 @@
...
+            .then(function () {
+                expect(templates.render).to.have.been.calledOnce;
+                expect(templates.render).to.have.been.calledWith('123');
+            });
+        });
+    });
+
+    describe('2.0 DELETE /templates/library/:name', function () {
+        beforeEach('should PUT new mockfile', function () {
```
yyscamper: Should this `beforeEach` be `it`?

spec/lib/api/2.0/templates-spec.js
```diff
@@ -0,0 +1,225 @@
...
+            return helper.request().get('/api/2.0/templates/123')
+            .expect(200)
+            .then(function () {
+                expect(templates.render).to.have.been.calledOnce;
+                expect(templates.render).to.have.been.calledWith('123');
+            });
+        });
+    });
+
+    describe('2.0 DELETE /templates/library/:name', function () {
```
yyscamper: Seems the word `DELETE` should be removed from test-suite name.

spec/lib/api/2.0/templates-spec.js
yyscamper: If this `PUT` is to insert a mockfile so that  the `DELETE` can be success, why not use stub?

lib/api/2.0/templates.js
```diff
@@ -0,0 +1,89 @@
...
+    .then(function(template) {
+        if (!template || !template.contents) {
+            throw new Errors.NotFoundError('template not found');
+        }
+        return template.contents;
+    });
+});
+
+// PUT /templates/library/:name
+var templatesLibPut = controller(function(req, res) {
```
yyscamper: I want to confirm the functionality for this API. This is to insert a new template file or to replace the existing template file?
If it is insert,  the `POST` is reasonable, `PUT` is to replace existing file.

lib/api/2.0/templates.js
yyscamper: OK, that's good.

yyscamper:
:+1: after my comment is addressed or clarified.

srinia6:
+1

lib/api/2.0/templates.js
brianparry: The behavior is: _create_ or _replace_. Since it is idempotent, PUT is the appropriate HTTP method.

yyscamper: Make sense

spec/lib/api/2.0/templates-spec.js
yyscamper: I remembered mocha doesn't promise the execution sequence of all test cases within a test suite, mocha can sort out test cases, filtering test cases, so you cannot expect mocha will always run in your designed order: PUT --> GET -> DELETE -> GET ERROR.

So the each test case should be independent, all sequence should use the hook function `before`/`after`/`beforeEach`/`afterEach`.

srinia6:
above build error has been fixed....FYI

PR #301, opened 2016-05-24
[1.2] add PUT /lookups API to fix esx-ks installation timeout issue

yyscamper:
This is a hot fix for recent esx-ks installation timeout issue [ODR-728](https://hwjiraprd01.corp.emc.com/browse/ODR-728).

Unlike @benbp 's proposal to remove the duplicated lookup entry before installation, my solution is to add a new API "PUT /lookups",this API is to create or replace lookup entry. So if a duplicated lookup is sent, the lookup table will be updated.

I have tested this for ESXi installation with Quanta, all success. But the CentOS/RHEL installation will still fail as (unlike esxi) they don't try to PUT the lookup entry to RackHD, so node will fail to complete the wait-completion-uri. So another PR will be created to fix CentOS/RHEL.

@RackHD/corecommitters @benbp @richav1 @zyoung51 @jlongever @johren

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -47,7 +47,7 @@ postLookup () {
   BODY=$BODY"}"
   BODYLEN=$(echo -n ${BODY} | wc -c )
   echo ${BODY} >> /vmfs/volumes/datastore1/firstboot.log
-  echo -ne "POST /api/1.1/lookups HTTP/1.0\r\nHost: $2\r\nContent-Type: application/json\r\nContent-Length: ${BODYLEN}\r\n\r\n${BODY}" | nc -i 3 <%=server%> <%=port%>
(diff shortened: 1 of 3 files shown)


benbp:
👍 

While I'd like to stay out of the IP Address Management space in kickstart files (could be very confusing behavior if you don't know your kickstart script is changing existing lookups), I guess I'm okay with this since it's just for 1.2. In master, we can just create new entries only for node->mac and let the arp cache based lookups handle the IP address detection.

yyscamper:
@benbp : I totally agree with your opinion, the lookup management should be moved out of kickstart file, and the arp poller should improve a lot for such use cases. 

We had intensive discussion on this via mail, and we agree it is an invalid case right now to set static IP address to RackHD control network, and it's risk to merge this code at the last minutes before 1.2.0 release. So I will close this PR.

PR #303, opened 2016-05-25
Script to create jbod for each physical disk under a virtual disk

pengz1:
JBOS status is required for secure erase as some tools can't pass through RAID.
This script is to create JBOD for each physical disk under virtual disk. 
Below is an example to run the script:
sudo python create_jbod.py -v lsi -d '{"diskName": "/dev/sdr", "vd": "/c0/v18", "scsi":"0:2:18:0", "did": ["8", "24", "19","23","18"], "sid": ["/c0/e7/s19", "/c0/e7/s20","/c0/e7/s21", "/c0/e7/s22", "/c0/e7/s23"]}'
-v: used to specify raid controller type. We currently support two types: lsi and dell
-d: specify one virtual disk parameters. diskName/vd/scsi/did/sid are required
If there are multiple virtual disks, multiple -d options are required.

### DIFF
--- a/data/templates/create_jbod.py
+++ b/data/templates/create_jbod.py
@@ -27,15 +27,15 @@ def create_jbod(disk, tool):
     :param tool: tools used for JBOD creation, storcli and perccli are supported
     :return: a list contains disk OS names, like ["/dev/sda", "/dev/sdb", ...]
     '''
-    for slot_id in disk["sid"]:
+    for slot_id in disk["slotId"]:
         cmd = [tool, slot_id, "set", "jbod"]
         exit_code = subprocess.call(cmd, shell=False)
         assert exit_code == 0, "Can't create JBOD for drive " + slot_id
     time.sleep(1)  #Wait 1 second for OS to scan new drives
     disk_list = []
-    scsi_id_bits = disk["scsi"].split(":")
+    scsi_id_bits = disk["scsiId"].split(":")
     #map jbod to disk device name with JBOD
-    for did in disk["did"]:
+    for deviceId in disk["deviceId"]:
         #scsi id is used to map virtual disk to new JBOD
         #scsi id is made up of adapter:scsi:dev:lun as below:
         #   adapter id [host]: controller ID, ascending from 0.
@@ -45,7 +45,7 @@ def create_jbod(disk, tool):
         #   device id [target]: displayed as DID in Megaraid for each physical drives.
         #   LUN id [LUN]: Logic Unit Numbers, LUN is not used for drive mapping
         scsi_info = scsi_id_bits[:]
-        scsi_info[2] = str(did)
+        scsi_info[2] = str(deviceId)
         anti_patten = re.compile(":".join(scsi_info[0:3]))  #anti-patten to exclude scsi id for RAID
         scsi_info[1] = '[0-9]*'
         patten = re.compile(":".join(scsi_info[0:3]))
@@ -63,7 +63,7 @@ def create_jbod(disk, tool):
             if patten.search(line) and not anti_patten.search(line) and line.find("part") == -1:
             #lines contain part is partition info
                 disk_name = line.split("/")[-1]
-        assert disk_name, "Disk OS name is not found for did " + str(did)
+        assert disk_name, "Disk OS name is not found for deviceId " + str(deviceId)
         disk_list.append("/dev/" + disk_name)
     return disk_list
 
@@ -73,10 +73,10 @@ if __name__ == '__main__':
     #ARG_LIST.d should include at least following items as a string
     #   {
     #   "devName": "/dev/sdx"
-    #   "sid": "/c0/e252/sx"
-    #   "did": "0"
-    #   "vd": "/c0/vx"
-    #   "scsi": "0:0:0"
+    #   "slotId": "/c0/e252/sx"
+    #   "deviceId": "0"
+    #   "virtualDisk": "/c0/vx"
+    #   "scsiId": "0:0:0"
     #   }
     for argu in ARG_LIST.d:
         disk_argument_list.append(json.loads(argu))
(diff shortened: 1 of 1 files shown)


pengz1:
@RackHD/corecommitters @WangWinson @iceiilin

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+    disk_argument_list = []
+    #ARG_LIST.d should include at least following items as a string
+    #   {
+    #   "devName": "/dev/sdx"
+    #   "slotId": "/c0/e252/sx"
+    #   "deviceId": "0"
+    #   "virtualDisk": "/c0/vx"
+    #   "scsiId": "0:0:0"
+    #   }
+    for argu in ARG_LIST.d:
```
anhou: 'argu' could be 'arg'

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python
+
+# Copyright 2015, EMC, Inc.
+
+'''
+This script is to delete RAID and create JBOD for each physical disk of a virtual disk with RAID
```
anhou: Could add more description why delete RAID before create JBOD?

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+    disk_list_with_jbod = []
+    for disk_argument in disk_argument_list:
+        #if virtualDisk doesn't exit, push disk directly into disk list
+        if not disk_argument["virtualDisk"]:
+            disk_list_with_jbod.append("/dev/" + disk_argument["diskName"])
+        else:
+            command = [tool_path, "/c0", "set", "jbod=on"]
+            exit_status = subprocess.call(command, shell=False)
+            assert exit_status == 0, "Failed to enable jbod"
+            command = [tool_path, disk_argument["virtualDisk"], "del", "force"]
```
anhou: set JBOD then del RAID here, if funtioinality is fine, adjust the sequence to keep consistent with above description is better

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+    for argu in ARG_LIST.d:
+        disk_argument_list.append(json.loads(argu))
+    assert disk_argument_list != [], "no disk arguments includes"
+
+    #Idenfity tools used for raid operation
+    raid_controller_vendor = ARG_LIST.v
+    assert raid_controller_vendor in RAID_VENDOR_LIST, "RAID controller vendor info is invalid"
+    if raid_controller_vendor == "lsi":
+        tool_path = "/opt/MegaRAID/storcli/storcli64"
+    else:
```
anhou: could 'dell' is specified explictily here, because perccli is only for dell, and add else branch to print others are not support.

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+        tool_path = "/opt/MegaRAID/perccli/perccli64"
+    assert path.exists(tool_path), "Overlay doesn't include tool path: " + tool_path
+
+    disk_list_with_jbod = []
+    for disk_argument in disk_argument_list:
+        #if virtualDisk doesn't exit, push disk directly into disk list
+        if not disk_argument["virtualDisk"]:
+            disk_list_with_jbod.append("/dev/" + disk_argument["diskName"])
```
anhou: diskName is not the consistant with 'devName": "/dev/sdx" annotation above, one place need to change

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+RAID_VENDOR_LIST = ["lsi", "dell"]  ##RackHD supported RAID controller vendor list
+
+def create_jbod(disk, tool):
+    '''
+    Create JBOD for each physical disk under a virtual disk.
+    :param disk: a dictionary contains disk argument
+    :param tool: tools used for JBOD creation, storcli and perccli are supported
+    :return: a list contains disk OS names, like ["/dev/sda", "/dev/sdb", ...]
+    '''
+    for slot_id in disk["slotId"]:
```
WangWinson: convention: 
should we use `slotIds` instead ? Since it's array.

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+        cmd = [tool, slot_id, "set", "jbod"]
+        exit_code = subprocess.call(cmd, shell=False)
+        assert exit_code == 0, "Can't create JBOD for drive " + slot_id
+    time.sleep(1)  #Wait 1 second for OS to scan new drives
+    disk_list = []
+    scsi_id_bits = disk["scsiId"].split(":")
+    #map jbod to disk device name with JBOD
+    for deviceId in disk["deviceId"]:
```
WangWinson: should we use `deviceIds` instead ? Since it's array.

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+import subprocess
+import argparse
+import json
+import re
+import time
+from os import path as path
+
+ARG_PARSER = argparse.ArgumentParser(description=
+                                     'delete virtual disk and create JBODs for physical disks')
+ARG_PARSER.add_argument("-d", action="append", default=[], type=str, help="Create JBOD")
```
yyscamper: Since type is string, so why default is `[]`

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+This script is to delete RAID and create JBOD for each physical disk of a virtual disk with RAID
+'''
+import subprocess
+import argparse
+import json
+import re
+import time
+from os import path as path
+
+ARG_PARSER = argparse.ArgumentParser(description=
```
yyscamper: As name convention, `ARG_PARSER` is not constant, so it should not be all in upper cases.

data/templates/create_jbod.py
```diff
@@ -0,0 +1,109 @@
...
+    #   }
+    for argu in ARG_LIST.d:
+        disk_argument_list.append(json.loads(argu))
+    assert disk_argument_list != [], "no disk arguments includes"
+
+    #Idenfity tools used for raid operation
+    raid_controller_vendor = ARG_LIST.v
+    assert raid_controller_vendor in RAID_VENDOR_LIST, "RAID controller vendor info is invalid"
+    if raid_controller_vendor == "lsi":
+        tool_path = "/opt/MegaRAID/storcli/storcli64"
```
yyscamper: Suggest moving the tool path to a dictionary at the beginning of this script, so it's easy to maintain if we add new tool or we changed to tool path.

PR #307, opened 2016-05-27
Add southbound SSDP notifications

zyoung51:
- addMembership for the proper router to the SSDP server
- advertise the 2.0 southbound endpoint

### DIFF
--- a/lib/services/upnp-service.js
+++ b/lib/services/upnp-service.js
@@ -45,9 +45,10 @@ function UPnPServiceFactory(
         this.options = options || {};
         this.options.ssdpSig = 'node.js/' + nodeVersion + ' uPnP/1.1 on-http';
         this.options.ssdpTtl = 2;
-        this.locations = _.filter(configuration.get('httpEndpoints', []), 
+        this.northbound = _.filter(configuration.get('httpEndpoints', []), 
                          _.matches({routers:'northbound-api-router'}));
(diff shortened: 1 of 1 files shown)


lib/services/upnp-service.js
```diff
@@ -137,14 +148,24 @@ function UPnPServiceFactory(
...
                 server.addUSN(r.urn);
                 server.on('advertise-alive', _alive.bind(self)); //jshint ignore: line
                 server.on('advertise-bye', _bye.bind(self));    //jshint ignore: line
-                server.start(self.bindAddress);
+                server.start();
+                server.sock.on('listening', function() {
+                    try {
+                        server.sock.addMembership(server._ssdpIp, ep.address);
+                    } catch(e) {
+                        logger.debug('addMembership', e);
```
yyscamper: A minor suggestion.
could you please elaborate the debug message here. As `addMembership` is so general, it's better to show the message is related with  upnp service.

yyscamper:
:+1:

lib/services/upnp-service.js
zyoung51: The logger is already tagged with the Upnp service name from the providers export:
[debug] [2016-05-27T19:03:48.745Z] [on-http] [Http.Services.uPnP] [Server] addMembership
 -> /lib/services/upnp-service.js:166

PR #309, opened 2016-05-27
Fix for Issue #235

brianparry:
* Add contents field to 1.1 API templates/profiles library routes.

Resolves https://github.com/RackHD/RackHD/issues/235

### DIFF
--- a/lib/api/1.1/northbound/templates.js
+++ b/lib/api/1.1/northbound/templates.js
@@ -12,13 +12,15 @@ di.annotate(templatesRouterFactory, new di.Provide('Http.Api.Templates'));
 di.annotate(templatesRouterFactory,
     new di.Inject(
 		'Templates',
-		'common-api-presenter'
+		'common-api-presenter',
+        'Promise'
 	)
 );
 
 function templatesRouterFactory (
     templates,
-    presenter
+    presenter,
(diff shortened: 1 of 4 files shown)


brianparry:
@RackHD/corecommitters @johren

srinia6:
+1

lib/api/1.1/northbound/templates.js
```diff
@@ -32,8 +34,12 @@ function templatesRouterFactory (
      */
 
     router.get('/templates/library', function (req, res) {
-        presenter(req, res)
-            .render(templates.getAll());
+        Promise.map(templates.getAll(), function(t) {
+            return templates.get(t.name);
+        })
+        .then(function(t) {
+            presenter(req, res).render(t);
```
benbp: The presenter can render an array?

dalebremner:
+1

lib/api/1.1/northbound/templates.js
brianparry: @benbp Presenter.render can handle arrays.  I think the method name `render` may be a little misleading because it doesn't actually do any rendering; it ultimately just deals with promises and calls res.json.

benbp:
👍

PR #310, opened 2016-05-30
Refine ubuntu installation mirror directory

anhou:
when the ubuntu installation's payload's mirror directory is '/mirrors/ubuntu' not '/ubuntu' like  "repo": "http://172.31.128.1:9080/mirrors/ubuntu"  ubuntu os installation workflow cannot work. refine it to be compatible with ubuntu mirror directory like  '/mirrors/ubuntu' 

@RackHD/corecommitters @WangWinson @pengz1

### DIFF
--- a/data/templates/install-ubuntu/ubuntu-preseed
+++ b/data/templates/install-ubuntu/ubuntu-preseed
@@ -24,8 +24,12 @@ d-i netcfg/wireless_wep string
 ### Mirror settings
 # If you select ftp, the mirror/country string does not need to be set.
 d-i mirror/country string manual
-d-i mirror/http/hostname string <%=server%>:<%=port%>
-d-i mirror/http/directory string /ubuntu
+# Get server:port from repo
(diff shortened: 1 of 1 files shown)


WangWinson:
:+1: after CI pass

yyscamper:
:+1:

benbp:
Maybe I'm misunderstanding what's happening here, but wouldn't it be better to just make these task definition variables that could or could not reference each other? Putting regex replacement logic into the template itself seems like something that could be very hard to debug in the future should problems/variations arise.

anhou:
@benbp I had considered breaking repo url to repoHostname and repoDirectory for all OS in install-os task definiton, but I think this is a special case for Ubuntu, and this way is not straightforward compared with a repo url from user perspective. and let repoHostname, repoDirectory to be handled in template, not in install-os.js is intented to make ubuntu repo handling independent.  and regex is also simple, each of them is one line javascript code. If it hadmore logic, I also intended to move them to install-os.js.

benbp:
Thanks @anhou. Yep, makes sense to me, normalizing the task definition. I imagine @yyscamper's proposal/work to allow for different schema's will alleviate issues like this going forward.

PR #312, opened 2016-06-01
API 2.0 OBM Routes

dalebremner:
@RackHD/corecommitters @jlongever @zyoung51 @srinia6 @BillyAbildgaard 

Depends on:
RackHD/on-core#164
RackHD/on-tasks#243

### DIFF
--- a/lib/api/2.0/obms.js
+++ b/lib/api/2.0/obms.js
@@ -4,35 +4,74 @@
 
 var injector = require('../../../index.js').injector;
 var controller = injector.get('Http.Services.Swagger').controller;
-var obms = injector.get('Http.Services.Api.Obms');
+var schemaApiService = injector.get('Http.Api.Services.Schema');
+var nameSpace = '/api/2.0/obms/definitions/';
+var waterline = injector.get('Services.Waterline');
+var Errors = injector.get('Errors');
+var nodes = injector.get('Http.Services.Api.Nodes');
+var _ = injector.get('_'); // jshint ignore:line
 
-
-/**
- * @api {get} /api/2.0/obms/library GET /library
- * @apiVersion 2.0.0
- * @apiDescription get list of possible OBM services
- * @apiName obms-library-get
- * @apiGroup obms
- * @apiSuccess {json} obms list of the available obms.
- */
-var getObmLib = controller(function() {
-    return obms.getObmLib();
+// GET /obms/definitions
+var obmsDefinitionsGetAll = controller(function() {
+    return schemaApiService.getNamespace(nameSpace);
 });
-/**
- * @api {get} /api/2.0/obms/library/:identifier GET /library/:identifier
- * @apiVersion 2.0.0
- * @apiDescription get a single OBM service
- * @apiName obms-library-service-get
- * @apiGroup obms
- * @apiParam {String} identifier The obm service name.
- * @apiParamExample {String }Identifier-Example:
- *      "amt-obm-service"
- */
-var getObmLibById = controller(function(req) {
-    return obms.getObmLibById(req.swagger.params.identifier.value);
+
+// GET /obms/definitions/{name}
+var obmsDefinitionsGetByName = controller(function(req) {
+    var schemaUid = nameSpace + req.swagger.params.name.value;
+    var schema = schemaApiService.getSchema(schemaUid);
+    if (schema) {
+        return schema;
+    }
+    throw new Errors.NotFoundError(schemaUid + ' Not Found');
+});
+
+// GET /api/2.0/obms
+var obmsGet = controller(function(req) {
+    return waterline.obms.find(req.swagger.params.query.value);
+});
+
+// PUT /api/2.0/obms
+var obmsPut = controller({success: 201}, function(req) {
+    var nodeId = req.swagger.params.body.value.nodeId;
+    delete req.swagger.params.body.value.nodeId;
+    return waterline.obms.upsertByNode(nodeId, req.swagger.params.body.value);
+});
+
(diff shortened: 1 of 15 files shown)


benbp:
I'm less familiar with how these are all linked together, so pardon me if this is a bad suggestion, but it seems like it would be simpler to just have one abstract "OBM" schema, with #ref links for the config field, since all of the variation takes place in that field. It seems a little circular to name a schema like "ipmi-obm-service" then also require a service field for that schema.

benbp:
I'm less familiar with how these are all linked together, so pardon me if this is a bad suggestion, but it seems like it would be simpler to just have one abstract "OBM" schema, with #ref links for the config field, since all of the variation takes place in that field. It seems a little circular to title a schema like "ipmi-obm-service" then also require a service field for that schema that is just going to equal "ipmi-obm-service". I'm wondering something like:

```
{
    "title": "obm-service",
    "definitions": {
        "Obm": {
            "description": "",
            "type": "object",
            "properties": {
                "service": {
                    "type": "string"
                },
                "config": {
                    "$ref": "#/definitions/<filled in dynamically>"
                }
            },
            "required": ["service", "config"]
        }
    }
}
```

Does this make sense or totally off base?

dalebremner:
@benbp Totally makes sense Ben... Thanks! However, we decided to create separate schema files to facilitate adding/deleting schemas with a future API...

static/schemas/obms/amt-obm-service.json
```diff
@@ -0,0 +1,29 @@
+{
+    "title": "amt-obm-service",
+    "definitions": {
```
yyscamper: Why not put it under `defintions` rather than `properties`? My understanding is the `definitions` is something like `typedef` in C language, but `properties` is where you put properties that the object contains.

static/schemas/obms/ipmi-obm-service.json
```diff
@@ -0,0 +1,33 @@
...
+                },
+                "config": {
+                    "type": "object",
+                    "properties": {
+                        "host": {
+                            "description": "Host address",
+                            "type": "string"
+                        },
+                        "user": {
+                            "description": "Host username",
```
yyscamper: ipmi user name

static/schemas/obms/ipmi-obm-service.json
```diff
@@ -0,0 +1,33 @@
...
+            "type": "object",
+            "properties": {
+                "service": {
+                    "type": "string"
+                },
+                "config": {
+                    "type": "object",
+                    "properties": {
+                        "host": {
+                            "description": "Host address",
```
yyscamper: bmc address

static/schemas/obms/ipmi-obm-service.json
```diff
@@ -0,0 +1,33 @@
...
+                            "type": "string"
+                        },
+                        "user": {
+                            "description": "Host username",
+                            "type": "string"
+                        },
+                        "password": {
+                            "description": "Host password",
```
yyscamper: ipmi password

static/schemas/obms/panduit-obm-service.json
```diff
@@ -0,0 +1,54 @@
...
+                                    "type": "string"
+                                },
+                                "community": {
+                                    "type": "string"
+                                },
+                                "cyclePassword": {
+                                    "type": "string"
+                                },
+                                "pduNumber": {
+                                    "type": "string"
```
yyscamper: shouldn't be a `integer`?

static/schemas/obms/panduit-obm-service.json
```diff
@@ -0,0 +1,54 @@
...
+                                },
+                                "outletNumber": {
+                                    "type": "string"
```
yyscamper: integer?

static/schemas/obms/amt-obm-service.json
dalebremner: Hi @yyscamper. Changed descriptions and integers types, as you suggested...

Could you clarify your comment regarding "definitions"? I'm not clear about what you would like to see under "definitions"...

yyscamper: @dalebremner : this is the definition of `definitions` in the JSON-Schema: http://json-schema.org/latest/json-schema-validation.html#anchor94:
```
This keyword plays no role in validation per se. Its role is to provide a standardized location for schema authors to inline JSON Schemas into a more general schema.
```
If you want it in validation, then it's better not place it in `definitions`, if you just want to provide a partial of a larger schema, place it in `definition` is reasonable.  So just to confirm the usage?

srinia6:
+1

lib/api/2.0/obms.js
```diff
@@ -4,35 +4,74 @@
...
+});
+
+// POST /api/2.0/obms/led
+var obmsPostLed = controller(function(req) {
+    if (_.has(req, 'swagger.params.body.value.nodeId')) {
+        var nodeId = req.swagger.params.body.value.nodeId;
+        delete req.swagger.params.body.value.nodeId;
+        return nodes.postNodeObmIdById(nodeId, req.body);
+    } else {
+        throw new Errors.BadRequestError('Must specify nodeId in body');
```
brianparry: @dalebremner Shouldn't schema validation catch this?

brianparry:
@dalebremner :+1:, but looks like you need to rebase.  Also one minor question about schema validation for the LED setting route.

lib/api/2.0/obms.js
dalebremner: There's currently no schema for this, but I suppose there could be...

PR #313, opened 2016-06-01
refactor nodes/workflows

BillyAbildgaard:
refactored nodes/workflows to be consistent with the /workflows route

@RackHD/corecommitters @dalebremner @srinia6 @brianparry

### DIFF
--- a/lib/api/2.0/nodes.js
+++ b/lib/api/2.0/nodes.js
@@ -7,6 +7,7 @@ var controller = injector.get('Http.Services.Swagger').controller;
 var nodes = injector.get('Http.Services.Api.Nodes');
 var workflowApiService = injector.get('Http.Services.Api.Workflows');
 var _ = injector.get('_'); // jshint ignore:line
+var constants = injector.get('Constants');
 
 var nodesGetAll = controller(function(req) {
     return nodes.getAllNodes(req.query);
@@ -62,17 +63,26 @@ var nodesGetPollersById = controller(function(req) {
 });
 
 var nodesGetWorkflowById = controller(function(req) {
+    var newQuery;
+    var swaggerQuery = req.swagger.query || [];
+    if (_(swaggerQuery).has('active')) {
+        if (swaggerQuery.active) {
+            newQuery = ({
+                _status: constants.Task.ActiveStates
+            });
+        } else {
+            newQuery = ({
+                _status: {'!': constants.Task.ActiveStates}
+            });
+        }
+        newQuery = _.merge({}, newQuery, req.query);
+    } else {
+        newQuery = req.query;
+    }
+
     return workflowApiService.getWorkflowsByNodeId(req.swagger.params.identifier.value,
-                                                   req.query)
-        .then(function(workflows){
-            if( _(req.swagger.query).has('active')) {
-                return _(workflows)
-                    .filter(function(workflow) {
-                        return workflow.active() === req.swagger.query.active;
-                    });
-            }
-            return workflows;
-        });
+                                                   newQuery);
+
 });
 
 var nodesPostWorkflowById = controller({success: 201}, function(req) {
--- a/lib/api/2.0/nodes.js
+++ b/lib/api/2.0/nodes.js
@@ -65,7 +65,7 @@ var nodesGetWorkflowById = controller(function(req) {
     return workflowApiService.getWorkflowsByNodeId(req.swagger.params.identifier.value,
                                                    req.query)
         .then(function(workflows){
-            if(req.swagger.query && _(req.swagger.query).has('active')) {
+            if( _(req.swagger.query).has('active')) {
                 return _(workflows)
                     .filter(function(workflow) {
                         return workflow.active() === req.swagger.query.active;
--- a/lib/api/2.0/nodes.js
+++ b/lib/api/2.0/nodes.js
@@ -5,6 +5,7 @@
 var injector = require('../../../index.js').injector;
 var controller = injector.get('Http.Services.Swagger').controller;
 var nodes = injector.get('Http.Services.Api.Nodes');
+var workflowApiService = injector.get('Http.Services.Api.Workflows');
 var _ = injector.get('_'); // jshint ignore:line
 
 var nodesGetAll = controller(function(req) {
@@ -61,7 +62,17 @@ var nodesGetPollersById = controller(function(req) {
 });
 
 var nodesGetWorkflowById = controller(function(req) {
-    return nodes.getNodeWorkflowById(req.swagger.params.identifier.value);
+    return workflowApiService.getWorkflowsByNodeId(req.swagger.params.identifier.value,
+                                                   req.query)
+        .then(function(workflows){
+            if(req.swagger.query && _(req.swagger.query).has('active')) {
+                return _(workflows)
+                    .filter(function(workflow) {
+                        return workflow.active() === req.swagger.query.active;
+                    });
+            }
+            return workflows;
+        });
 });
 
 var nodesPostWorkflowById = controller({success: 201}, function(req) {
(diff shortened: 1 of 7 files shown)


lib/api/2.0/nodes.js
```diff
@@ -61,24 +62,45 @@ var nodesGetPollersById = controller(function(req) {
 });
 
 var nodesGetWorkflowById = controller(function(req) {
-    return nodes.getNodeWorkflowById(req.swagger.params.identifier.value);
+    return workflowApiService.getWorkflowsByNodeId(req.swagger.params.identifier.value,
+                                                   req.query)
+        .then(function(workflows){
+            if(req.swagger.query && _(req.swagger.query).has('active')) {
```
benbp: Not a big deal, but FYI the null check here is unnecessary since lodash will handle that for you.

```
> var l = require('lodash')
undefined
> l(undefined).has('active')
false
> l(null).has('active')
false
```

lib/api/2.0/nodes.js
```diff
@@ -61,24 +62,45 @@ var nodesGetPollersById = controller(function(req) {
...
+                return _(workflows)
+                    .filter(function(workflow) {
+                        return workflow.active() === req.swagger.query.active;
```
benbp: Couldn't we just update the getWorkflowsByNodeId to take an `active` option that just returns zero or one documents, rather than getting the entire workflow history (which could be quite large) and iterating through it?

spec/lib/api/2.0/nodes-spec.js
```diff
@@ -603,24 +600,39 @@ describe('Http.Api.Nodes', function () {
...
+            };
 
+            waterline.graphobjects.find.resolves(node.workflows);
+
+            return helper.request().get('/api/2.0/nodes/123/workflows')
+                .expect('Content-Type', /^application\/json/)
+                .expect(200, node.workflows)
+                .then(function(res){
+                    expect(res.body[0]).to.have.property('status', 'pending');
+            });
```
benbp: Indentation

lib/api/2.0/nodes.js
BillyAbildgaard: @benbp I was trying to make this consistent with the way it is done in the workflows route.  I believe the original intent to do post filtering was so it would be more database agnostic

BillyAbildgaard:
working on addressing the function test issues in another PR.  Travis had a time out, but worked locally so....

test this please

BillyAbildgaard:
Functional test issues resolved in https://github.com/RackHD/RackHD/pull/260

Both of these PRs are dependent on each other.  In addition, pypi needs to be updated as well.  

As far as the order for merging: 
I believe PR should get merged, pypi should be updated, and then the functional test PR should get merged

BillyAbildgaard:
@benbp addressed your comments.  Thanks for the tips about lodash! :)

BillyAbildgaard:
timeout error.

test this please

BillyAbildgaard:
travis, test this please

tannoa2:
+1

lib/api/2.0/nodes.js
```diff
@@ -61,24 +62,45 @@ var nodesGetPollersById = controller(function(req) {
 });
 
 var nodesGetWorkflowById = controller(function(req) {
-    return nodes.getNodeWorkflowById(req.swagger.params.identifier.value);
+    return workflowApiService.getWorkflowsByNodeId(req.swagger.params.identifier.value,
```
benbp: I still think this should be changed to just get the required document from the query. If the worry is about being database agnostic, then I would suggest the query should be handled at the ORM layer rather than doing an iterative approach at this layer. Part of my concern is that I've seen developers in the past calling this route _a lot_ when building automation on top of us, primarily just to check if the node has an active workflow or not, and because doing it this way can potentially be so expensive, I think it's possible for us to run into a performance impact here.

brianparry: @benbp So, I think your concern is valid, and I'm fine with changing this query for now.  The problem we have is that Cassandra does not support AND queries, so it's not something that can easily be abstracted away by the ORM.  Effectively we'd be doing the same thing just in a different place.  This will probably require changes to the models themselves to allow a direct query of active workflows associated with a node.

benbp: @brianparry it doesn't support AND queries? Just looking at the CQL docs (https://cassandra.apache.org/doc/cql/CQL.html) I'm finding examples like `SELECT ... WHERE KEY >= 'AM' and KEY =< 'AZ' AND module = 17;`. Though this definitely isn't my wheelhouse so I could be misunderstanding.

I think regardless I'd be definitely more in favor of a model change as you suggest to optimize queries like this, or perhaps creating new tables and/or collections that store these relationships so that we can make direct queries.

benbp: Reading up more on Cassandra, I guess if we wanted to support it it might make sense to have a wholly different data model given how different Cassandra is from other DBs we might use. At that point though, I guess I'm questioning what value there is in supporting Cassandra altogether?

brianparry: @benbp I'm not sure where the requirement for Cassandra comes from, but it's something that @zyoung51 has been working on for a while.  One possible way to address this particular problem is to add an `activeOnNode` property to workflows.  The value associated with this property would either be the node ID on which the workflow is active or `null` if the workflow is either inactive or detached from a node.  It would need to be updated atomically with the `_status` property on transitions between active and inactive states.  Then the active query becomes `{activeOnNode: nodeId}`. The drawback is that it becomes more difficult to list inactive nodes.  We could do it by taking the difference of the set of workflows active on a node and the complete set of workflows associated with a node but this has the potential for concurrency problems since a node could become active between the two queries.

benbp: @brianparry it would be easy from the workflow engine side to mark an active or inactive key. Since it's convenient with mongo, we do most things by inference from a couple fields now (state, node, etc.), but it's just as easy to be very explicit and add fields which contain the AND logic within them, and update them accordingly when we do our write queries.

BillyAbildgaard:
@benbp @brianparry @zyoung51 @jlongever 

updated query usage so we are now doing the query before passing it into the database.

passed all unit tests and functional tests locally.  Dependencies still remain on the Functional test PR ( https://github.com/RackHD/RackHD/pull/260 ) and on pypi being updated

benbp:
👍  thanks @BillyAbildgaard !

BillyAbildgaard:
@tannoa2 you gave a plus 1, but there were some changes to how we're doing queries now.  You might want to review again after my latest commit

lib/api/2.0/nodes.js
```diff
@@ -61,24 +63,55 @@ var nodesGetPollersById = controller(function(req) {
 });
 
 var nodesGetWorkflowById = controller(function(req) {
-    return nodes.getNodeWorkflowById(req.swagger.params.identifier.value);
+    var newQuery;
+    var swaggerQuery = req.swagger.query || [];
```
zyoung51: Should this be || {}; ?

zyoung51:
+1, looks good to me

lib/api/2.0/nodes.js
```diff
@@ -61,24 +63,55 @@ var nodesGetPollersById = controller(function(req) {
...
+            newQuery = ({
+                _status: constants.Task.ActiveStates
+            });
+        } else {
+            newQuery = ({
+                _status: {'!': constants.Task.ActiveStates}
+            });
+        }
+        delete swaggerQuery.active;
+        newQuery = _.merge({}, swaggerQuery, newQuery, req.query);
```
brianparry: @BillyAbildgaard swaggerQuery shouldn't be included here because can include parameters that don't directly map to workflow model properties.

lib/api/2.0/nodes.js
```diff
@@ -61,24 +63,55 @@ var nodesGetPollersById = controller(function(req) {
...
+    } else {
+        newQuery = _.merge({}, swaggerQuery, req.query);
```
brianparry: @BillyAbildgaard No need to merge swaggerQuery with req.query.  You can just use req.query directly if no active parameter is present.

BillyAbildgaard:
@brianparry I was just doing it that way in case we add more swagger query parameters in the future. That way they'll still get passed in once active is handled and removed from the query

BillyAbildgaard:
@brianparry made the change you recommended

lib/api/2.0/nodes.js
```diff
@@ -61,24 +63,55 @@ var nodesGetPollersById = controller(function(req) {
...
+        if (swaggerQuery.active) {
+            newQuery = ({
+                _status: constants.Task.ActiveStates
+            });
+        } else {
+            newQuery = ({
+                _status: {'!': constants.Task.ActiveStates}
+            });
+        }
+        delete swaggerQuery.active;
```
brianparry: Delete is not needed.

brianparry:
@BillyAbildgaard :+1: Thanks for making these changes.

PR #314, opened 2016-06-02
List on-http static files

lacarb:
Api path to static/monrail-2.0.yaml
Added lib/services/files-api-service.js
Added lib/api/2.0/files.js
Added spec/lib/services/files-api-services-spec.js

curl http://localhost:9090/api/2.0/files/static/list | jq .
Example output
{
  "uri": "/common/vmlinuz-3.16.0-25-generic"
},
{
  "uri": "dell-racadm-get-bios-graph.json",
  "sku": "Dell R730"
},


lacarb:
@benbp @VulpesArtificem

yyscamper:
@lacarb : please squash the commit history

package.json
```diff
@@ -48,8 +48,10 @@
     "swagger-tools": "~0.9.11",
     "tar": "~2.2.1",
     "tv4": "~1.2.7",
+    "walk": "~2.3.9",
```
anhou: add twice here

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+var fs = require('fs');
+var path = require('path');
+var express = require('express');
+var app = express();
+var walking = require('walk');
+
+module.exports = filesApiServiceFactory;
+
+di.annotate(filesApiServiceFactory, new di.Provide('Http.Services.Api.Files'));
+di.annotate(filesApiServiceFactory,
```
anhou: why not leveraging "lib/services/file-service.js"?

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+   */
+
+  FilesApiService.prototype.getAllStaticFiles = function() {
+
+    app.use(express.static(onhttpPath));
+    var staticPath = "static/http";
+    var skuPath = "skupack.d";
+    var directories = [staticPath,
+                       skuPath,
+                       "/opt/monorail/static/http"];
```
anhou: hard coded path, it could be got from config.json

anhou:
@lacarb Any requirements to get static file's list? on-http static service should be loosely-coupled with on-http service, maybe it could  enhance on-http static service to get file list from get URL , like when get http://archive.ubuntu.com/ubuntu/ , it will show directories and files list.

lib/api/2.0/files.js
```diff
@@ -4,6 +4,7 @@
 
 var injector = require('../../../index.js').injector;
 var controller = injector.get('Http.Services.Swagger').controller;
+var files = injector.get('Http.Services.Api.Files');
```
brianparry: To avoid confusion, maybe we should call this Http.Services.Api.StaticFiles

benbp: Agreed.

lib/services/files-api-service.js
benbp: Yeah, this should be synced with all the static directories in on-http. Might make sense to make these a shared constant array value or something.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Get list of static files
+   * @returns {array}
+   */
+
+  FilesApiService.prototype.getAllStaticFiles = function() {
+
+    app.use(express.static(onhttpPath));
```
benbp: What is this for?

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+    var staticPath = "static/http";
+    var skuPath = "skupack.d";
+    var directories = [staticPath,
+                       skuPath,
+                       "/opt/monorail/static/http"];
+
+    var skuInfo = {};
+    fs.readdirAsync(skuPath).then(function(list) {
+      list.forEach(function(item){
```
benbp: Should be using four spaces instead of two.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+
+    app.use(express.static(onhttpPath));
+    var staticPath = "static/http";
+    var skuPath = "skupack.d";
+    var directories = [staticPath,
+                       skuPath,
+                       "/opt/monorail/static/http"];
+
+    var skuInfo = {};
+    fs.readdirAsync(skuPath).then(function(list) {
```
benbp: So this whole async call chain isn't related to anything, so essentially it either has no bearing on the rest of the function's code, or is extremely likely to generate race conditions (given that you modify the function-global variable `skuInfo`).

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+      list.forEach(function(item){
+        if (fs.lstatSync(path.join(skuPath, item)).isDirectory()){
+          waterline.skus.findOne({id : item}).then(function(sku) { 
+            skuInfo[item] = sku.name;
+          });
+        }
+      });
+    });
+
+    function walkDirectory(dir) {
```
benbp: If you make this function a prototype method of `FilesApiService` then you can unit test it independently.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+      return new Promise(function(resolve) {
+        var walkResults = [];
+        walker = walk.walk(dir, {followLinks: false});
+    
+        walker.on("file", function(root, fileStats, next) {
```
benbp: I'm not entirely certain, but I think that because you don't clean up this listener, every time getAllStaticFiles is called we end up with a new walker/EventEmitter object that just hangs around in memory, resulting in a minor memory leak.

benbp:
-1 in the current state because of some asynchronous/cleanup issues, but shouldn't be hard to fix and turn into a +1 :)

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+      'Services.Waterline'
+    )
+);
+
+
+function filesApiServiceFactory(
+  Promise,
+  waterline
+) {
+  var walker;
```
yyscamper: RackHD code prefers indent 4 spaces

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+var fs = require('fs');
+var path = require('path');
+var express = require('express');
+var app = express();
```
yyscamper: Seems it is not necessary to create an express instance.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+    var skuPath = "skupack.d";
+    var directories = [staticPath,
+                       skuPath,
+                       "/opt/monorail/static/http"];
+
+    var skuInfo = {};
+    fs.readdirAsync(skuPath).then(function(list) {
+      list.forEach(function(item){
+        if (fs.lstatSync(path.join(skuPath, item)).isDirectory()){
+          waterline.skus.findOne({id : item}).then(function(sku) {
```
yyscamper: The promise is not returned, so the execution timing is not correct.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,107 @@
...
+    var staticPath = "static/http";
+    var skuPath = "skupack.d";
+    var directories = [staticPath,
+                       skuPath,
+                       "/opt/monorail/static/http"];
+
+    var skuInfo = {};
+    fs.readdirAsync(skuPath).then(function(list) {
+      list.forEach(function(item){
+        if (fs.lstatSync(path.join(skuPath, item)).isDirectory()){
```
yyscamper: Use `async` version lstat rather than `sync` version

lib/api/2.0/files.js
lacarb: Should this extend to renaming "files-api-service.js" to something like "staticfiles-api-service.js"?

lib/services/files-api-service.js
benbp: Yes, definitely. Otherwise this single API call is blocking for all other API calls.

benbp: You can probably use `Promise.filter` here (http://bluebirdjs.com/docs/api/promise.filter.html):

```
fs.readdirAsync(skuPath).then(function(list) {
    return Promise.filter(list, function(item) {
        return (fs.lstatAsync(path.join(skuPath, item))
        .then(function(stat) {
            return stat.isDirectory();
        });
    })
}).then(function(item) {
    return waterline.skus.findOne({id : item});
}).then(function(sku) { 
        skuInfo[item] = sku.name;
});
```

anhou:
@lacarb when you get a chance, could you have a feedback for my questions above? because I think on-http static file service could be separated with on-http service, this PR will bring trouble if on-http static service is separated. and we could get static file list from static file service directly, if RackHD don't support it yet. we could enhance it. did I misunderstood your purpose? if yes, please let me know. so temporarily:-1:  before it's clear.

lib/services/files-api-service.js
lacarb: That line is used to enable the relative paths I set for within on-http and needs to be updated. I found that the paths to static files were inconsistent with what I was told. For example, on my monorail server the default value of httpStaticRoot is `/opt/monorail/static/http` which does not exist in my instance. The paths below are the presumed likely places to check for static files.
The above line sets the working directory to be wherever the api call was sent from. In this case it will always be on-http

lib/services/files-api-service.js
lacarb: This has been addressed and will be put in the next push

lib/services/files-api-service.js
lacarb: I should have checked with others. I saw that the file structure used to implement it was named differently than other api calls so I made one which I saw as consistent. Would you advise that I move the staticfiles api call into "lib/services/file-service.js"?

benbp:
@lacarb I understand you're still working on this. I'm going to close for now, and you can re-open when you're ready :)

PR #302, opened 2016-05-24
Os install params

benbp:
This will make it easier to point our OS installer workflows toward custom kickstart/preseed files, and in general make changes in this location.

I'm inclined to not add documentation as I consider these more system specific options, since the kickstart files contain special callback scripts for our workflows. However, It's better that they're not hardcoded in the templates themselves (It is much, much harder to trace what files link to what files with OS installers, and will be easier to just look at a task definition and see all the files that will be downloaded).

Requires https://github.com/RackHD/on-tasks/pull/218
Docs: https://github.com/RackHD/docs/pull/283

@RackHD/corecommitters @heckj 

@zyoung51 I also updated the common-api-presenter rendering to add `<%=api.lookups|templates|profiles%>` support (same for task rendering in the linked PR).

PR #315, opened 2016-06-02
Pass boot config template instead of install script template to esx i…

benbp:
…nstaller

Unfortunately I missed this in testing to my latest update to https://github.com/RackHD/on-http/pull/302

We are passing in the wrong variable for the boot config template (passing in the kickstart instead of the boot script which has the kernel options, and points to the kickstart). This will fix the early installation error:

```
kernel=<FILEPATH> must be set in http://172.31.128.1:9080/api/current/templates/esx-ks
Fatal error: 32 (Syntax)
```

Just ran a dry run of ESX with these options and things work now.

@RackHD/corecommitters @jlongever @WangWinson

### DIFF
--- a/data/profiles/install-esx.ipxe
+++ b/data/profiles/install-esx.ipxe
@@ -1,10 +1,10 @@
 iseq ${platform} efi && goto is_efi || goto not_efi
 
 :not_efi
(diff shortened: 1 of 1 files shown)


jlongever:
+1

heckj:
👍

WangWinson:
:+1:

anhou:
:+1:

PR #317, opened 2016-06-03
Allow proxy lookup to fail in workflow service 

jlongever:
- Some targets may not have a lookup entry (e.g. Redfish enclosure node only has OBM settings).
- Remove all listeners in upnp service spec.

@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2 @keedya @shanemsully


benbp:
👍

PR #320, opened 2016-06-06
Add a script to remove BMC user credentials

keedya:


keedya:
@RackHD/corecommitters @zyoung51 @jlongever @uppalk1 @tannoa2

PR #322, opened 2016-06-07
Enable snmpSettings for switch nodes

pscharla:
Adding in support for snmpSettings for switch nodes to enable snmp pollers. I put community setting straight in the code - there may be a better place for it, but not sure.

@RackHD/corecommitters @benbp @heckj

### DIFF
--- a/lib/services/profiles-api-service.js
+++ b/lib/services/profiles-api-service.js
@@ -99,7 +99,7 @@ function profileApiServiceFactory(
         return Promise.resolve();
     };
 
-    ProfileApiService.prototype.getNode = function(macAddresses, options) {
+    ProfileApiService.prototype.getNode = function(macAddresses, ip, options) {
         var self = this;
         return waterline.nodes.findByIdentifier(macAddresses)
         .then(function (node) {
@@ -123,7 +123,7 @@ function profileApiServiceFactory(
 
                 });
             } else {
-                return self.createNodeAndRunDiscovery(macAddresses, options);
+                return self.createNodeAndRunDiscovery(macAddresses, ip, options);
             }
         });
     };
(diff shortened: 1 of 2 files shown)


pscharla:
Working on updates to unit tests...

lib/services/profiles-api-service.js
```diff
@@ -213,6 +215,14 @@ function profileApiServiceFactory(
             });
         })
         .then(function () {
+            if (options.type === 'switch') {
+                return waterline.nodes.updateByIdentifier(
+                    node.id,
+                    { snmpSettings: { host: ip, community: community} }
```
benbp: I'm not sure there's any guarantee that these settings will work though. The switch config could not enable snmp or could specify a different community. It seems more versatile to me to expose this functionality as a task that just calls `updateByIdentifier`, but at least in that case a user can change the settings via the API rather than them being hardcoded.

pscharla: It's true that there would be more flexilibity in exposing as a task.  I did think it was pretty safe however since likely we will control startup-config settings and it will probably be deployed only through the automated discovery process. Do you think users will create and deploy their own startup-config files?  The nice part of doing it this way, in my mind, is it is all automated.

benbp: @pscharla I do think users will create their own startup-config files, and can use RackHD to deploy them more easily. Even if we do control the startup config settings all the way through, I think it would be better regardless to have this in a task, because then the values (snmp community, etc.) could be synced between the startup config template and the node document via graph options.

pscharla: @benbp Works for me. I'll kill this PR and move the work to a task.

PR #323, opened 2016-06-10
Update 2.0 api workflow operations to operate on instanceId attribute

benbp:
We should not be using the mongo ID for anything, but rather passing around the uuid.v4() instanceId attribute. The API docs reference instanceId, but the code does not, so updating for accuracy.

@RackHD/corecommitters 

Requires an integration test change: https://github.com/RackHD/RackHD/pull/268 which is why Jenkins is failing.


yyscamper:
:+1:

yyscamper:
jenkins: test this please

benbp:
@yyscamper tests passing require the linked RackHD PR in the description.

PR #325, opened 2016-06-14
[1.2] Set bmc ip only when ip source is static when flashing quanta bmc

iceiilin:
Fix ODR 771 in flashing quanta BMC scripts, about error to set BMC IP after flashing Quanta BMC, if ip source is DHCP. 

It happened in MN lab, where BMC IP source is DHCP. Though the "sudo ipmitool lan set 1 ipaddr xxx" after flashing returns the same message as in static IP source, the exit code is 1 rather than 0. Modify the script to skip these network setting steps if the source is DHCP.

Add retry count after mc reset so that total wait time (maximum) is 90s, the same as suggested by vendor after flashing BMC. The original value is 60s, which is not sufficient for stack 13 node 1 at a certain ratio.

@uppalk1 @zyoung51

### DIFF
--- a/data/templates/flash_quanta_bmc.sh
+++ b/data/templates/flash_quanta_bmc.sh
@@ -60,50 +60,56 @@ until [ $counter -le 0 ]; do
            echo "Check didn't pass, ip source wasn't set properly"
            exit 1;
         fi;
-        sleep 2
 
-        #IPMI tool command to set IP address
-        ipmitool lan set 1 ipaddr $ipaddrInfo
-        sleep 8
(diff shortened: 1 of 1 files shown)


yyscamper:
:+1:

WangWinson:
:+1:

anhou:
:+1:

anhou:
if some PR message is also added to codes annotation, it's more readable in future's codes maintaince

iceiilin:
@anhou add comments.

anhou:
Thanks @iceiilin

PR #329, opened 2016-06-15
Fix debian package builder script

jlongever:
Only search for `.in` template files in debianstatic when rendering variables.

@RackHD/corecommitters @ytjohn @zyoung51 @uppalk1 @tannoa2 @keedya

### DIFF
--- a/build-package.bash
+++ b/build-package.bash
@@ -50,7 +50,7 @@ cat > /tmp/sed.script << EOF
 s%{{name}}%${3}%
 EOF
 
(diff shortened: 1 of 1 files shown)


heckj:
👍

jlongever:
Don't merge this,  the python library packages aren't building right.

ytjohn:
good catch

jlongever:
Should be good now +1

anhou:
:+1:

PR #338, opened 2016-06-21
WIP PR Redact passwords

DavidjohnBlodgett:
resolves https://github.com/RackHD/RackHD/issues/241

@RackHD/corecommitters 

just wanted to get this code up for some reviews, open to suggestions for changes.

(test is broke, currently a place holder)


lib/services/workflow-api-service.js
```diff
@@ -237,7 +237,39 @@ function workflowApiServiceFactory(
...
+                                        break;
+                                    }
+                                }
+                                if ( !alreadyFound ) {
+                                    walked.push( workflow[ property ]);
+                                    stack.push( { workflow : workflow[ property ], stack: item.stack + '.' + property });
+                                }
+                            } else {
+                                if ( property.match( /password/i )) {
+                                    workflow[property] = '*** REDACTED ***';
```
jlongever: Does it make more sense to just omit this field from the model?

lib/services/workflow-api-service.js
```diff
@@ -237,7 +237,39 @@ function workflowApiServiceFactory(
...
+                var walked = [];
+                var stack = [ { workflow : workflow , stack : '' } ];
+                while ( stack.length > 0 ) {
+                    var item = stack.pop();
+                    var workflow  = item.workflow;
+                    for ( var property in workflow ) {
+                        if ( workflow.hasOwnProperty( property )) {
+                            if ( typeof workflow[ property ] == "object" ) {
+                                var alreadyFound = false;
+                                for ( var i = 0; i < walked.length; i++ ) {
```
jlongever: The lodash or native forEach methods would work here too.

lib/services/workflow-api-service.js
```diff
@@ -237,7 +237,39 @@ function workflowApiServiceFactory(
     WorkflowApiService.prototype.getWorkflowsByNodeId = function(id, query) {
         var nodeId = ({node: id});
         var mergedQuery = _.merge({}, nodeId, query);
-        return waterline.graphobjects.find(mergedQuery);
+        return waterline.graphobjects.find(mergedQuery).then(
+            function sanitize(workflow) {
```
benbp: I think it would be better to define this function as its own prototype method for clearer/easier unit testing.

spec/lib/services/workflow-api-service-spec.js
```diff
@@ -376,6 +376,20 @@ describe('Http.Services.Api.Workflows', function () {
         });
     });
 
+    it('should return a sanitized workflow document', function(){
+        var workflow = {
+            id       : 'testgraphid',
+            _status  : 'pending',
+            password : 'testPassword'
```
benbp: I'd like to see more unit tests here. Does it work if password is nested one level down? What about two, three levels down?

For example, we'll see objects like this:
```
{
    definition: {
        options: {
            <task-label>: {
                password: "foo"
            }
        }
    }
}
```

lib/services/workflow-api-service.js
zyoung51: I think it would be nice to define the sanitize function within the model code in on-core rather than here.

lib/services/workflow-api-service.js
DavidjohnBlodgett: @jlongever So I had this discussion with a few people, and I'm of the camp that by redacting the field it insures there is no confusion in testing... if this output is used to verify the workflow has all the expected keys.

lib/services/workflow-api-service.js
DavidjohnBlodgett: I'm open to a refactor in either direction... besides... this current commit has the code in the wrong spot anyway... sorry :( my next push with correct that...

spec/lib/services/workflow-api-service-spec.js
DavidjohnBlodgett: so I did fully vet the code, and it will traverse nested objects and arrays... I agree the unit tests can and will be expanded... that was just a place holder with a simple test.

lib/services/workflow-api-service.js
jlongever: What confusion does it prevent? If the property is omitted then the test checks that there is no `password` field in the output,  as opposed to matching a value of `*** REDACTED ***` ?   Tests that run validations to check credential based operations are in place to ensure an actual set password is effective (regardless if it's in the response body).

yyscamper:
At least I know we have 4 kinds of redactions:
- https://github.com/RackHD/on-core/blob/master/lib/serializables/log-event.js#L140-L208
- https://github.com/RackHD/on-http/blob/master/lib/serializables/v1/obm.js
- (Ongoing PR): https://github.com/RackHD/on-core/pull/164/files
- Yours.

It's not good to write different redaction code, I think we can create a "redaction-service" in on-core, so every module can leverage that code to do redaction. So I give my :-1: for this PR and I think this is a chance to refactor for current redaction design.

jlongever:
Couldn't we just use the toJson() serializers in waterline models to omit fields we don't want to expose?  Why invent new services or random blobs of code that redact things differently?

yyscamper:
@jlongever , some of redaction doesn't belong to waterline, such as the LogEvent and the serializable in on-http. So we don't have an unified place to do the redaction.

lib/services/workflow-api-service.js
yyscamper: Actually this function is duplicated with https://github.com/RackHD/on-core/blob/master/lib/serializables/log-event.js#L190-L208

jlongever:
@yyscamper Isn't the serializable stuff in on-http being superseded by schema validation?  The LogEvent logs data coming out of the database.  How do you ensure this new redaction service will be used across the codebase and won't just be yet another redaction method?  IMHO, the dB/waterline access is the centralized place to "redact" (omit) sensitive properties such as passwords.

yyscamper:
@jlongever : I agree the source of serializable stuff is from waterline, but the source of LogEvent is not all from waterline.  For example, an IPMI related job need the ipmi-obm-setting, I first get the obm-setting from waterline, the password cannot be redacted because I need the real password to send IPMI command. So a copy of password exists in the memory as a property of that job. If the job is failed, I may dump all properties of that job to log file for debugging (The dumping like `logger.error` is to call the `print` function of LogEvent), the real password may be leaked here. I think we don't want to visit the waterline again and get a redacted version data before dumping.

The central redaction service is for sake of code reuse and generate uniform redaction result, we don't want multi module to implement a same function to scan all nesting properties to do redaction accordingly, and we also don't want the some module redacts to `[REDACTED]`, but other module redacts to `[***REDACTED***]`. After the central redaction service is ready, if any module want to redact some data, it just call redaction service, it will eventually generate a redacted result but the module never need to touch the detail.

jlongever:
@yyscamper If a developer leaves stale debug logs that leaks secrets then that change should violate any security requirements and be rejected, not to mention the case where the developer uses `console.log` and fails to call those new redaction service methods.  I'll have to disagree on that approach,  personally, I'd prefer to see secret properties encrypted/omitted close to the source (i.e. from within the model).

yyscamper:
@jlongever : If we can control all sources, I agree it will be better. If the source is from db, we can easily control it because it will always pass a common path. If the source is from memory, then it is very hard to control all of them, if we can successfully inspect the security defect in advance then that's better. But we may always have a miss, if so the LogEvent is the final guard.

So our current redaction is to control both source and terminal, I agree controlling source should take precedence. Controlling terminal is to bring some convenient when we use the logger and also acts as the last guard to catch up some missing.

brianparry:
@yyscamper @jlongever @DavidjohnBlodgett A couple of additional comments on this:
1. The plan is to deprecate serializeables with the 1.1 API, so any confusion that these cause should be short lived.
2. I agree with @jlongever and others that encryption and redaction should be entirely contained within the models and we should be using `toJSON()` or overriding model class methods like `find` and `findOne` to delete sensitive data.  That's what I've done here: https://github.com/RackHD/on-core/pull/164/files

DavidjohnBlodgett:
@yyscamper @jlongever @DavidjohnBlodgett @brianparry  So sorry for the late response... needed so time to digest this info... so it's clear, my approach is not abstracted enough to pass checkin. I am alittle unsure if, from the comments by brian, that we are already beginning to implement a fix? I do think tagging sensitive data in the DB and restricting query access to collections that are not of the correct security level to be a good way to go... Just unsure of any path for me to contribute. Any thoughts?

yyscamper:
@DavidjohnBlodgett : We will discuss this topic on tomorrow's CC meeting.  For your case, the redaction should belong to the db model (https://github.com/RackHD/on-core/blob/master/lib/models/graph-object.js) or somewhere that is common so that every model has the functionality of redaction. if so, when you want to retrieve the a graph object from database, the result will be auto redacted.
As for the detail, please wait the CC meeting's result.

brianparry:
@DavidjohnBlodgett I haven't started any work on generalized encryption / redaction handling, I just implemented a similar change in the OBMs model.  I'd suggest moving this change to the graphobjects model for now and we'll work through the details of refactoring into a more general solution later.

DavidjohnBlodgett:
closing, this will want to be a PR against on-core.

PR #340, opened 2016-06-22
Change swagger yaml file for processing by swagger-ui

zyoung51:
- The swagger-ui code throws an exception while parsing the schema
  object if it is not a reference.  Change to reference to make
  swagger-ui happy.
- Added the missing api/2.0 tag to the node ssh GET


anhou:
:+1:

keedya:
:+1:

PR #342, opened 2016-06-27
Fix adding DNS search domain not working in esx-ks template and set domain optional

lanchongyizu:
**Fix ODR-748** Adding DNS search domain not working in esx-ks template
**solution:**
In the section that processes 'dnsServers,' this
esxcli network ip dns search add <%=domain%>
needs to be changed to this
esxcli network ip dns search add --domain=<%=domain%>

**Fix ODR-751** Setting 'domain' option is effectively required in esx-ks
**solution:**
According to RackHD docs, the default value of 'domain' is 'rackhd', and now it's changed to null. rackhd/on-tasks#250

@RackHD/corecommitters @panpan0000

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -79,7 +79,12 @@ esxcli system settings advanced set -o /UserVars/ESXiShellInteractiveTimeout -i 
 esxcli network firewall set --default-action false --enabled no
 
 # set hostname and domain
-esxcli system hostname set --domain <%=domain%> --host <%=hostname%>
+# if hostname is 'localhost', domain cannot be set.
+<% if (domain) { %>
+esxcli system hostname set --host <%=hostname%> --domain <%=domain%>
+<% } else { %>
+esxcli system hostname set --host <%=hostname%>
+<% } %>
 
 #config root account
 <% if (typeof rootSshKey !== 'undefined') { %>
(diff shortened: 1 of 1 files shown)


anhou:
@lanchongyizu I think it's worth that other OS should have the same behavior that domain default is null,  so could you help create another story to track this, and after that update it in RackHD/docs?

anhou:
:+1:

lanchongyizu:
@anhou OK, I have created a story for your proprosal.

data/templates/esx-ks
```diff
@@ -79,7 +79,12 @@ esxcli system settings advanced set -o /UserVars/ESXiShellInteractiveTimeout -i
 esxcli network firewall set --default-action false --enabled no
 
 # set hostname and domain
-esxcli system hostname set --domain <%=domain%> --host <%=hostname%>
+# if hostname is 'localhost', domain cannot be set.
+<% if (domain) { %>
```
yyscamper: There is a special case that if the `domain` is not specified rather than `null`, this rendering will fail as it cannot find the reference of `domain`. If the Task-Schema takes effect, the domain cannot be set to null, so your code will fail. I suggest changing to following syntax: 
https://github.com/RackHD/on-http/blob/master/data/templates/centos-ks#L83

Meanwhile, you still need to delete the `domain` from options list if it is null, just like the `kvm`:
https://github.com/RackHD/on-tasks/blob/master/lib/jobs/install-os.js#L251-L253

yyscamper:
:+1:

PR #344, opened 2016-06-28
Fix double npm install in HWIMO-BUILD

lanchongyizu:
The redfish-node package dependency was removed from RackHD.
So the workaround should be cleared to avoid **double npm install** which takes **almost double** time to build deb package.

@RackHD/corecommitters @panpan0000 @jlongever @davidzuhn

### DIFF
--- a/HWIMO-BUILD
+++ b/HWIMO-BUILD
@@ -18,13 +18,7 @@ rsync -ar ../debianstatic/on-http/ debian
 rm -rf node_modules
 npm install --cache=`pwd`
 npm run apidoc
-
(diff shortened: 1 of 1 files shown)


panpan0000:
+1

benbp:
👍 
Would it be even simpler to just call `npm install --production` instead of `npm install` + `npm prune --production`, since it doesn't seem like anything is being done with the devDependencies?

lanchongyizu:
@benbp You could notice that **npm run apidoc** is done before it gets pruned, which is devDependency.

anhou:
:+1:

PR #345, opened 2016-06-28
Create linux diag profile

jlongever:
Added Linux diag boot profile for EMC diagnostic images.

@RackHD/corecommitters @uppalk1 @tannoa2 @zyoung51 @keedya


benbp:
👍

benbp:
Would the profile name be clearer if it was just like "boot-initrd" or something?

jlongever:
Probably,  or `linux-generic.ipxe` ?  The diag image booting can take special `bootargs` which is just a string of `mem=<map>` and other parameters,  so I guess the name could be generic.

jlongever:
@benbp renamed it to `linux-generic.ipxe`

anhou:
@jlongever  A minor comment at https://github.com/RackHD/on-http/pull/345

jlongever:
@anhou This follows the same pattern as `linux.ipxe`.  Are you suggesting there is a design issue here? Is this a gating issue for you?

PR #346, opened 2016-06-28
fixing issue #259 and unit test cleanup

BillyAbildgaard:
resolves https://github.com/RackHD/RackHD/issues/259

The actual fix for the bug was the line I added to the swagger_render.  @zyoung51 helped me find that fix.

@RackHD/corecommitters @dalebremner @srinia6 @brianparry

### DIFF
--- a/lib/fittings/swagger_render.js
+++ b/lib/fittings/swagger_render.js
@@ -6,6 +6,7 @@ var injector = require('../../index.js').injector;
 var _ = injector.get('_');
 
 module.exports = function create(fittingDef) {
+    injector = require('../../index.js').injector;
     var swag = injector.get('Http.Services.Swagger');
     return function swagger_render(context, next) {    // jshint ignore:line
         if (!context.response.headersSent) {
--- a/spec/lib/api/2.0/config-spec.js
+++ b/spec/lib/api/2.0/config-spec.js
@@ -6,22 +6,22 @@
 describe('Http.Api.Config v2.0', function () {
     var configuration;
     before('start HTTP server', function () {
-        this.timeout(5000);
-        return helper.startServer();
-    });
-
-    beforeEach('set up mocks', function () {
-        configuration = helper.injector.get('Services.Configuration');
-        sinon.stub(configuration, 'set').returns(configuration);
-        sinon.stub(configuration, 'getAll').returns({});
+        this.timeout(10000);
+        return helper.startServer().then(function() {
+            configuration = helper.injector.get('Services.Configuration');
+            sinon.stub(configuration, 'set').returns(configuration);
+            sinon.stub(configuration, 'getAll').returns({});
+        });
     });
 
     afterEach('tear down mocks', function () {
-        configuration.set.restore();
-        configuration.getAll.restore();
+        configuration.set.reset();
+        configuration.getAll.reset();
     });
 
     after('stop HTTP server', function () {
+        configuration.set.restore();
+        configuration.getAll.restore();
         return helper.stopServer();
     });
 
--- a/spec/lib/api/2.0/tasks-spec.js
+++ b/spec/lib/api/2.0/tasks-spec.js
@@ -37,8 +37,26 @@ describe('Http.Api.Tasks', function () {
           });
     });
 
+    afterEach('restoring stubs', function() {
+        function resetStubs(obj) {
+            _(obj).methods().forEach(function (method) {
+                if (obj[method] && obj[method].reset) {
+                    obj[method].reset();
+                }
+            }).value();
+        }
+
+        resetStubs(taskProtocol.activeTaskExists);
+        resetStubs(taskProtocol.requestCommands);
+        resetStubs(taskProtocol.respondCommands);
+        resetStubs(tasksApiService.getNode);
+        resetStubs(lookupService.ipAddressToMacAddress);
+    });
+
     after('stop HTTP server', function () {
-        return helper.stopServer();
+        return helper.reset().then(function(){
+            return helper.stopServer();
+        });
     });
 
     describe('GET /tasks/:id', function () {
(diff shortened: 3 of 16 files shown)


BillyAbildgaard:
test this please

srinia6:
+1

BillyAbildgaard:
@yyscamper not sure if you want to take a look at this fix since it addresses github issue RackHD/RackHD#259 that you worked on

lib/fittings/swagger_render.js
```diff
@@ -6,6 +6,7 @@ var injector = require('../../index.js').injector;
 var _ = injector.get('_');
 
 module.exports = function create(fittingDef) {
+    injector = require('../../index.js').injector;
```
yyscamper: line #5 has the same code, what's the difference? Could you please explain more why this is the root case?

spec/lib/api/2.0/tasks-spec.js
```diff
@@ -37,8 +37,26 @@ describe('Http.Api.Tasks', function () {
...
+        resetStubs(taskProtocol.activeTaskExists);
+        resetStubs(taskProtocol.requestCommands);
+        resetStubs(taskProtocol.respondCommands);
+        resetStubs(tasksApiService.getNode);
+        resetStubs(lookupService.ipAddressToMacAddress);
+    });
+
     after('stop HTTP server', function () {
-        return helper.stopServer();
+        return helper.reset().then(function(){
```
yyscamper: why does this need call the `helper.reset()` but other unit-test files don't need?

yyscamper:
@BillyAbildgaard thanks for your fix. Could you please clarify my above questions? and also the following question:
Do you have run the unit-test files in random order with enough times?

BillyAbildgaard:
@yyscamper I ran the tests in general and ran the specific case you saw where running workflows before profiles caused the failure.  The fix eliminated this case.

This story was just to fix the bug.  Stryker team has an additional story in our backlog to look into running the unit tests randomly

lib/fittings/swagger_render.js
BillyAbildgaard: The difference is in workflows we are using simpleWrapper to mock an injectable and what's happening is the swagger_render takes this injectable and exports the function with it.  The Profiles test then continues to use this "stale" injectable.
By adding in this line we are essentially refreshing the injectable to make sure if there were any changes the exported function takes them into account.

@zyoung51 can correct me if my understanding is wrong or if he wants to add more :)

spec/lib/api/2.0/tasks-spec.js
BillyAbildgaard: @yyscamper I added this because I saw the database was being cleared and then injecting Views so I just wanted to make sure that no stale Views were being carried over from this test.  Some tests don't use Views so that's probably why you don't see this in all of the tests

lib/fittings/swagger_render.js
yyscamper: But I know the require has cache (see https://nodejs.org/api/modules.html#modules_caching), so the two require operation should return the same object, so I think this line will not refresh the injector

spec/lib/api/2.0/tasks-spec.js
yyscamper: thx

yyscamper:
@BillyAbildgaard : The whole solution of random unit-test will need relative much consideration, but if you want to try, so can simplify use `sort -R`, the full command is:
```
./node_modules/.bin/mocha $(find spec -name '*-spec.js' | sort -R ) -R spec --require spec/helper.js
```

Could you please try above random testing with enough times? Because I think that issue should not be considered as fully resolved if the random testing fails.

BillyAbildgaard:
@yyscamper I can confirm the fix does resolve the bug that was reported.  I tested it multiple times.  

We have a story for randomized testing, but that was outside the scope of what I was trying to do for this, which was just to resolve the bug reported in RackHD/RackHD#259.

lib/fittings/swagger_render.js
brianparry: @yyscamper This change is a work-around for an issue with the way mocking is being done.  Without this line, the http server will run with an old version of the injector that still has mocks from the previous test.  We have work planned to make structural changes to all of the unit tests to avoid the need for this work-around in the future.

brianparry:
@yyscamper As @BillyAbildgaard has mentioned, this PR is a general cleanup of unit test mocking and a fix for RackHD/RackHD#259.  Full random-order testing support is planned as part of a future story and is outside the scope of this PR.

spec/lib/api/2.0/config-spec.js
```diff
@@ -6,22 +6,22 @@
...
     before('start HTTP server', function () {
-        this.timeout(5000);
-        return helper.startServer();
-    });
-
-    beforeEach('set up mocks', function () {
-        configuration = helper.injector.get('Services.Configuration');
-        sinon.stub(configuration, 'set').returns(configuration);
-        sinon.stub(configuration, 'getAll').returns({});
+        this.timeout(10000);
```
brianparry: Does this test fail consistently without the longer timeout?

BillyAbildgaard: @brianparry not that I saw, but I figured I would just increase the timeout since I was in there to decrease the likelihood of a failure because of a timeout.

lib/fittings/swagger_render.js
yyscamper: @brianparry : thanks for your explanation. I just wondering how this line of code works, because of the require caching, normally you should not get a new version of injector. But depends on your testing, this should work. Maybe some of my understanding is not correct.

yyscamper:
@BillyAbildgaard @brianparry : I agree on the scope for this PR. So if there is any more unit-test error if run in random order, we can fix it then.

lib/fittings/swagger_render.js
brianparry: @yyscamper I'm not entirely sure why this works.  For some reason there is a difference in the injector object between initially requiring the fitting and loading the fitting.  Given that this is a temporary work-around for a test-only problem, I think it's acceptable for now.

PR #347, opened 2016-06-29
install-os: fix template defect when a value is not existing; change hostname/domain to be optional

yyscamper:
When task-schema is enabled, if an parameter is optional it will not existing in the rendering context, (previously the optional value will be set to null or []). But current template rendering will throw error if the parameter is missing, currently ,we use following syntax to check a parameter is not missing:
```
if (param !== undefined) {
 ...
}
```
This will cause the ejs error with failing to find the `param`.
The correct syntax should be:
```
if (typeof param !== 'undefined') {
 ...
}
```

Meanwhile, I have updated all install-os template to let the `domain`/`hostname` to be optional. This is a requirement from this PR (https://github.com/RackHD/on-http/pull/342), that PR only change on ESXi, my PR changes all install-os's template.

Depends on: https://github.com/RackHD/on-tasks/pull/257

**NOTE**: merge together with https://github.com/RackHD/on-tasks/pull/257, otherwise various OS installation may be broken down (when some optional parameters are missing).

@RackHD/corecommitters @WangWinson @iceiilin @cgx027 @pengz1

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -26,18 +26,20 @@ authconfig --enableshadow --passalgo=sha512 --enablefingerprint
 rootpw --iscrypted <%=rootEncryptedPassword%>
 
 #create all users
+<% if (typeof users !== 'undefined') { %>
 <% users.forEach(function(user) { %>
 <%_  if( typeof user.uid !== 'undefined' ) { _%>
         user --name=<%=user.name%> --uid=<%=user.uid%> --iscrypted --password <%=user.encryptedPassword%>
 <%_  } else { _%>
         user --name=<%=user.name%>  --iscrypted --password <%=user.encryptedPassword%>
 <%_ }}) _%>
+<% } %>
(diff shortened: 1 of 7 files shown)


iceiilin:
+1

anhou:
:+1:

WangWinson:
A very disgusting case: what if the option value set to `null` ...
```
> typeof null
'object'
```

yyscamper:
@WangWinson  if the value is `null`, the json-schema will catch it and also if it is null, the ejs rendering will also fail.

Actually, the template doesn't cover all kinds of  type checking, for example, the `domain` is string, we can use alternative `typeof domain === 'string'`, but the type checking has been covered by json-schema. So to make the template logic simple, I let the template only judge the existing of parameter.

WangWinson:
OK, make sense
:+1:

benbp:
👍

PR #348, opened 2016-06-29
Refactor nodes/obm 1.1 and 2.0 route to use new OBM model

srinia6:
Part of OBMs refactoring effort. Dependents on on-core PR RackHD/on-core#164 changes made by Brian. @brianparry @dalebremner @BillyAbildgaard @zyoung51 @RackHD/corecommitters 

Build will fail due to on-core dependency. Unit-test pass locally with all dependencies.


srinia6:
CI error was due to removal of obmSettings from 2.0 views, and unit-test had old mock-up data. Fixed that.

data/views/node.2.0.json
```diff
@@ -1,4 +1,13 @@
 {
+<%
+    var _ = injector.get('_');
```
brianparry: @srinia6 I know that this is done in other places as well, so it is probably outside the scope of this PR, but we should change the template render context to just include the injectables we actually need (lodash in this case) rather than the whole injector.  This will prevent a malicious user from posting a view and getting full access to all modules.

data/views/node.2.0.json
```diff
@@ -27,4 +43,4 @@
 <% } %>
 <% } %>
 "updatedAt": "<%=updatedAt%>"
```
brianparry: @srinia6 Could you remove createdAt and updatedAt?

lib/api/1.1/northbound/nodes.js
```diff
@@ -44,10 +71,15 @@ function nodesRouterFactory (
...
     router.get('/nodes', rest(function (req) {
-        return nodeApiService.getAllNodes(req.query);
-    }, {
-        serializer: 'Serializables.V1.Node',
-        isArray: true
+        return nodeApiService.getAllNodes(req.query)
+        .then(function (nodes){
+            return Promise.map(nodes, function(node) {
+                return _renderNodeObmSettings(node);
+                })
```
brianparry: minor indentation issue.

lib/api/1.1/northbound/nodes.js
```diff
@@ -44,10 +71,15 @@ function nodesRouterFactory (
...
+            .then(function (modifiedNodes){
```
brianparry: @srinia6 is this `.then` block necessary?  It looks like it just returns the fulfillment value from the previous chained promise.

lib/api/1.1/northbound/nodes.js
```diff
@@ -111,12 +142,15 @@ function nodesRouterFactory (
      */
 
     router.get('/nodes/:identifier', rest(function (req) {
-        return nodeApiService.getNodeById(req.params.identifier);
-    }, {
-        serializer: 'Serializables.V1.Node',
+        return nodeApiService.getNodeById(req.params.identifier)
+        .then(function (node){
+            return _renderNodeObmSettings(node)
+            .then(function (modifiedNodes){
```
brianparry: @srinia6 Same comment as above: it looks like this `.then` is not needed.

lib/api/2.0/nodes.js
```diff
@@ -8,9 +8,10 @@ var nodes = injector.get('Http.Services.Api.Nodes');
 var workflowApiService = injector.get('Http.Services.Api.Workflows');
```
brianparry: @srinia6 Did you mean to include this file?  Does @dalebremner's PR supersede these changes.

data/views/node.2.0.json
srinia6: sure

lib/api/1.1/northbound/nodes.js
srinia6: @brianparry Hmm, not sure why i added it... Will change it.

lib/api/1.1/northbound/nodes.js
srinia6: Will make changes

lib/api/2.0/nodes.js
srinia6: Yes, this is part of my PR. Not dale's.

data/views/node.2.0.json
srinia6: ok, makes sense.

yyscamper: Actually, now the template file lacks the unit-test, so I don't suggest putting too much logic in the template file. It's more difficult to reading the template code than code itself.  So adding either `injector` or lodash into template is not good for me.

data/views/node.2.0.json
```diff
@@ -13,18 +21,24 @@
 <% } else  {%>
 "tags": [],
 <% } %>
-<% if (hasOwnProperty("obmSettings")) { %>
-"obmSettings": <%- JSON.stringify(obmSettings) %>,
-<% } else  {%>
-"obmSettings": [],
+<% if (obms == null) { %>
```
yyscamper: whether there is a case the `obms` is not in the rendering context? If so, this line will fail. Actually I just fixed a lot such kind of issue for install-os auto-install scripts, see https://github.com/RackHD/on-http/pull/347

data/views/node.2.0.json
```diff
@@ -13,18 +21,24 @@
...
+{
+"obm":"<%=basepath %>/obms/<%=ids%>"
+}
+<%= ( arr.length > 0 && i < arr.length-1 ) ? ',': '' %>
+<% }); %>
+],
 <% } %>
-"type": "<%=type%>",
 <% if (hasOwnProperty("sku")) { %>
 <% if (sku === null) { %>
```
yyscamper: Also for this line, if the `sku` is not existing in the rendering context, this will fail.

lib/api/1.1/northbound/nodes.js
```diff
@@ -28,13 +30,38 @@ function nodesRouterFactory (
...
+    _,
+    assert
 ) {
     var router = express.Router();
 
     /**
+     * Renders obmSettings property into a node.
+     *
+     * @param {object} node - node model instance
+     * @param {Array} obms - array of obm model instances
```
yyscamper: The function definition doesn't have this param.

data/views/node.2.0.json
srinia6: @yyscamper: These are views, and we don't have unit-test for them. And, yes, @brianparry indicated that we could remove injector code here. We can add a story to our backlog to revist this file. The intent was to add a ref to obms model for a particular node. I can add a story to ice-box to re-visit this file for clean-ups.

data/views/node.2.0.json
srinia6: @yyscamper This piece of code was always present...i did not change it. Also, if the sku is null, it will display an empty array.

data/views/node.2.0.json
srinia6: @yyscamper We are checking if the obms are not present on line 24.

lib/api/1.1/northbound/nodes.js
srinia6: @yyscamper  yes, turns out that we did not need it. I can remove it.

PR #349, opened 2016-06-29
Implement BootStrap Status Infinite Retries 

DavidjohnBlodgett:
This PR addresses a case were, if we are performing tasks such as flashing BIOS, we would  currently stop attempting to report a nodes success to a server (in the event that the server is down for some reason), after a set amount of retries. In such an event, we could have a disconnection between current and expected state on that node. This implements a fix that will have the node attempt to report its success indefinitely, until the server responds.

@RackHD/corecommitters

### DIFF
--- a/data/templates/bootstrap.js
+++ b/data/templates/bootstrap.js
@@ -11,8 +11,7 @@ var http = require('http'),
     port = '<%=port%>',
     tasksPath = '/api/current/tasks/<%=identifier%>',
     MAX_BUFFER = 1000 * 1024,
-    RETRIES = 5;
-
+    MAX_RETRY_TIMEOUT = 60 * 1000
 /**
  * Synchronous each loop from caolan/async.
  * @private
@@ -54,9 +53,9 @@ function eachSeries(arr, iterator, callback) {
  * @private
  * @param data
  * @param timeout
- * @param retries
  */
-function updateTasks(data, timeout, retries) {
+function updateTasks(data, timeout, retry, retries) {
+
     var request = http.request({
         hostname: server,
         port: port,
@@ -79,33 +78,22 @@ function updateTasks(data, timeout, retries) {
                 }, timeout);
             } else {
                 console.log("Task Execution Complete");
-
                 process.exit(data.exit.code || 0);
             }
         });
     }).on('error', function (err) {
-        console.log("Update Tasks Error: " + err);
-
-        if (retries === undefined) {
-            retries = 1;
-        } else {
-            retries = retries + 1;
-        }
-
-        if (retries < RETRIES) {
+            console.log("Update Tasks Error: " + err);
+            if (retries === undefined){
+                retries = 1;
+            }else {
+                retries = retries + 1;
+            }
             console.log("Retrying Update Tasks Attempt #" + retries);
 
             setTimeout(function () {
-                updateTasks(data, timeout, retries);
-            }, timeout * retries);
-        } else {
-            console.log("Update Tasks retries completed, getting new tasks.");
-
-            setTimeout(function () {
-                getTasks(timeout);
-            }, timeout);
-        }
-    });
+                updateTasks(data, timeout, retry, retries);
+            }, Math.min(timeout * retries, MAX_RETRY_TIMEOUT));
+        });
 
     // Call error.toString() on certain errors so when it is JSON.stringified
     // it doesn't end up as '{}' before we send it back to the server.


data/templates/bootstrap.js
```diff
@@ -54,9 +53,13 @@ function eachSeries(arr, iterator, callback) {
  * @private
  * @param data
  * @param timeout
- * @param retries
  */
-function updateTasks(data, timeout, retries) {
+function updateTasks(data, timeout) {
+    // value to check succes case.
```
benbp: These comments are redundant, it's pretty clear from the code what is happening already.

benbp:
👍

benbp:
Would be nice if you could change the commit string to be a little more descriptive.

data/templates/bootstrap.js
DavidjohnBlodgett: understood.

DavidjohnBlodgett:
@benbp oh ok, sorry about that I'll fix it up post haste.

benbp:
@DavidjohnBlodgett no worries! I'm just being picky :)

DavidjohnBlodgett:
test this please

tldavies:
test this please

tldavies:
👍 when jenkins passes

tldavies:
test this please

tldavies:
test this please

yyscamper:
:+1:

brianparry:
test this please

DavidjohnBlodgett:
@jlongever Hey Joel, I'm sorry to have to bug you... but I'm hoping I could tap your knowledge to assist in finding the root cause behind Jenkins timing out, I wouldn't put it beyond the realm of possibility that me introducing an action that no longer times out or has a max retry could mess with something, but it isn't jumping out to me that a smoke test is failing or anything... Just wanted a 3rd set of eyes on it, if I could... thanks!

brianparry:
@DavidjohnBlodgett We've had some issues with our CI system over the past few days.  The timeouts on this PR are probably not caused by your change.  Jenkins, test this please

DavidjohnBlodgett:
@brianparry oh well thats comforting in a way lol... understood, I'll stand by until more info is available.

brianparry:
@DavidjohnBlodgett This looks like a real failure.  I think the discovery graph is getting stuck.

DavidjohnBlodgett:
@brianparry @jlongever @benbp Hi guys, so I've been trying to wrap my brain around what's the issue here, and can you guys help with some insight to what is happening here? https://github.com/RackHD/RackHD/blob/master/test/tests/api/v1_1/nodes_tests.py#L120

It feels to me like this code maybe actually waiting for the "else" case found here: https://github.com/RackHD/on-http/pull/349/files#diff-fa828f1c5e9bca4ba39e54d7c31b1c2bR98

does that seem to jive with what you see? I am assuming this is failing our tests because we are expecting the update tasks function to only retry a limited amount of time, then bail out to pickup the next task... and this code would keep it attempting to respond indefinitely.

Just wanted to see if my logic is sound...

jlongever:
@DavidjohnBlodgett That discovery test spawns a thread that listens on AMQP for a `graph.finished` event for each node that completed discovery graph.   The `wait_for_compeletion` is waiting for each worker thread to complete,  it will timeout after 20 minutes.   I think your assumption is correct,  the discovery graph never triggered a `graph.finished` event because discovery never completed?  Looking at the log here http://rackhdci.lss.emc.com/job/on-http/1747/artifact/build/vagrant.log I see `Graph.Discovery` task starting,  but doesn't seem like it runs through all of the catalog tasks.

tldavies:
@jlongever @DavidjohnBlodgett So what would be a good way to change the test? As it used to complete cause it would timeout after 5 tries and move on the next step. Now we want infinity until we get a response from the monorail server. Which it seems is why it is timing out as it still wanting for the response?

DavidjohnBlodgett:
@RackHD/corecommitters bug fixed, commit squashed. This should be good to go.

tldavies:
test this please

data/templates/bootstrap.js
```diff
@@ -79,33 +78,22 @@ function updateTasks(data, timeout, retries) {
...
+                retries = 1;
+            }else {
+                retries = retries + 1;
+            }
             console.log("Retrying Update Tasks Attempt #" + retries);
 
             setTimeout(function () {
-                updateTasks(data, timeout, retries);
+                updateTasks(data, timeout, retry, retries);
             }, timeout * retries);
```
zyoung51: @DavidjohnBlodgett Now that you are allowing for an infinite number of retries should you still scale this up by the number of retries without some kind of upper limit?

DavidjohnBlodgett: @zyoung51 I agree with your suggestion, and have added a ceiling of about 5mins. commit just posted.

tldavies:
test this please

DavidjohnBlodgett:
@RackHD/corecommitters will need to squash again... but just looking for thumbs up~ I believe I have addressed everyone's concerns.

DavidjohnBlodgett:
test this please

PR #350, opened 2016-06-29
POC Dockerized Discovery image in RACKHD

keedya:
@RackHD/corecommitters 

depends on : https://github.com/RackHD/on-taskgraph/pull/135
depends on : https://github.com/RackHD/on-tasks/pull/258


data/profiles/linux.ipxe
```diff
@@ -1,4 +1,4 @@
 kernel <%=kernelUri%>
 initrd <%=initrdUri%>
-imgargs <%=kernelFile%> initrd=<%=initrdFile%> auto=true SYSLOGSERVER=<%=server%> API_CB=<%=server%>:<%=port%> BASEFS=<%=basefs%> OVERLAYFS=<%=overlayfs%> BOOTIF=01-<%=macaddress%> console=tty0 console=<%=comport%>,115200n8 <%=kargs%>
+imgargs <%=kernelFile%> initrd=<%=initrdFile%> console=tty0 rancher.password=root rancher.cloud_init.datasources=['url:http://<%=server%>:<%=port%>/api/common/templates/cloud-config.yaml']
```
benbp: I'd prefer to actually make a new ipxe profile, maybe rancher.ipxe. We have a bunch of people using linux.ipxe for other custom tasks that we don't want to break functionality for.

data/templates/cloud-config.yaml
```diff
@@ -0,0 +1,35 @@
+#cloud-config
+write_files:
+  - path: /etc/rc.local
+    permissions: "0755"
+    owner: root
+    content: |
+      #!/bin/bash
+      modprobe ipmi_devintf
+      wget -O /tmp/micro.tar.xz <%= api.server %>/micro.tar.xz
```
benbp: Can we make the image name (or entire download path) a template? That way we can change it on the fly in the workflow.

data/profiles/linux.ipxe
keedya: @benbp agreed. i ll create a separate profile  for Rancher

data/templates/cloud-config.yaml
```diff
@@ -0,0 +1,35 @@
+#cloud-config
+write_files:
+  - path: /etc/rc.local
+    permissions: "0755"
+    owner: root
+    content: |
+      #!/bin/bash
+      modprobe ipmi_devintf
```
benbp: Things will still work even if this fails, I assume?

data/templates/cloud-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+      wget -O /tmp/micro.tar.xz <%= api.server %>/micro.tar.xz
+      while [ $(docker images | grep -c micro) == "0" ]; do
+         xz -cd /tmp/micro.tar.xz | docker load
+      done
+      docker run -it -d \
```
benbp: As I mentioned earlier, would be awesome to do this in a loop so we can change what containers are downloaded and run dynamically by the workflow, but seems more appropriate for follow-on work than necessary for this PR.

data/templates/cloud-config.yaml
keedya: trying to start ipmi driver. ipmi cataloging will fail if driver doesn't exist or fail to start.

benbp: I just mean the for the rest of the content script, obviously bmc cataloging will fail as you say.

keedya:
closing POC and open a new request:
https://github.com/RackHD/on-http/pull/392

PR #352, opened 2016-06-30
Fix sata disk identifier SN

anhou:
when install ESXi,  lodash ‘_’  may be before or after disk SN,  like 'XXXXXX_____________' or '_____________YYYYYY' so the esxiWwid may be 
"esxiWwid":"t10.ATA_____QEMU_HARDDISK_XXXXXX_____________" or
"esxiWwid":"t10.ATA_____QEMU_HARDDISK______________YYYYYY"  it's decided by the raw data read by 'hdparm', previously, lodash ’_‘ is fixed before 'SN' in codes. 
this issue is found when installing ESXi6.0 in InfraSIM. with this fix, ESX6.0 could be installed with InfraSIM.

depends on: https://github.com/RackHD/on-imagebuilder/pull/45

@RackHD/corecommitters @RackHD/rackhd_dev

### DIFF
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -15,11 +15,30 @@ var options = {
 
 /**
  * Get SATA drive's SN string.
- * @param {String} SATA drive's name, eg. 'sda'
+ * @param {String} sataDrive SATA drive's name, eg. 'sda'
  * @return {String} SATA drive's SN string with 20 characters like 'QM00007_____________'
  */
 function getSataSnStr(sataDrive) {
     var output = execSync(cmdSataRawInfo + ' /dev/' + sataDrive);
+    /*  output data like below:
+     *  $ sudo hdparm --Istdout /dev/sda
+     *  dev/sda:
+     *
+     *  0040 3fff 0000 0010 7e00 0200 003f 0000
+     *  0000 0000 514d 3030 3030 3520 2020 2020
+     *  2020 2020 2020 2020 0003 0200 0004 322e
+     *  322e 3120 2020 5145 4d55 2048 4152 4444
+     *  4953 4b20 2020 2020 2020 2020 2020 2020
+     *  2020 2020 2020 2020 2020 2020 2020 8010
+     *  .....
+     *
+     *  snHexStr is from 21 to 40 bytes, total 20 bytes, shown as below
+     *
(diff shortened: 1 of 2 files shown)


data/templates/get_driveid.js
```diff
@@ -3,15 +3,45 @@
...
 var cmdVdInfo = 'ls -l /dev/disk/by-path';
 var cmdScsiId = 'lsscsi';
+var cmdSataRawInfo = 'sudo hdparm --Istdout';
 var options = {
     timeout: 10000 //10 seconds
 };
 
 /**
+ * Get SATA drive's SN string.
+ * @param {String} SATA drive's name, eg. 'sda'
```
WangWinson: the param name need to match, here is `sataDrive` ?

data/templates/get_driveid.js
```diff
@@ -3,15 +3,45 @@
...
+ * @return {String} SATA drive's SN string with 20 characters like 'QM00007_____________'
+ */
+function getSataSnStr(sataDrive) {
+    var output = execSync(cmdSataRawInfo + ' /dev/' + sataDrive);
+    var lines = output.toString().split('\n');
+    var snHexStr = lines.reduce(function (result,line) {
```
WangWinson: maybe add comments about the example data will be helpful.

yyscamper:
:+1:

data/templates/get_driveid.js
```diff
@@ -3,15 +3,64 @@
...
+     *  snHexStr is from 21 to 40 bytes, total 20 bytes, shown as below
+     *
+     *  .....
+     *            514d 3030 3030 3520 2020 2020
+     *  2020 2020 2020 2020
+     *  .....
+     */
+    var lines = output.toString().split('\n');
+    var snHexStr = lines.reduce(function (result,line) {
+        if(line.length === 39) {
```
WangWinson: I think the logic can work here, 
but what if some corner case make some line length equal 39 coincidentally. ? Maybe check the line format is hex will be more safe ?

WangWinson:
still :+1:

data/templates/get_driveid.js
anhou: Great, add more accurate judgement

pengz1:
+1

PR #353, opened 2016-07-01
fix issue#240: POST invalid workflow API will always triggers Graph.BootLiveCD

yyscamper:
resolves https://github.com/RackHD/RackHD/issues/240

Throw error if the graph name is missing or in wrong format.

@RackHD/corecommitters @iceiilin @WangWinson @cgx027 @pengz1

### DIFF
--- a/lib/services/workflow-api-service.js
+++ b/lib/services/workflow-api-service.js
@@ -42,13 +42,17 @@ function workflowApiServiceFactory(
 
     WorkflowApiService.prototype.createAndRunGraph = function(configuration, nodeId) {
         var self = this;
-        return Promise.resolve()
+        return Promise.try(function() {
+            if (!configuration.name || !_.isString(configuration.name)) {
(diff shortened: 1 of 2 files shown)


anhou:
:+1:

lib/services/workflow-api-service.js
```diff
@@ -42,13 +42,17 @@ function workflowApiServiceFactory(
 
     WorkflowApiService.prototype.createAndRunGraph = function(configuration, nodeId) {
         var self = this;
-        return Promise.resolve()
+        return Promise.try(function() {
+            if (!configuration.name || !_.isString(configuration.name)) {
+                throw new Errors.BadRequestError('Graph name is missing or in wrong format');
```
anhou: maybe it could be 422.  not 400, https://httpstatuses.com/422 

refer to https://developer.github.com/v3/  'Sending invalid fields will result in a 422 Unprocessable Entity response.'

benbp:
👍

PR #354, opened 2016-07-06
Fix POST /api/1.1/nodes/{identifier}/obm

brianparry:
* Allow posting OBM object or array of OBM objects
* Populate response body with obmSettings property

This problem was discovered by @uppalk1 
@RackHD/corecommitters @srinia6 @dalebremner @keedya @BillyAbildgaard


keedya:
@uppalk1 @tannoa2

benbp:
:+1:

PR #356, opened 2016-07-07
Add UUID prefix to SSDP UDN

jlongever:
This is per upnp/ssdp specification,  node-ssdp library doesn't construct the UDN as expected.

@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2 @keedya @king-jam


benbp:
👍

PR #358, opened 2016-07-08
Fix ODR787: defineTaskGraph only validate definition rather than creating whole graph instance

yyscamper:
This PR is to fix [ODR#787](https://hwjiraprd01.corp.emc.com/browse/ODR-787).

Previously the `defineTaskGraph` leverage creating graph instance to validate whether the graph definition is valid, but some required option may not be ready at  that time (for example when insert a SKU pack which contains some SKU specific graph). The change is to use the new API to validate the graph definition only, it will skip the task rendering and schema validation.

@RackHD/corecommitters @zyoung51 @iceiilin @pengz1 

Depends on: https://github.com/RackHD/on-tasks/pull/273

### DIFF
--- a/lib/services/workflow-api-service.js
+++ b/lib/services/workflow-api-service.js
@@ -129,6 +129,11 @@ function workflowApiServiceFactory(
             });
         })
         .catch(function(error) {
+            logger.error('createGraph fails', {
+                definition: definition,
+                options: options,
+                error: error
+            });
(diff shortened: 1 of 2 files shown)


iceiilin:
:+1:

anhou:
test this please

anhou:
:+1:

PR #359, opened 2016-07-09
Consistency for ways to use 2.0 API (common / current)

keedya:
- Able to switch on / by default in one place

### DIFF
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -12,6 +12,7 @@ var fs = require('fs');
 var proxyMiddleware = require('http-proxy-middleware');
 var path = require('path');
 var bodyParser = require('body-parser');
+var rewriter = require('express-urlrewrite');
 
 /**
  * Get remote address of the client.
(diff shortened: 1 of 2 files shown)


keedya:
@zyoung51 @brianparry

lib/services/http-service.js
```diff
@@ -218,6 +219,11 @@ function httpServiceFactory(
             )
         );
 
+        //re-route common and current
+        var varsionPath = configuration.get('versionBase') || '1.1';
```
zyoung51: versionPath instead of varsionPath?

zyoung51:
Just a nitpick on the variable name. LGTM, +1

lib/services/http-service.js
keedya: Oh wow, totally missed that!! Thanks @zyoung51

keedya:
@zyoung51 all set

yyscamper:
:+1:

lib/services/http-service.js
```diff
@@ -218,6 +219,11 @@ function httpServiceFactory(
             )
         );
 
+        //re-route common and current
+        var versionPath = configuration.get('versionBase') || '1.1';
```
brianparry: @keedya I think we should add an assert or some other type of check to verify that `versionBase` is set to a valid value.

brianparry:
:+1: One minor comment related to config validation, but I think that's more of a general problem outside the scope of this PR.

brianparry:
test this please

PR #361, opened 2016-07-11
Add a wait time for the boilerplate ipxe to allow the links to come up

miramm:
This is a fix for ODR-786 "Onrack is failing to discover high node count"
For convenience i pasted the description of this ODR:
https://hwjiraprd01.corp.emc.com/browse/ODR-786

"When powering up 18 physical nodes only 5 of those nodes got discovered, 1 of those nodes got fully discovered but 4 of them were not and displayed the error message below. The problem seems to be around the discovery workflow, Onrack assigned an ID for the discovered node but it failed along the process causing major issues with the rest of the nodes trying to be discover. This is an intermittent issue as I can discover all the nodes too.
Below is the error message from one of the discovered nodes:
https://10.241.197.32/redfish/v1/Chassis/577bb2834f428e6e0b4b285f
Response Body
{ "status": 404, "message": "get not found (Could not find node with identifier 577bb2834f428e6e0b4b285f)" }
Debugging so far:
While having a clear database and checked that all onrack services are running, I have brought up each of the node up separately and make sure that it gets discovered properly prior to power up the next node. This has been succesfull.

While having a clear database and checked that all onrack services are running, brought up two to four nodes at a time and monitor the discovery process. sometime works.
While having a clear database and checked that all onrack services are running, brought up all the 18 nodes at a time. 3 out of 4 times it fails.
Same issue has been seen with a virtual environment.
Results URL: https://web.hwimo.lab.emc.com/qa/log/bug_logs/ODR-786"

### DIFF
--- a/data/profiles/boilerplate.ipxe
+++ b/data/profiles/boilerplate.ipxe
@@ -32,6 +32,8 @@ iseq ${interface} ${net${ifIndex}/mac} && goto ifBoot || inc ifIndex && goto ifF
 :ifBoot
 iseq ${ifBootAttempt} ${ifBootAttemptMax} && goto ifBootFailure || goto ifBootContinue
 :ifBootContinue
+echo "sleeping to allow link up"
+sleep 10
 ifopen net${ifIndex}
 ifconf net${ifIndex} && goto ifConfigured || inc ifBootAttempt && goto ifBoot
 


miramm:
@RackHD/corecommitters @keedya

jlongever:
jenkins, test this please

keedya:
@miramm  my change was in on-imagebuilder:
https://github.com/RackHD/on-imagebuilder/commit/fc8aed20ac94c5cc2f370ef9182d2fc060965f32

keedya:
:+1:

amymullins:
@miramm similar phenomena was seen during some MFG poc efforts in the past 1-2 weeks.  did a 10 second sleep consistently discover/boot 18 nodes at a time?  The rack config we were experimenting with has 32 physical servers and had to stagger the power on of single servers.

FYI @RackHD/corecommitters , can we spend some looking for root cause and adding some guidelines/criteria for staggering powering on nodes (ie, how many servers are expected to .power on at a time)

miramm:
@amymullins yes, we were consistently getting discovery failures with 18 physical nodes when coming up at a same time but it got worst as the number of nodes increases. We're moving to 100's nodes but with a minimal delay in between the power on of the nodes.

leoyjchang:
The interesting thing is Payne Wang from InfraSIM team deployed a 218 vNode test environment and boot them all up at same time, the partial discovery phenomenon like ODR-786 was not found.  We'd better to figure out reason as well.

brianparry:
@miramm Have you confirmed that this problem is caused by network induced delay?  The original problem that required this sleep was a lookup race condition.  We should be sure we're not just masking another race in RackHD.

If we find that the problem is caused by the network, I have a couple other concerns:
1. It seems dangerous to rely on empirical data in setting a fixed timeout like this.  Since we can't take measurements with every possible network configuration, we can't know how much margin we have with this particular timeout value.
2. With this approach we force all networks - slow and fast - to wait for a period of time long enough to cover the worst case.  This forces a typical system to be slower than necessary.

As an alternative, could we try adding retries at short intervals to https://github.com/RackHD/on-http/blob/3cd07f97173c09f9278cbbdb6a85126a3e1c970a/data/profiles/redirect.ipxe?  We could potentially also add a retry count and delay per attempt as configuration parameters so these values can be rendered into the profile and tuned by the user for their particular need.  This would allow long waits in instances where required by the network without forcing the delay into network topologies that are available more quickly.  See http://ipxe.org/scripting for an example using retries with chain loading.

benbp:
I agree with @brianparry here, we should avoid hardcoded sleeps as much as possible, as we can't reliably predict what the right timeout is for an environment, and it effectively becomes a question of what's the lowest common denominator we can use as a timeout for all environments, which ends up capping our maximum performance for all environments without the issue as well. It's much better to use an exponential backoff mechanism, or a retry mechanism, etc. 

We first attempted to solve network convergence issues by extending the DHCP timeouts in our ipxe image, which seemed to help for many cases, so I'd be curious if we can generate more data here about where in the network these issues are happening.

yyscamper:
@benbp submit a better solution for this issue (https://github.com/RackHD/on-http/pull/376), I think this PR should be closed.

brianparry:
I agree with @yyscamper that this PR should be closed.  @miramm do you have any objection?

PR #364, opened 2016-07-12
Fix Ubuntu installation's workflow

keedya:
- This PR gives the user the choice to install Ubuntu using mounted ISO image or public mirror
- Documentation for this changes will be traced in a different PR

depends on : https://github.com/RackHD/on-tasks/pull/280

@RackHD/corecommitters @jlongever @tannoa2 @uppalk1 @zyoung51

### DIFF
--- a/data/profiles/install-ubuntu.ipxe
+++ b/data/profiles/install-ubuntu.ipxe
@@ -1,6 +1,6 @@
 echo Starting Ubuntu x64 installer for ${hostidentifier}
-set base-url <%=repo%>/dists/<%=version%>/main/installer-amd64/current/images/netboot/ubuntu-installer/amd64
+set base-url <%=repo%>/<%=baseUrl%>
 kernel ${base-url}/linux
 initrd ${base-url}/initrd.gz
-imgargs linux auto=true url=<%=installScriptUri%> hostname=<%=hostname%> log_host=<%=server%> BOOTIF=01-<%=macaddress%> interface=auto console=<%=comport%>,115200n8 console=tty0
+imgargs linux auto=true url=<%=installScriptUri%> hostname=<%=hostname%> log_host=<%=server%> BOOTIF=01-<%=macaddress%> interface=<%=interface%> console=<%=comport%>,115200n8 console=tty0 <%=kargs%>
 boot || prompt --key 0x197e --timeout 2000 Press F12 to investigate || exit shell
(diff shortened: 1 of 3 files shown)


keedya:
depends on: https://github.com/RackHD/on-tasks/pull/278

leoyjchang:
Add @WangWinson

yyscamper:
@keedya : Could you please explain what's the problem you are trying to fix?

keedya:
@yyscamper: Ubuntu workflow didn't install and when I tried to debug it I found the following errors:
1- missing  squashfs image 
2- if nodes disk is been partitioned prior to the installation, the preceed fails
3- essential package were not available and curl renasar.pub was failing
4- create users script fails every time to set the right password 

Finally Isc dhcp server needs 2 options to be added for the Ubuntu installer to accept shop offer
Options are:
1) dns-server-option 
2) routers option

tannoa2:
+1

yyscamper:
@keedya : All you listed problems is critical and should decisively lead to installation failure. But I know both @WangWinson and @cgx027 had verified the ubuntu installation passed, I think they should have already known such big problem if it is true, so I want to understand what's the difference between your case and their.
@WangWinson , @cgx027 : What's your comment for this change?

keedya:
test this please

WangWinson:
In @keedya 's issue list above:
>
1- missing squashfs image 
2- if nodes disk is been partitioned prior to the installation, the preceed fails
3- essential package were not available and curl renasar.pub was failing
4- create users script fails every time to set the right password

- [ ] 1, if I remember correctly, the squashfs image was once removed from the preseed file in history PR. It works on my side when I test it.
- [x] 2, looks like I haven't cover that case in my dev test. So this could be an issue.
- [ ] 3, Not met that issue before.
- [x] 4, should be an issue.

>
Finally Isc dhcp server needs 2 options to be added for the Ubuntu installer to accept dhcp offer
Options are:
1) dns-server-option 
2) routers option

- [x] Yes, this is known issue and I didn't aware of that we packaged dhcpd.conf which could be updated. I think @anhou added this note in readdocs.

keedya:
@anhou are you planning on retesting Ubuntu or Do you agree that these changes are necessary?
Note: my node doesn't have internet access
Network configuration: 
- eth0 connected to mgmt vswitch(172.31.128.1)

keedya:
@WangWinson : squashfs  image can be optional. It is only required when installing Ubuntu from a mounted ISO I can modify the profile to make net image optional
For 3 : fail every time when installing Ubuntu from a mounted ISO

keedya:
@here I will submit more changes to allow user to choose between using mounted ISO or public repo

keedya:
test this please

WangWinson:
Thanks @keedya . It's clear now, install form Ubuntu from mounted ISO not supported previously.

WangWinson:
:+1:

PR #365, opened 2016-07-12
2.0 API error handling modifications

dalebremner:
@RackHD/corecommitters @BillyAbildgaard @keedya @zyoung51

### DIFF
--- a/lib/services/swagger-api-service.js
+++ b/lib/services/swagger-api-service.js
@@ -153,7 +153,7 @@ function swaggerFactory(
     }
 
     function swaggerRenderer(req, res, viewName, next) {
-        var status = req.swagger.options.success || 200;
+        var status = res.locals.errorStatus || req.swagger.options.success || 200;
         return Promise.try(function() {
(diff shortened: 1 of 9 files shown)


keedya:
looks good to me :+1:

lib/services/swagger-api-service.js
```diff
@@ -207,7 +210,12 @@ function swaggerFactory(
             res.status(status).send(data);
         })
         .catch(function(err) {
-            next(_processError(err));
+            if (res.locals.errorStatus) {
+                next();
```
brianparry: @dalebremner What are we handling off to from the error handler fitting?  I think the error handler should always close the request in some way, so I would think that calling next will complain that headers have already been sent.  Should we be doing something like calling `res.status(500).json(someDefaultError)`

brianparry: So it looks like the default swagger error handler fitting does the same thing: https://github.com/theganyo/swagger-node-runner/blob/master/fittings/json_error_handler.js, so ignore my comment.

brianparry:
:+1:

PR #362, opened 2016-07-12
Feature/photon os bootstrap

cgx027:
Photon OS bootstrap support

This is one of three PRs for Photon OS bootstrap:
RackHD/on-tasks #275  https://github.com/RackHD/on-tasks/pull/275
RackHD/on-taskgraph #138 https://github.com/RackHD/on-taskgraph/pull/138 
RackHD/on-http #362 https://github.com/RackHD/on-http/pull/362

@RackHD/corecommitters @iceiilin @pengz1 @WangWinson

PR #367, opened 2016-07-13
add photon os post install script

iceiilin:
The post-install-photon.sh script will be called in post install section of photon ks file, to configure bootstrap settings that cannot be covered in ks file.
Add hook in ks file to run the post -install-script.sh script.
Move photon boostrap file into a foler.

@yyscamper @WangWinson @cgx027


data/templates/post-install-photon.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+<%_ if (undefined !== networkDevices) { _%>
```
yyscamper: Use `typeof networkDevies !== 'undefined'` rather than yours as the networkDevices is optional value, it it doesn't exist, the ejs will throw error that it cannot find the reference of networkDevices.

data/templates/post-install-photon.sh
```diff
@@ -0,0 +1,92 @@
...
+chmod 644 /etc/systemd/network/<%=d.device%>-static.network
+
+        <%_ for (conf in d) { _%>
+            <%_ if (undefined === d[conf]) continue; _%>
+            <%_ if ('ipv4' === conf || 'ipv6' === conf) { _%>
+                <%_ ipconfig = d[conf] _%>
+                <%_ cidr = 0 _%>
+                <%_ maskNodes = ipconfig.netmask.match(/(\w+)/g) _%>
+                <%_ maskNodes.forEach(function(i) { _%>
+                    <%_ if ('ipv4' === conf) { _%>
```
yyscamper: A NIC can have both ipv4 and ipv6 address, so if both ipv4 & ipv6 is present, you need set two kinds of address to the NIC, but your code only choose the ipv4.

data/templates/post-install-photon.sh
```diff
@@ -0,0 +1,92 @@
...
+echo "DNS=<%=dns%>" >> /etc/systemd/resolved.conf
+    <%_ }) _%>
+systemctl restart systemd-resolved
+<%_ } _%>
+
+# Enable sshd
+sed -i 's/PermitRootLogin no/PermitRootLogin yes/g' /etc/ssh/sshd_config
+systemctl restart sshd
+
+<%_ if (undefined !== domain && undefined != hostname) { _%>
```
yyscamper: In other OSes, the `domain` and `hostname` are individual parameters, if only one of them is present is still a valid case, your change is to enforce both the `domain` and `hostname` should be present otherwise ignored. If this is your design intention, you'd better change the task-schema to reflect the limitation using the `dependancies` (see http://epoberezkin.github.io/ajv/keywords.html#dependencies)

yyscamper:
Where is the code to trigger this script?

data/templates/post-install-photon.sh
```diff
@@ -0,0 +1,92 @@
...
+# search domain in DNS cannot be set. Issue opened https://github.com/vmware/photon/issues/488
+line=`wc -l < /etc/hosts`
+sed -i "${line}i 127.0.0.1\t<%=hostname%>.<%=domain%>\t<%=hostname%>" /etc/hosts
+<%_ } _%>
+
+<%_ if (undefined !== users) { _%>
+# Set user info
+    <%_ users.forEach(function(user) { _%>
+        <%_ if (undefined != user.uid) { _%>
+useradd -m <%=user.name%> -p <%=user.encryptedPassword%> -u <%=user.uid%>
```
WangWinson: An issue just found, should quote the password ` -p '<%=user.encryptedPassword%>'`

WangWinson:
Looks like there are already several new templates for photon created in @cgx027 's [PR](https://github.com/RackHD/on-http/pull/362) 
So I suggest to create a /template/install-photon folder and put related template in.

data/templates/post-install-photon.sh
iceiilin: The address setting are the same here
https://github.com/RackHD/on-http/pull/367/files#diff-bc6f79dc614a4cfa1170c1f853e02b1cR32
The branch is just used to calculate cidr code for both ipv4 and ipv6

iceiilin:
@WangWinson yes I planned to create directory in @cgx027 's PR by merging into his branch.
@yyscamper I planned to add the hook in ks file in @cgx027 's PR

data/templates/post-install-photon.sh
iceiilin: I will modify it

data/templates/post-install-photon.sh
iceiilin: Will update it

data/templates/post-install-photon.sh
iceiilin: It is kind of special case in current Photon. Hostname is configureed in ks file, this script just deals with domain. I thought before that adding domain when localhost is not a valid case, but after searching, it is still possible. I will update the script.

yyscamper:
test this please

cgx027:
:+1: Looks great!

anhou:
:+1:

PR #373, opened 2016-07-14
List static files

lacarb:
Re-opening PR

Requires https://github.com/RackHD/on-core/pull/183
which adds constants to to constants.js

Introduces api call, '/files/static/list', which returns a list of static files to the user. Example output as follows...

  {
    "uri": "/common/vmlinuz-3.16.0-25-generic"
  },
  {
    "sku": "Cisco Nexus 3000 Switch - 54 port",
    "uri": "static/common/aTestFile.we"
  },

### DIFF
--- a/lib/services/static-files-api-service.js
+++ b/lib/services/static-files-api-service.js
@@ -1,0 +1,148 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+var fs = require('fs');
+var path = require('path');
+var walking = require('walk');
+
+module.exports = staticFilesApiServiceFactory;
+
+di.annotate(staticFilesApiServiceFactory, new di.Provide('Http.Services.Api.StaticFiles'));
+di.annotate(staticFilesApiServiceFactory,
+    new di.Inject(
+      'Constants',
+      'Logger',
+      'Promise',
+      'Services.Configuration',
+      'Services.Waterline'
+    )
+);
+
+
+function staticFilesApiServiceFactory(
+    constants,
+    Logger,
+    Promise,
+    configuration,
+    waterline
+) {
+    var logger = Logger.initialize(staticFilesApiServiceFactory);
+    var skuPath = constants.HttpStaticDir.skupack;
+    var staticPath = constants.HttpStaticDir.systemDefault;
+    var userPath = constants.HttpStaticDir.userDefault;
+    var walker;
+    var walk = Promise.promisifyAll(walking);
+    Promise.promisifyAll(fs);
+
+    function StaticFilesApiService() {
+    }
+
+
+    StaticFilesApiService.prototype.pairSkupackIds = function() {
+        return fs.readdirAsync(skuPath).filter(function(item) {
+            return (fs.lstatAsync(path.join(skuPath, item)))
+            .then(function(stat) {
+                return stat.isDirectory();
+            }); 
+        }).reduce(function(result, item) {
+            return Promise.try(function() {
+                return waterline.skus.findOne({id : item})
+                .then(function(sku) {
+                    result[item] = sku.name;
+                    return result;
+                });
+            })
+            .catch(function(e) {
+                logger.debug("Directory " + skuPath + "/" + item + " is not in the database",
+                    e);
+                return result;
+            });
+        },{}).then(function(pairs) {
+            return pairs;
+        });
+    };
+
+
+    StaticFilesApiService.prototype.walkDirectory = function(dir) {
+        return new Promise(function(resolve, reject) {
+            var walkResults = [];
+            if (typeof dir !== 'string') {
+                console.error("[ERROR] Object of type " + typeof dir + 
+                    " was passed to walkDirectory where a string was expected.");
+                resolve(walkResults);
+            }
+            walker = walk.walk(dir, {followLinks: false});
+
+            walker.on("file", function(root, fileStats, next) {
+                Promise.resolve().then(function() {
+                    var walkObj = {};
+                    if (root.indexOf(userPath) > -1){
+                        root = root.replace(userPath, '');
+                        walkObj.uri = path.join(root, fileStats.name);
+                        walkResults.push(walkObj);
+                    } else if (root.indexOf(staticPath) > -1){
+                        var longPath = path.join(root, fileStats.name).split("/");
+                        var httpIndex = longPath.indexOf("http") + 1;
+                        walkObj.uri = longPath.slice(httpIndex).join("/");
+                        walkResults.push(walkObj);
+                    } else if ((root.indexOf(skuPath) > -1) && 
+                        (root.indexOf("static") > -1)){
+                        var filePath = path.join(root, fileStats.name).split("/");
+                        var skuId = filePath[1];
+                        var skuUri = filePath.slice(3).join("/");
+                        walkObj.uri = skuUri;
+                        walkObj.sku = skuId;
+                        walkResults.push(walkObj);
+                    }
+                }).then(function() {
+                    next();
+                }).catch(function(e) {
+                    reject(e);
+                });
+            });
+
+            walker.on("errors", function(root, nodeStatsArray, next) {
+                nodeStatsArray.forEach(function (n) {
+                    console.error("[ERROR] " + n.name);
+                    console.error(n.error.message || (n.error.code + ": " + n.error.path));
+                });
+                next();
+            });
+
+            walker.on("end", function() {
+                resolve(walkResults);
+            });
+        });
+    };
+
+
+    StaticFilesApiService.prototype.getAllStaticFiles = function() {
+
+        var self = this;
+        var directories = [staticPath,
+                           configuration.get('httpStaticRoot', userPath),
+                           skuPath];
+
+        return self.pairSkupackIds(directories).then(function(skuInfo) {       
+            return [Promise.map(directories, self.walkDirectory), skuInfo];
+        }).spread(function(paths, skuInfo) {
+            var flattenedPaths = [].concat.apply([], paths);
+            flattenedPaths.forEach(function(staticFile) {
+                if ("sku" in staticFile) {
+                    if (skuInfo.hasOwnProperty(staticFile.sku)) {
+                        staticFile.sku = skuInfo[staticFile.sku];
+                    } else {
+                        delete staticFile.sku;
+                    }
+                }
+            });
+            return flattenedPaths;
+        });
+    };
+
+
+    return new StaticFilesApiService();
+
+}
--- a/spec/lib/services/static-files-api-service-spec.js
+++ b/spec/lib/services/static-files-api-service-spec.js
@@ -1,0 +1,200 @@
+// Copyright 2016, EMC, Inc.
+/* jshint node:true */
+
+"use strict";
+
+describe("Http.Services.Api.StaticFiles", function () {
+    var path;
+    var staticFilesApiService;
+    var waterline;
+
+    before("Http.Services.Api.StaticFiles before", function() {
+        helper.setupInjector([
+            helper.require("/lib/services/static-files-api-service.js")
+        ]);
+        staticFilesApiService = helper.injector.get("Http.Services.Api.StaticFiles");
+        path = helper.injector.get('path');
+        waterline = helper.injector.get('Services.Waterline');
+        waterline.skus = {
+            findOne:function() {}
+        };
+        this.sandbox = sinon.sandbox.create();
+    });
+
+    afterEach("Http.Services.Api.StaticFiles afterEach", function() {
+        this.sandbox.restore();
+    });
+
+
+    describe("pairSkupackIds", function() {
+        var nodeFs;
+        before(function() {
+            nodeFs = helper.injector.get('fs');
+            sinon.stub(nodeFs, 'lstatAsync');
+            sinon.stub(nodeFs, 'readdirAsync');
+            this.sandbox.stub(waterline.skus, 'findOne');
+        });
+
+        beforeEach(function() {
+            nodeFs.lstatAsync.reset();
+            nodeFs.readdirAsync.reset();
+        });
+
+        after(function () {
+            nodeFs.lstatAsync.restore();
+            nodeFs.readdirAsync.restore();
+        });
+
+
+        it('should return an id:sku pair given valid directories and database', function() {
+            nodeFs.readdirAsync.resolves(['567432']);
+            nodeFs.lstatAsync.resolves({ isDirectory: function() { return true; }});
+            waterline.skus.findOne.resolves({
+                id: '567432',
+                name: 'testSku'
+            });
+            return staticFilesApiService.pairSkupackIds([]).then(function(value) {
+                expect(nodeFs.readdirAsync).to.have.been.called;
+                expect(nodeFs.lstatAsync).to.have.been.called;
+                expect(waterline.skus.findOne).to.have.been.called;
+                expect(value).to.deep.equal({'567432': 'testSku'});
+            });
+
+        });
+
+        it('returns {} when there is no sku data in the database', function() {
+            nodeFs.readdirAsync.resolves(['123']);
+            nodeFs.lstatAsync.resolves({ isDirectory: function() { return true; }});
+            var stubFind = sinon.stub(waterline.skus, "findOne");
+            stubFind.rejects(new Error('Error: no available sku in the db'));
+            return staticFilesApiService.pairSkupackIds([]).then(function(value) {
+                expect(nodeFs.readdirAsync).to.have.been.called;
+                expect(nodeFs.lstatAsync).to.have.been.called;
+                expect(waterline.skus.findOne).to.have.been.called;
+                expect(value).to.deep.equal({});
+            });
+        });
+
+    });
+
+
+    describe("walkDirectory", function() {
+
+        it('returns an array of file names following "static" in the path', function() {
+            var dirPath = path.join("spec/mocks/static");
+            var result = [ 
+                { uri: 'found' },
+                { uri: 'foundFile2' }
+            ];
+            return staticFilesApiService.walkDirectory(dirPath).then(function(value) {
+                expect(value).to.deep.equal(result);
+            });
+        });
+
+        it('returns an array of file names following "skupack.d" in the path', function() {
+            var dirPath = "spec/mocks/skupack.d";
+            var result = [
+                {
+                    sku: "mocks",
+                    uri: "sku-id/static/found-2"
+                },
+                {
+                    sku: "mocks",
+                    uri: "sku-id/static/common/found-3"
+                }
+            ];
+            return staticFilesApiService.walkDirectory(dirPath).then(function(value) {
+                expect(value).to.deep.equal(result);
+            });
+        });
+
+        it('returns [] when provided a bad path', function() {
+            var dirPath;
+            var result = [];
+            return staticFilesApiService.walkDirectory(dirPath).then(function(value) {
+                expect(value).to.deep.equal(result);
+            });
+        });
+
+    });
+
+
+    describe("getAllStaticFiles", function() {
+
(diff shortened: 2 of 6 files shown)


lib/services/files-api-service.js
```diff
@@ -0,0 +1,129 @@
...
+function filesApiServiceFactory(
+    constants,
+    Promise,
+    configuration,
+    waterline
+) {
+    var walker;
+    var walk = Promise.promisifyAll(walking);
+    var skuPath = constants.HttpStatic.onhttpSkupack;
+    var staticPath = constants.HttpStatic.onhttpStaticHttp;
```
yyscamper: I don't find these two constants from on-core.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,129 @@
...
+'use strict';
+
+var di = require('di');
+var fs = require('fs');
+var path = require('path');
+var walking = require('walk');
+
+module.exports = filesApiServiceFactory;
+
+di.annotate(filesApiServiceFactory, new di.Provide('Http.Services.Api.StaticFiles'));
```
yyscamper: The source file name doesn't well match with this injectable name, suggest you to change the file name to something like `static-files-api-service`

lib/services/files-api-service.js
```diff
@@ -0,0 +1,129 @@
...
+    var skuPath = constants.HttpStatic.onhttpSkupack;
+    var staticPath = constants.HttpStatic.onhttpStaticHttp;
+    Promise.promisifyAll(fs);
+
+    function FilesApiService() {
+    }
+
+
+    FilesApiService.prototype.pairSkupackIds = function() {
+        return fs.readdirAsync(skuPath).then(function(list) {
```
yyscamper: You can flatten this promise chain:
```javascript
return fs.readdirAsync(...)
.filter(...)
.reduce(...)
```

yyscamper:
I see following example from your PR comment, what's this means, why the `sku` is mixed with static files in common folder?
```
{
"sku": "Cisco Nexus 3000 Switch - 54 port",
"uri": "static/common/aTestFile.we"
}
```

lib/services/files-api-service.js
lacarb: I still have to put up a pr for that. The on-core constants.js code is 
`HttpStatic: {
            onhttpStaticHttp: 'static/http',
            onhttpSkupack: 'skupack.d'
        }`

lib/services/files-api-service.js
lacarb: I like that suggestion. Does anyone else have an opinion on that name change?

lacarb:
@yyscamper the `sku` field is only included for static files within skupacks. I don't believe any current skupacks have a static directory within them however I followed the format presented here which includes a `static/common` path. http://rackhd.readthedocs.io/en/latest/rackhd/skus.html?highlight=skupack#sku-pack-tar-gz-format

lib/services/files-api-service.js
benbp: Put a link to the PR in your description here when you do, and make sure to say "requires <link>" otherwise we don't have a way of quickly knowing that we _shouldn't_ merge this PR until the other one goes in.

lib/services/files-api-service.js
benbp: Maybe even `list-static-files-api-service` ?

lib/services/files-api-service.js
benbp: Good point! (sorry ben 😄 )

benbp:
@yyscamper the sku field is there since you can't actually download that static file if the request isn't coming from a client identified as being of that sku.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,129 @@
...
+                },{});
+            });
+        });
+    };
+
+
+    FilesApiService.prototype.walkDirectory = function(dir) {
+        return new Promise(function(resolve, reject) {
+            var walkResults = [];
+            walker = walk.walk(dir, {followLinks: false});
```
benbp: Is it possible for this to fail (like if dir is undefined)? If so, might be good to wrap it in a try/catch so we can reject and exit on failure.

lib/services/files-api-service.js
```diff
@@ -0,0 +1,129 @@
...
+
+
+    FilesApiService.prototype.getAllStaticFiles = function() {
+
+        var self = this;
+        var directories = [staticPath,
+                           configuration.get('httpStaticRoot', '/opt/monorail/static/http')];
+
+        return fs.lstatAsync(skuPath).then(function(stat) {
+            stat.isDirectory();
```
benbp: This seems like a noop?

lib/services/files-api-service.js
```diff
@@ -0,0 +1,129 @@
...
+            directories.push(skuPath);
+            return self.pairSkupackIds();
+        }).catch(function(e) {
+            if ((e.code !== 'ENOENT') && (e.path !== skuPath)) {
```
benbp: If you want, you can use bluebird's error filtering here (up to you though, I'll +1 without it).

```
.catch({code: 'ENOENT'}, function(e) {
    return [];
}).catch(function (e) {
    if (e.path !== skuPath)) {
        throw e;
    }
    return [];
})
```

lib/services/files-api-service.js
```diff
@@ -0,0 +1,129 @@
...
+
+            walker.on("file", function(root, fileStats, next) {
+                Promise.resolve().then(function() {
+                    var walkObj = {};
+                    if (root.indexOf(staticPath) > -1){
+                        root = root.replace(staticPath, '');
+                        walkObj.uri = path.join(root, fileStats.name);
+                        walkResults.push(walkObj);
+                    } else if ((root.indexOf(skuPath) > -1) && 
+                      (root.indexOf("static/common") > -1)){
```
benbp: We support directories other than `static/common`, i.e. `static/centos` etc. Basically any directory or file within static is valid.

spec/lib/services/files-api-service-spec.js
```diff
@@ -0,0 +1,86 @@
...
+                "workflowRoot": "workflows",
+                "httpProfileRoot": "profiles",
+                "httpTemplatesRoot": "templates",
+                "httpStaticRoot": "static"
+            };  
+            waterline.skus.findOne.withArgs({name: "my test sku"}).resolves(sku);
+            skuService.regenerateSkus.resolves();
+            sinon.spy(filesApiService, "getAllStaticFiles");
+            filesApiService.getAllStaticFiles();
+            filesApiService.getAllStaticFiles.called.should.be.true;
```
benbp: This doesn't appear to really be testing anything other than that a call to `getAllStaticFiles` doesn't throw an exception. It would be better to provide several test cases that exercise your code more fully, such as (at a minimum):

- The walker adds `sku` attributes to directories within skupack.d, as well as constructing result objects with a `uri` attribute like you expect (it's also good just so we have an example of the expected output in the test to look at).
- getAllStaticFiles handles ENOENT errors properly.
- getAllStaticFiles returns the expected flattenedPaths output data.
- pairSkuPackIds adds sku names to the result data.

I'm happy to show you ways in which you can create mocks for the walker, fs calls, etc. if you need examples.

yyscamper:
@lacarb @benbp : Thanks, got your point about the `sku` filed. But for SKU static files, I'd suggest asking for sku-pack-service (https://github.com/RackHD/on-http/blob/master/lib/services/sku-pack-service.js) to get the static files for all skus. Because skupack service owns how the sku data is stored. If in future, the sku data structure has something change, developer may not be aware the static api service will still be changed.

lib/services/files-api-service.js
lacarb: Added

yyscamper:
Meanwhile, we have a special key in config.json to let user define his own static folder, I think you still need to take this folder into consideration, check following code:
https://github.com/RackHD/on-http/blob/master/lib/services/http-service.js#L218

benbp:
I agree with @yyscamper, it would be good to encapsulate the logic for "what sku files are there" within the sku-pack-service, even it uses a shared mechanism under the hood today. Should just require moving some of your existing code around.

lib/services/files-api-service.js
lacarb: I've got it down to this. Sparing a push in progress, is there more reduction that I'm missing or care for things like which line the `then` statements are on?
```
StaticFilesApiService.prototype.pairSkupackIds = function() {
    return fs.readdirAsync(skuPath).filter(function(item) {
        return (fs.lstatAsync(path.join(skuPath, item)))
        .then(function(stat) {
            return stat.isDirectory();
        }); 
    }).reduce(function(result, item) {
        return waterline.skus.findOne({id : item})
        .then(function(sku) {
            result[item] = sku.name;
            return result;
        }); 
    },{});
};
```

benbp: Looks great to me.

lib/services/files-api-service.js
lacarb: This check is here to see if the path in `skuPath` exists. If it does not then an `ENOENT` error is thrown. If it does exists then it gets pushed to the directories array.

My concern here is to look within skupath when it exists. If anyone has a problem with this approach and a suggested alternative then please let me know.

benbp: No it's fine, just non-obvious behavior that isDirectory would throw, may be worth a comment.

yyscamper:
@lacarb : you need rebase your branch to make the Jenkin/Travis success.

changev:
test this please

brianparry:
test this please

changev:
test this please

rackhdmirror:
Build finished.

changev:
Jenkins: depends on https://github.com/RackHD/on-core/pull/183
test this please

rackhdmirror:
Build finished.

changev:
Hi @lacarb  for test interdependent PRs, you can write
Jenkins: depends on https://github.com/RackHD/on-core/pull/183
And later the test result can see in both PRs.

changev:
test this please

rackhdmirror:
Build finished.

yyscamper:
@lacarb , @benbp, @pscharla : Is this still a valid PR? As I see the error is pending for a long time and not have a fix. If this is not valid, I will close it.

blacara:
@yyscamper This PR is valid. I was asked to switch focus to another ticket when I was working to resolve the error in this one. I was talking with @brianparry about this.
ATM I'm not sure why my code throws the error above when I did not modify the code that is raising the error. I'll bring my focus back to resolving this soon.

rackhdmirror:
Build finished.

PengTian0:
test this please

pscharla:
test this please

pscharla:
Jenkins: test this please

rackhdmirror:
Build finished.

PengTian0:
test this please

PengTian0:
Jenkins: test this please

rackhdmirror:
Build finished.

lacarb:
Finally! @tannoa2 found the solution, thanks again!
@benbp @yyscamper @brianparry @anhou

lacarb:
Hmm hold off. The fix, while correct, exposed something.

lacarb:
Assuming the -0.03 coverage is forgivable, this is ready.
@benbp @yyscamper @brianparry @anhou

yyscamper:
@lacarb : you committed some test files please remove them:

image

lacarb:
@yyscamper The test files create a file tree used for testing the walk method in the code. Without them I am unsure how to go about testing that method. What would you recommend?

lib/services/static-files-api-service.js
```diff
@@ -0,0 +1,155 @@
...
+
+
+    StaticFilesApiService.prototype.getAllStaticFiles = function() {
+
+        var self = this;
+        var directories = [staticPath,
+                           configuration.get('httpStaticRoot', userPath)];
+
+        return fs.lstatAsync(skuPath).then(function(stat) {
+            stat.isDirectory();
```
lacarb: Good catch, those lines have been removed and the inclusion of the skuPath directory has been fixed.

lib/services/static-files-api-service.js
```diff
@@ -0,0 +1,155 @@
...
+
+    StaticFilesApiService.prototype.pairSkupackIds = function(dirs) {
+        return Promise.try(function() {
+            return fs.readdirAsync(skuPath).filter(function(item) {
+                return (fs.lstatAsync(path.join(skuPath, item)))
+                .then(function(stat) {
+                    return stat.isDirectory();
+                }); 
+            }).reduce(function(result, item) {
+                return Promise.try(function() {
```
yyscamper: The promise.try is not necessary here.

lib/services/static-files-api-service.js
```diff
@@ -0,0 +1,155 @@
...
+        })
+        .catch(function(e) {
+            logger.debug("StaticFilesApiService.pairSkupackIds", e);
+            return {};
+        });
+    };
+
+
+    StaticFilesApiService.prototype.walkDirectory = function(dir) {
+        return Promise.try(function() {
```
yyscamper: The Promise.try is not necessary here.

lib/services/static-files-api-service.js
```diff
@@ -0,0 +1,155 @@
...
+            return new Promise(function(resolve, reject) {
+                var walkResults = [];
+                walker = walk.walk(dir, {followLinks: false});
+
+                walker.on("file", function(root, fileStats, next) {
+                    Promise.resolve().then(function() {
```
yyscamper: Some comment here: 
- The callback function doesn't have to be a promise
- The promise should be rejected if the walker emits the error events.
- Since you don't walk on the sub-directories, the sub-directories is intentionally to ignore?


I suggest using following code structure (only some code snippet):
```
StaticFilesApiService.prototype.walkDirectory = function(dir) {
    return new Promise(function(resolve, reject) {
        var walkResults = [];
        walker = walk.walk(dir, {followLinks: false});
        walker.on("file", function(...) {
             try {
                 dothing here and add result to walkResults;
                 next();
             } catch (err) {
                 reject(err);
             }
        });
        walker.on("errors", function(err) {
            reject(err);
        });
        walker.on("end", function() {
             resolve(walkResults);
        });
   });
}
```

spec/lib/services/static-files-api-service-spec.js
```diff
@@ -0,0 +1,213 @@
...
+        after(function () {
+            nodeFs.lstatAsync.restore();
+            nodeFs.readdirAsync.restore();
+        });
+
+
+        it('should return an id:sku pair given valid directories and database', function() {
+            nodeFs.readdirAsync.resolves(['567432']);
+            nodeFs.readdirAsync().then(function (dirOutput) {
+                expect(dirOutput).to.deep.equal(['567432']);
```
yyscamper: This assertion is not necessary here, it is only to test the line #50, but that line can be trusted.

spec/lib/services/static-files-api-service-spec.js
```diff
@@ -0,0 +1,213 @@
...
+                expect(nodeFs.readdirAsync).to.have.been.called;
+                expect(nodeFs.lstatAsync).to.have.been.called;
+                expect(waterline.skus.findOne).to.have.been.called;
+                expect(value).to.deep.equal({'567432': 'testSku'});
+            });
+
+        });
+
+        it('returns {} when there is no skupack.d directory', function() {
+            nodeFs.readdirAsync.throws();
```
yyscamper: readdirAsync should return a promise, so use `nodeFs.readdirAsync.rejects();` here

spec/lib/services/static-files-api-service-spec.js
```diff
@@ -0,0 +1,213 @@
...
+            return staticFilesApiService.pairSkupackIds([]).then(function(value) {
+                expect(value).to.deep.equal({});
+            });
+        });
+
+        it('returns {} when there is no sku data in the database', function() {
+            nodeFs.readdirAsync.resolves(['123']);
+            nodeFs.lstatAsync.resolves({ isDirectory: function() { return true; }});
+            var stubFind = sinon.stub(waterline.skus, "findOne");
+            stubFind.throws('error');
```
yyscamper: `waterline.skus.findOne` is designed  to return a promise, so use `stubFind.rejects(new Error('some error'))` rather than `throws`

yyscamper:
@lacarb : I didn't realize those files are used in your unit-testing, if so, that's OK for me.

lib/services/static-files-api-service.js
lacarb: This Promise.try protects from bad dir values being passed to the method. Such as null.

lib/services/static-files-api-service.js
lacarb: - The callbacks are structured according to the module, walk, I'm leveraging
- Sub-directories are walked. For the above and this you can refer to the github page https://github.com/Daplie/node-walk

lib/services/static-files-api-service.js
```diff
@@ -0,0 +1,155 @@
...
+    var walker;
+    var walk = Promise.promisifyAll(walking);
+    Promise.promisifyAll(fs);
+
+    function StaticFilesApiService() {
+    }
+
+
+    StaticFilesApiService.prototype.pairSkupackIds = function(dirs) {
+        return Promise.try(function() {
```
lacarb: Agreed, removed.

spec/lib/services/static-files-api-service-spec.js
lacarb: Thanks for the explanation. They're removed.

spec/lib/services/static-files-api-service-spec.js
lacarb: Due to changes in static-files-api-service.js this test is no longer necessary and has been removed.

spec/lib/services/static-files-api-service-spec.js
lacarb: Changed it to rejects.

lacarb:
@anhou @yyscamper @brianparry @benbp 
Incorporated changes from Felix's feedback.
Can I get some more eyes on this?

pscharla:
@yyscamper : Good with @lacarb's changes?

lib/services/static-files-api-service.js
yyscamper: If so, you should validate the `dir` at beginning. Using a big `Promise.try` to wrap all the code actually will hide many problems since all exception, not only the bad dir exception, will also be caught.

lib/services/static-files-api-service.js
yyscamper: @lacarb : the sub directories may have already been covered, if your testing is OK, that's OK.

But for this promise itself, Since the code didn't listen on "error" event, so you want to intentionally ignore all the walk error? But some error cannot be ignored, for example, for a wrong directory input, the walk should send some error and you will get empty result during `end` event.

yyscamper:
@pscharla : Most code is OK to me, there are two questions need either fix or clarification.

lib/services/static-files-api-service.js
lacarb: walker.on("errors...
has been added. Thank you for pointing this out.

lib/services/http-service.js
```diff
@@ -48,6 +48,7 @@ module.exports = httpServiceFactory;
 di.annotate(httpServiceFactory, new di.Provide('Http.Server'));
 di.annotate(httpServiceFactory,
     new di.Inject(
+        'Constants',
```
lacarb: Thank you for pointing this out. It has been removed.

pscharla:
test this please

lacarb:
@yyscamper How does this look now?

DavidjohnBlodgett:
👍 may want to squash those commits before merge though.

lacarb:
@DavidjohnBlodgett @yyscamper @benbp 
Commits are squashed and ready for merge.
Thanks for all the feedback and patience.

PR #376, opened 2016-07-19
Add retries to redirect.ipxe chainloading. Remove error.ipxe in favor…

benbp:
… of HTTP error responses

This is intended to solve some load issues where RPC timeouts occur when the HTTP server requests a profile rendering context from a task before it has become active/is listening on AMQP. Currently, we have no retry logic at all for this pathway, so this adds some quick and dirty retry logic to the caller (redirect.ipxe). As a long term solution, we'll want to make our RPC calls more robust with retries at the protocol layer. However, my thinking here is that it's better to retry the API from the client rather than have our API server cause the initial request to take longer while it waits on AMQP retries.

This is also intended as a more robust alternative to https://github.com/RackHD/on-http/pull/361

I removed the error.ipxe behavior we had previously as part of this. While it may in some cases have been helpful to see the API error messages in the ipxe console, I think for most people without context they are meaningless, and they can still be found in the on-http logs. The root issue I had was that I couldn't determine if chainloading had failed because the API would return the error.ipxe script with a 200 status. This change makes the API return a 400/500/503 status so that we can determine a chainload failure and trigger a retry. Additionally, chainloading multiple error.ipxe scripts back to back while we retry for a success caused my machine to kernel panic once it succeeded in chainloading linux.ipxe and booting the initrd, so I just removed that whole sequence for the reasons above.

(See https://hwjiraprd01.corp.emc.com/browse/ODR-786)

NOTE: I think I may have exposed some test setup issues. I made the test error checking more explicit and I think it exposed some global test state leaks (returning the right error code but the wrong error type/message). We'll see if this is just local for me, otherwise I'll investigate those test state issues as part of this.

@RackHD/corecommitters @heckj @pscharla @manfrednde @miramm


yyscamper:
:+1:

yyscamper:
jenkins: test this please

heckj:
👍

anhou:
@benbp error.ipxe still need to be used by  _handleProfileRenderError() in file "lib/services/profiles-api-service.js" and it's also very useful for debug

benbp:
@anhou I'm just reading through the section of the code you're referring to again, and I think there is probably some further refinement we can do. For example, the "Unable to retrieve node bootSettings" error will _only_ happen if the node document _has_ a bootSettings field but it is invalid. Ideally, this is something we would enforce with API and database schemas, such that it would be impossible to even hit this part of the code. So I think I would rather fix it there and remove this whole section of the code (though there could be issues with bootSettings being added through the 1.1 API prior to these changes).

Additionally, the final error case "Unable to locate active workflow or there are no bootSettings" should only happen in the chance that a workflow is active for the node when dhcp-proxy hands out a bootfile (https://github.com/RackHD/on-dhcp-proxy/blob/master/lib/message-handler.js#L144), but is then completed or cancelled at the time of the next active workflow check (https://github.com/RackHD/on-http/blob/master/lib/services/profiles-api-service.js#L277). I would argue that seeing this error message in the node console window doesn't really help you figure out what the actual problem was if you don't have intimate knowledge of this part of the system since it's likely an extremely rare race condition. Maybe as a bare minimum though we could add the nodeId to the final "retries failed" message, so that it's easier to look up log messages after the fact from a console screenshot?

anhou:
@benbp Thanks for details' feedback. I have some thoughts for the whole RackHD <-> ipxe communication.
I think ipxe console should show more information as more as possible, either for debug, or report bugs, or let user know what had happened. Of course, it's better that returned information from RackHD more descriptive, and easy for user understanding, and don't need to search the logs. It's like the northbound API, when user sending/getting data from RackHD northbound by REST API,  RackHD acknowledge a status is not enough, it's better with some details such as schema failure with details in payloads, although these information also be be got from the logs. It's the similar with southbound, the difference is the client is ipxe, and the chain commands only have success and failure two status, don't have a payload,  but we could workaround it by let it always return success, but describe the error or non-error message in returned ipxe profile if it's not an connection failure. and these information could be user-friendly understanding. and also more error handling could be handled in returned ipxe profile such as retry, or different retry time, or maybe a new /profile API access with customized additional query. 
so totally I want to have a generic method to handle the communication of RackHD and ipxe,  so maybe 'error.ipxe' could be extended to 'profile-return-error.ipxe' and 'profile-return-info.ipxe' .  _handleProfileRenderError() also could be extended  to have a method _handleProfileRenderInfo(). so from RackHD side, it will be flexible if we don't only return status. I'm fine with this retry when chainloading fail in this PR, just think it will be useful for RackHD still have the ability to return some communication info such as  error.ipxe( maybe 'profile-return-error.ipxe'  'profile-return-info.ipxe' in future, or more, 'profile-return-error-retry.ipxe) except functionality ipxe script and 'redirect.ipxe' script. How do you think?

And I agree 'Unable to retrieve node bootSettings' could be refined with API and database schema.  and for 'Unable to locate active workflow or there are no bootSettings' maybe you've noticed that my proposal in google groups https://groups.google.com/forum/#!topic/rackhd/I81Tl2qkQMI, after this enhancment, the ipxe actions will rely on GET /profiles. this message will be a positive info like 'No active workflow and boot settings for this discovered node, continue to boot up' to tell user what happened, not only words like 'nothing to boot' in server ipxe console.

benbp:
@anhou a couple points:

**1. regarding technical limitations**

The way the system behaves in master is as you describe: always return success, albeit with or without an error message (error.ipxe). When I added the retry logic, I was causing kernel panics during bootup by successfully chainloading multiple error.ipxe scripts together before finally chainloading the boot ipxe script. I have no idea why this was happening, but I suspect there's an issue with how ipxe handles chainloading that could cause problems with implementing a "succeed chainload with an error message and proceed" type of behavior. That said, I could have just been doing something incorrectly and this is a straw man argument.

The way I resolved the above issue was by forcing a non-2xx HTTP status (4xx/5xx), which would cause the chain command to exit with a non-zero status, letting me use ipxe's `goto` way of handling errors. As far as I know, there isn't another way to do it, and also no way to print out anything from the server given a non-2xx response. So I refactored the code to be explicit regarding HTTP status for this reason. 

An ideal compromise between your points and mine would be to render 5xx statuses down until the final retry, and then render down error.ipxe with a 200 response. The challenge here is we can't know on the server side when the retries have been exceeded, because we don't store that state, and it seems like a lot of unnecessary complexity to try to store it. Perhaps we could embed a retry value into the chainload GET request, and we can keep track of max retries on the API server side, rather than in the script (so the logic to stop retrying is on server side in javascript and not client side in ipxe-script).

**2. regarding usability**

Part of my hesitation to spend effort presenting error messages to the ipxe console is that I think it encourages a pattern of debugging that may be helpful in development cases, but is actually less useful than we think in real use cases. 

A couple reasons I think this:
- The error message on the console screen is not sufficient to determine what went wrong, and I think often confuses people more than it helps them.
- Related to above, since we know boot issues are an already only a handful of possibilities (boot settings, invalid lookup and arp state, workflow failures, bad database state), I think it's less useful than it seems to get these messages. If there were more than a dozen possible error cases, then I think it would be more useful, but today you can count them on one hand and they all require server-side logs and database dumps to figure out what went wrong. As such, the error message itself doesn't really help us narrow down the possibilities since it is always "couldn't render workflow properties".
- Often the error message only exists on the console for a second before it scrolls by, so it's never actually seen (or only partially seen), as opposed to an API where you can re-issue commands, inspect output, etc. Regardless, an error message scrolling by encourages users to wait by a console screen for minutes at a time trying to capture with a screenshot what happened, and we often get sent these screenshots in bug reports _instead_ of logs, but they don't actually tell us the root cause of what went wrong. So I see this as encouraging a pattern of debugging problematic behavior that ends up being more inefficient than if you just had to look at the logs no matter what.

benbp:
@zyoung51 can tell me if I've devolved this into bikeshedding or not ^^ 🚲 😄

anhou:
@benbp Hopefully we're not  in bikeshedding :) 

* For technical limitations
    - I also think it's fine that non-20x status, then let ipxe script to retry, it's also useful when network connection is not so good such as connection timeout failure
    - I'm not sure what's the kernel panic, in the beginning, do you also write ipxe script like below logic?
```
isset ${RetryCount} || goto firstretry
inc RetryCount
show RetryCount
iseq RetryCount 8 || goto retrydone
echo RackHD: retry exceed 8 times
exit

:firstretry
set RetryCount:int8 0
show RetryCount
:retrydone

chain http://xxxx
```
    - the compromise way to add 'retry'  query is also a choice, and it's also a way to let ipxe to communicate with RackHD,  Personally, I think it's better to avoid let RackHD maintain the retry state to increase the complexity. if the logic like above script cannot resolve the problem and still encounter kernel panic, I also prefer to use non-200 status to retry just like the PR's method.

* I just think let RackHD GET /profiles route continue to have the ability to return some communication scripts, it also wouldn't block this PR's retry mechanism, two benefits
    - allow more flexibility and extension, and handle more possible cases in future, for example, sometime, some errors just want ipxe to exit immediately, some errors want to ipxe retry 30s, some retry 60s. as I can think of, return 200 status with some specific ipxe error handling scripts could achieve this. 
    - logs in pxe and usability, maybe we still have different views here. if we cannot be consistent in this point,  how about considering above points :)

anhou:
:+1:  and test this please

PR #377, opened 2016-07-20
Add southbound API to pass notification to RackHD (refer to ODR682)

cgx027:
/notification API defined to replace ```completionUri``` used by os installation, with benifit of not depending to lookup entry, refer to ODR682.

In this PR:
* POST /notification southbound API is defined, used by node to signaling os install finish. 
* Both API 1.1 and API 2.0 compatible.
* Photon OS kickstart file modified to make use of the new notification mechanism.
 

This PR depends on following PRs:
* https://github.com/RackHD/on-core/pull/184 
* https://github.com/RackHD/on-tasks/pull/284

### DIFF
--- a/lib/api/1.1/southbound/notification.js
+++ b/lib/api/1.1/southbound/notification.js
@@ -11,7 +11,7 @@ di.annotate(notificationRouterFactory, new di.Provide('Http.Api.Internal.Notific
 di.annotate(notificationRouterFactory,
     new di.Inject(
         'Http.Services.RestApi',
-		'Http.Services.Api.Notification',
+        'Http.Services.Api.Notification',
         '_'
 	)
 );
--- a/lib/api/1.1/southbound/notification.js
+++ b/lib/api/1.1/southbound/notification.js
@@ -32,8 +32,8 @@ function notificationRouterFactory (
      */
 
     router.post('/notification/', rest(function (req) {
-        var configuration = _.defaults(req.query || {}, req.body || {});
-        return notificationApiService.postNotification(configuration);
+        var message = _.defaults(req.query || {}, req.body || {});
+        return notificationApiService.postNotification(message);
     }, {renderOptions: {success: 201}}));
 
     return router;
--- a/lib/api/1.1/southbound/notification.js
+++ b/lib/api/1.1/southbound/notification.js
@@ -1,0 +1,40 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di'),
+    express = require('express');
+
+module.exports = notificationRouterFactory;
+
+di.annotate(notificationRouterFactory, new di.Provide('Http.Api.Internal.Notification'));
+di.annotate(notificationRouterFactory,
+    new di.Inject(
+        'Http.Services.RestApi',
+		'Http.Services.Api.Notification',
+        '_'
+	)
+);
+
+function notificationRouterFactory (
+    rest,
+    notificationApiService,
+    _
+) {
+    var router = express.Router();
+
+    /**
+     * @api {post} /api/1.1/notification
+     * @apiVersion 1.1.0
+     * @apiDescription post a notification from node
+     * @apiName notification-post
+     * @apiGroup notification
+     */
+
+    router.post('/notification/', rest(function (req) {
+        var configuration = _.defaults(req.query || {}, req.body || {});
+        return notificationApiService.postNotification(configuration);
+    }, {renderOptions: {success: 201}}));
+
+    return router;
+}
--- a/lib/api/1.1/southbound/templates.js
+++ b/lib/api/1.1/southbound/templates.js
@@ -10,12 +10,12 @@ module.exports = templatesRouterFactory;
 di.annotate(templatesRouterFactory, new di.Provide('Http.Api.Internal.Templates'));
 di.annotate(templatesRouterFactory,
     new di.Inject(
-		'Services.Waterline',
-		'Http.Services.Api.Workflows',
-		'Protocol.Task',
-		'Services.Lookup',
-		'Promise',
-		'common-api-presenter',
+	    'Services.Waterline',
+	    'Http.Services.Api.Workflows',
+	    'Protocol.Task',
+	    'Services.Lookup',
+	    'Promise',
+	    'common-api-presenter',
         '_'
 	)
 );
@@ -43,7 +43,6 @@ function templatesRouterFactory (
     router.get('/templates/:identifier',
                lookupService.ipAddressToMacAddressMiddleware(),
                function (req, res) {
-        var nodeId = '';
         if (!req.macaddress) {
             return presenter(req, res)
                 .renderError(new Error("Unable to look up the relevant node from " +
@@ -51,7 +50,6 @@ function templatesRouterFactory (
         }
         waterline.nodes.findByIdentifier(req.macaddress).then(function (node) {
             if (node) {
-                nodeId = node.id;
                 return workflowApiService.findActiveGraphForTarget(node.id)
                     .then(function(taskGraphInstance) {
                         if (taskGraphInstance) {
@@ -74,7 +72,7 @@ function templatesRouterFactory (
             presenter(req, res)
                 .renderTemplate(
                     req.params.identifier,
-                    _.merge(properties.options, {identifier: nodeId}),
+                    properties.options,
                     200,
                     properties.context
                 );
--- a/lib/api/1.1/southbound/templates.js
+++ b/lib/api/1.1/southbound/templates.js
@@ -51,7 +51,6 @@ function templatesRouterFactory (
         }
         waterline.nodes.findByIdentifier(req.macaddress).then(function (node) {
             if (node) {
-                console.log("--------------------Ted:node", node);
                 nodeId = node.id;
                 return workflowApiService.findActiveGraphForTarget(node.id)
                     .then(function(taskGraphInstance) {
--- a/lib/api/1.1/southbound/templates.js
+++ b/lib/api/1.1/southbound/templates.js
@@ -15,8 +15,7 @@ di.annotate(templatesRouterFactory,
 	    'Protocol.Task',
 	    'Services.Lookup',
 	    'Promise',
-	    'common-api-presenter',
-        '_'
+	    'common-api-presenter'
 	)
 );
 
@@ -26,8 +25,7 @@ function templatesRouterFactory (
     taskProtocol,
     lookupService,
     Promise,
-    presenter,
-    _
+    presenter
 ) {
     var router = express.Router();
 
--- a/lib/api/1.1/southbound/templates.js
+++ b/lib/api/1.1/southbound/templates.js
@@ -15,7 +15,8 @@ di.annotate(templatesRouterFactory,
 		'Protocol.Task',
 		'Services.Lookup',
 		'Promise',
-		'common-api-presenter'
+		'common-api-presenter',
+        '_'
 	)
 );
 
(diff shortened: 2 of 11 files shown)


yyscamper:
@tannoa2 This is the notification API that you want.

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,94 @@
...
+            if (!configuration.type || !_.isString(configuration.type)) {
+                throw new Errors.BadRequestError('Missing notification type in query or body');
+            };
+            if (!configuration.data || !_.isString(configuration.data)) {
+                throw new Errors.BadRequestError('Missing notification data in query or body');
+            }            
+        })
+        .then(function() {
+            nodeId = configuration.nodeId;
+            return waterline.nodes.needByIdentifier(nodeId)
```
WangWinson: so this line is to confirm that the specific node does exist ?

lib/api/1.1/southbound/notification.js
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @api {post} /api/1.1/notification
+     * @apiVersion 1.1.0
+     * @apiDescription post a notification from node
+     * @apiName notification-post
+     * @apiGroup notification
+     */
+
+    router.post('/notification/', rest(function (req) {
+        var configuration = _.defaults(req.query || {}, req.body || {});
```
WangWinson: Will the notification info go into req.body ?
I guess we only put this info into req.query , right ? So no need to merge req.body here.

lib/api/1.1/southbound/templates.js
```diff
@@ -41,13 +43,16 @@ function templatesRouterFactory (
...
                function (req, res) {
+        var nodeId = '';
         if (!req.macaddress) {
             return presenter(req, res)
                 .renderError(new Error("Unable to look up the relevant node from " +
                     "the request, can't render template."));
         }
         waterline.nodes.findByIdentifier(req.macaddress).then(function (node) {
             if (node) {
+                console.log("--------------------Ted:node", node);
```
WangWinson: Tada ! I found your name here.

lib/api/2.0/notification.js
```diff
@@ -0,0 +1,17 @@
...
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var notificationApiService = injector.get('Http.Services.Api.Notification');
+var _ = injector.get('_');    // jshint ignore:line
+
+var notificationPost = controller({success: 201}, function(req) {
+    var configuration = _.defaults(req.swagger.query || {}, req.body || {});
```
WangWinson: Same as above

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,94 @@
...
+    Errors,
+    Promise,
+    _
+) {
+    var logger = Logger.initialize(notificationApiServiceFactory);
+
+    function notificationApiService() {
+    }
+
+    notificationApiService.prototype.postNotification = function(configuration) {
```
WangWinson: param name : `configuration`  ? will `message` been more meaningful ?

lib/api/1.1/southbound/templates.js
```diff
@@ -15,7 +15,8 @@ di.annotate(templatesRouterFactory,
 		'Protocol.Task',
 		'Services.Lookup',
 		'Promise',
-		'common-api-presenter'
+		'common-api-presenter',
+        '_'
```
iceiilin: inconsistent indent?

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,94 @@
...
+                taskGraphStore.findActiveGraphForTarget(nodeId),
+                taskProtocol.activeTaskExists(nodeId)
+            ]
+        })
+        .spread(function (activeGraph, activeTask) {
+            if (!activeGraph) {
+                throw new Error("No active task graph.");
+            }
+            if (!activeTask) {
+                throw new Error("No active task.");
```
iceiilin: Do we need to throw BadRequestError similar as above?

static/monorail-2.0-sb.yaml
```diff
@@ -118,7 +118,54 @@ paths:
...
+      summary: |
+        post a notification
+      description: |
+        post a notification
+      parameters:
+        - name: nodeId
+          in: query
+          description: |
+            node identifier
+          required: false
```
iceiilin: Is there parameters required here?

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,94 @@
...
+                configuration.data
+                )
+        })
+        .then(function () {
+            return {
+                node: nodeId,
+                activeTask: activeTaskId,
+                activeGraph: activeGraphId,
+                type: configuration.type,
+                data: configuration.data
```
iceiilin: Do we need to return the result here, which will be sent to those who sending the notification? I guess the response code can be used for the sender to decide whether they need a retry.

lib/api/1.1/southbound/notification.js
```diff
@@ -0,0 +1,40 @@
...
+var di = require('di'),
+    express = require('express');
+
+module.exports = notificationRouterFactory;
+
+di.annotate(notificationRouterFactory, new di.Provide('Http.Api.Internal.Notification'));
+di.annotate(notificationRouterFactory,
+    new di.Inject(
+        'Http.Services.RestApi',
+		'Http.Services.Api.Notification',
```
yyscamper: indent problem.

lib/api/1.1/southbound/templates.js
```diff
@@ -70,7 +74,7 @@ function templatesRouterFactory (
             presenter(req, res)
                 .renderTemplate(
                     req.params.identifier,
-                    properties.options,
+                    _.merge(properties.options, {identifier: nodeId}),
```
yyscamper: How about setting the key as `nodeId` rather than `identifier`, as the "identifier" is a so general word.

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,95 @@
...
+
+        return Promise.try(function() {
+            if (!message.nodeId || !_.isString(message.nodeId)) {
+                throw new Errors.BadRequestError('Missing node ID in query or body');
+            };
+            if (!message.type || !_.isString(message.type)) {
+                throw new Errors.BadRequestError('Missing notification type in query or body');
+            };
+            if (!message.data || !_.isString(message.data)) {
+                throw new Errors.BadRequestError('Missing notification data in query or body');
```
yyscamper: The empty string will throw error, is this expected?

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,95 @@
...
+            activeTaskId = activeTask.taskId;
+            activeGraphId = activeGraph.instanceId;
+
+            return eventsProtocol.publishTaskNotification(
+                nodeId,
+                activeTaskId,
+                activeGraphId,
+                message.type,
+                message.data
+                )
```
yyscamper: indent problem and forgot `;`

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,95 @@
...
+            // Check if the node exists
+            return waterline.nodes.needByIdentifier(nodeId)
+        })
+        .then(function () {
+            return [
+                taskGraphStore.findActiveGraphForTarget(nodeId),
+                taskProtocol.activeTaskExists(nodeId)
+            ]
+        })
+        .spread(function (activeGraph, activeTask) {
```
yyscamper: Since the activeGraph and activeTask is got by looking up the active task not passed from notification request.  There is a chance that the notification isn't the task wants.

Assume there is a Task A, it will create a thread in Microkernel, then it will post notification to tell RackHD it has finished.  In a case, the thread is hang for a long time (but actually it is still alive in microkernel), then RackHD forces to kill Task A, and retry. Another thread will be created in Microkernel, since previous thread is still alive, it finally success and sends the notification.  The new Task will receive the notification but actually this is not the notification it wants.

So I think it is still required to pass the taskId in notification request. If the taskId from notification doesn't match with the active task, then the notification should be ignored (we can log some warning here)

lib/services/notification-api-service.js
yyscamper: BadRequest means there is something wrong with notification request's input data, but not every case is BadRequest here. For example, mongo error causes fail to find active task, the task finishes unexpectedly.

lib/services/notification-api-service.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    notificationApiService.prototype.postNotification = function(message) {
+        var self = this;
+
+        return Promise.try(function() {
+            if (!message.taskId || !_.isString(message.taskId)) {
+                throw new Errors.BadRequestError('Missing task ID in query or body');
+            };
+            if (!message.data || !_.isString(message.data)) {
```
yyscamper: in on-core, you allowed the data  to be either object or string, but here you only allowed it to be string, I think they should be aligned.

static/monorail-2.0-sb.yaml
```diff
@@ -118,7 +118,48 @@ paths:
...
+      operationId: notificationPost
+      summary: |
+        post a notification
+      description: |
+        post a notification
+      parameters:
+        - name: taskId
+          in: query
+          description: |
+            node identifier
```
yyscamper: task instance identifier

static/monorail-2.0-sb.yaml
```diff
@@ -118,7 +118,48 @@ paths:
...
+          required: false
```
yyscamper: required: true, since you add assertion logic if the taskId is missing.

static/monorail-2.0-sb.yaml
```diff
@@ -118,7 +118,48 @@ paths:
...
+          type: string
+        - name: data
+          in: query
+          description: |
+            node identifier
```
yyscamper: The description is not correct.

static/monorail-2.0-sb.yaml
```diff
@@ -118,7 +118,48 @@ paths:
...
+          required: false
```
yyscamper: required: true, required: true, since you add assertion logic if the data is missing.

static/monorail-2.0-sb.yaml
```diff
@@ -118,7 +118,48 @@ paths:
...
+          type: string
```
yyscamper: string or object? or both?

static/monorail-2.0-sb.yaml
```diff
@@ -118,7 +118,48 @@ paths:
...
+          schema:
+            type: object
+        400:
+          description: |
+            bad request parameter passed or no active task or taskgraph.
+          schema:
+            $ref: '#/definitions/Error'
+        404:
+          description: |
+            Node not found.
```
yyscamper: Task not found?

lib/services/notification-api-service.js
cgx027: Good catch @yyscamper

cgx027:
Remove WIP flag.

lib/api/1.1/southbound/templates.js
```diff
@@ -10,12 +10,12 @@ module.exports = templatesRouterFactory;
 di.annotate(templatesRouterFactory, new di.Provide('Http.Api.Internal.Templates'));
 di.annotate(templatesRouterFactory,
     new di.Inject(
-		'Services.Waterline',
-		'Http.Services.Api.Workflows',
-		'Protocol.Task',
-		'Services.Lookup',
-		'Promise',
-		'common-api-presenter'
+	    'Services.Waterline',
```
WangWinson: Does this file need to update to support your PR?

WangWinson:
:+1:

yyscamper:
:+1:

yyscamper:
test this please

yyscamper:
test this please

PR #379, opened 2016-07-20
skus schema

BillyAbildgaard:
adding schema for /skus POST and PUT.

@RackHD/corecommitters @dalebremner @keedya @srinia6

### DIFF
--- a/static/schemas/2.0/skus.2.0.json
+++ b/static/schemas/2.0/skus.2.0.json
@@ -1,0 +1,248 @@
+{
+    "title": "Skus.2.0",
+    "definitions": {
+        "SkusRules": {
+            "description": "Possible Rules a Sku can use",
+            "type": "array",
+            "items": {
+                "type": "object",
+                "required": [
+                    "path"
+                ],
+                "additionalProperties": false,
+                "properties": {
+
+                    "empty": {
+                        "type": "string"
+                    },
+                    "notEmpty": {
+                        "type": "string"
+                    },
+                    "protected": {
+                        "type": "string"
+                    },
+                    "undefined": {
+                        "type": "string"
+                    },
+                    "object": {
+                        "type": "string"
+                    },
+                    "json": {
+                        "type": "string"
+                    },
+                    "geojson": {
+                        "type": "string"
+                    },
+                    "geometry": {
+                        "type": "string"
+                    },
+                    "mediumtext": {
+                        "type": "string"
+                    },
+                    "longtext": {
+                        "type": "string"
+                    },
+                    "text": {
+                        "type": "string"
+                    },
+                    "string": {
+                        "type": "string"
+                    },
+                    "binary": {
+                        "type": "string"
+                    },
+                    "boolean": {
+                        "type": "string"
+                    },
+                    "array": {
+                        "type": "string"
+                    },
+                    "number": {
+                        "type": "string"
+                    },
+                    "falsey": {
+                        "type": "string"
+                    },
+                    "null": {
+                        "type": "string"
+                    },
+                    "equals": {
+                        "type": "string"
+                    },
+                    "greaterThan": {
+                        "type": "string"
+                    },
+                    "lessThan": {
+                        "type": "string"
+                    },
+                    "path": {
+                        "type": "string"
+                    },
+                    "after": {
+                        "type": "string",
+                        "format": "date"
+                    },
+                    "alpha": {
+                        "type": "string"
+                    },
+                    "alphadashed": {
+                        "type": "string"
+                    },
+                    "alphanumeric": {
+                        "type": "string"
+                    },
+                    "alphanumericdashed": {
+                        "type": "string"
+                    },
+                    "before": {
+                        "type": "string",
+                        "format": "date"
+                    },
+                    "contains": {
+                        "type": "string"
+                    },
+                    "creditcard": {
+                        "type": "string",
+                        "format": "credit-card-number"
+                    },
+                    "datetime": {
+                        "type": "string",
+                        "format": "date-time"
+                    },
+                    "decimal": {
+                        "type": "string"
+                    },
+                    "email": {
+                        "type": "string",
+                        "format": "email"
+                    },
+                    "finite": {
+                        "type": "string"
+                    },
+                    "float": {
+                        "type": "string"
+                    },
+                    "hexadecimal": {
+                        "type": "string"
+                    },
+                    "hexColor": {
+                        "type": "string"
+                    },
+                    "in": {
+                        "type": "string"
+                    },
+                    "int": {
+                        "type": "string"
+                    },
+                    "integer": {
+                        "type": "string"
+                    },
+                    "ip": {
+                        "type": "string"
+                    },
+                    "ipv4": {
+                        "type": "string",
+                        "format": "ipv4"
+                    },
+                    "ipv6": {
+                        "type": "string"
+                    },
+                    "is": {
+                        "type": "string"
+                    },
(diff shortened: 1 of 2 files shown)


static/schemas/2.0/skus.2.0.json
```diff
@@ -0,0 +1,248 @@
+{
+    "title": "Skus.2.0",
+    "definitions": {
+        "SkusRules": {
```
yyscamper: This is all the rules that list in the docs (http://rackhd.readthedocs.io/en/latest/rackhd2.0/skus.html?highlight=contains#sku-json-format), your code doesn't match with it

static/schemas/2.0/skus.2.0.json
```diff
@@ -0,0 +1,248 @@
...
+                "rules": {
+                    "$ref": "#/definitions/SkusRules"
+                },
+                "discoveryGraphName": {
+                    "type": "string"
+                },
+                "discoveryGraphOptions": {
+                    "type": "object",
+                    "properties": {
+                        "username": {
```
yyscamper: The `discoveryGraphOptions` can be any object, the `properties` here should be empty.

static/schemas/2.0/skus.2.0.json
```diff
@@ -0,0 +1,248 @@
+{
+    "title": "Skus.2.0",
+    "definitions": {
+        "SkusRules": {
+            "description": "Possible Rules a Sku can use",
+            "type": "array",
+            "items": {
+                "type": "object",
```
yyscamper: The rule object isn't correct, firstly there should be a property for required `path`, but you forgot it; Then the rule words can only pick one of it or no rule word, the correct example should be:
```
"items": {
    "type": "object",
    "required": ["path"],
    "oneOf": [
        {
            "maxProperties": 1,
            "minProperties": 1
        },
        {
            "maxProperties": 2,
            "minProperties": 2
        }
    ],
    "additionalProperties": false,
    "properties": {
        "path": {
            "type": "string",
            "minLength": 1
        },
        "contains": {
            "type": "string"
        },
        "greaterThan": {
            "type": "string"
        },
        "minLength": {
            "type": "string"
        },
        {
            ...
        }
    }
}
```

static/schemas/2.0/skus.2.0.json
```diff
@@ -0,0 +1,248 @@
...
+                        "type": "string"
+                    }
+                }
+            }
+        },
+        "SkusUpsert": {
+            "description": "A sku for RackHD",
+            "type": "object",
+            "required": [
+                "name"
```
yyscamper: The `rules` is required as well.  See https://github.com/RackHD/on-core/blob/master/lib/models/sku.js#L52

static/schemas/2.0/skus.2.0.json
```diff
@@ -0,0 +1,248 @@
...
+            ],
+            "properties": {
+                "name": {
+                    "type": "string"
```
yyscamper: Add a `"minLength": 1` to ensure the name should not be empty.

static/schemas/2.0/skus.2.0.json
```diff
@@ -0,0 +1,248 @@
...
+                },
+                "rules": {
+                    "$ref": "#/definitions/SkusRules"
+                },
+                "discoveryGraphName": {
+                    "type": "string"
```
yyscamper: Add a `"minLength": 1` to ensure the name should not be empty.

yyscamper:
Set :-1: for current code, my comment need be fixed.

static/schemas/2.0/skus.2.0.json
BillyAbildgaard: @yyscamper I can change the list of accepted rules no problem.  

I was pointed to the following link to show the rules we use from anchor
https://github.com/sailsjs/anchor/blob/cbbc3a51f0b400d617a2def2002ce1c2bf210a4b/lib/match/rules.js

maybe @benbp can comment, but is there a chance we support all the anchor rules and our documentation just isn't updated?

static/schemas/2.0/skus.2.0.json
BillyAbildgaard: @yyscamper just to be clear, are you saying "discoveryGraphOptions" can also be a string, number, etc.?

static/schemas/2.0/skus.2.0.json
BillyAbildgaard: @yyscamper if you look at lines 9-11 I have the required path part you are talking about

BillyAbildgaard: I didn't know I could set mins and maxes for properties.  Thanks for the tip!

static/schemas/2.0/skus.2.0.json
BillyAbildgaard: addressed in my latest commit "addressing comments"

static/schemas/2.0/skus.2.0.json
BillyAbildgaard: @yyscamper added this

static/schemas/2.0/skus.2.0.json
yyscamper: No, discoveryGraphOptions should be object, but its properties can be any. You shouldn't preinstall `username` `password` `hostname` to the discoveryGraphOptions.

BillyAbildgaard: They aren't required properties so that shouldn't really matter because users can add or not use those fields. 
I just figured they are commonly used so I'd specify if they're added they would need to be a string.

static/schemas/2.0/skus.2.0.json
yyscamper: @BillyAbildgaard : Just see the `path` property, it is hidden in so many rules words, I overlook it.

yyscamper: @BillyAbildgaard : You can see the full supported keyword for each type from this link: http://json-schema.org/latest/json-schema-validation.html

static/schemas/2.0/skus.2.0.json
yyscamper: Though it's not required, but you are trying to restrict their type.  I know usually the `username` `password` `hostname` is string, but it's free to customer to define the `hostname` as number in its customized discovery graph. this is not the scope for this general SKU schema.

static/schemas/2.0/skus.2.0.json
yyscamper: I didn't know the history, maybe we intentionally wanted to pick a portion, @benbp, could you please confirm?

static/schemas/2.0/skus.2.0.json
yyscamper: I think you forgot to update here

yyscamper:
Thanks @BillyAbildgaard , change to :+1: after @benbp confirm the supported Rule operator.

BillyAbildgaard:
test this please

static/schemas/2.0/skus.2.0.json
benbp: @BillyAbildgaard I think it makes sense to enforce what's in the documentation with the schema, rather than the full list (since a lot of those validations aren't really applicable to this task anyways).

BillyAbildgaard: @benbp @yyscamper edited rules list to match those listed in documentation and specified types

BillyAbildgaard:
test this please

yyscamper:
:+1:

anhou:
:+1:

PR #381, opened 2016-07-22
Fix API 2.0 node view

anhou:
Fix bug for PR https://github.com/RackHD/on-http/pull/378 

Errors:
```
 {"message":"Unexpected token }","status":"500","stack":"Error: Unexpected token }    
at _processError (/home/vagrant/src/on-http/lib/services/swagger-api-service.js:45:19)   
at /home/vagrant/src/on-http/lib/services/swagger-api-service.js:219:22  
```

I think it's time to write render unit tests for profile/templates/view/...

@RackHD/corecommitters @srinia6 @jlongever

PR #384, opened 2016-07-22
Add redfish-obm-service schema and fix node view

jlongever:
- Create Redfish OBM service schema file.
- Fix nodes view template to handle a target list type.

@RackHD/corecommitters @zyoung51 @tannoa2 @uppalk1 @keedya @BillyAbildgaard @srinia6 @dalebremner @geoff-reid

@brianparry @srinia6: One of the changes below changes the relation target value.  Not all relationship values could be a reference to another resource,  for example Redfish endpoint relations:
```
       "targets": [
          "ComputeElement0",
          "StorageElement1"
        ],
        "relationType": "elementEndpoints"
      }
``` 

Was the intention to make all properties a reference link?

### DIFF
--- a/static/schemas/obms/redfish-obm-service.json
+++ b/static/schemas/obms/redfish-obm-service.json
@@ -1,0 +1,57 @@
+{
+    "title": "redfish-obm-service",
+    "definitions": {
+        "Obm": {
+            "description": "Redfish OBM settings",
+            "type": "object",
+            "properties": {
+                "nodeId": {
+                    "type": "string"
+                },
+                "service": {
+                    "type": "string"
+                },
+                "config": {
+                    "type": "object",
+                    "properties": {
+                        "host": {
+                            "description": "Redfish address",
+                            "type": "string"
+                        },
+                        "user": {
+                            "description": "Redfish username",
+                            "type": "string"
+                        },
+                        "password": {
+                            "description": "Redfish password",
+                            "type": "string"
+                        },
+                        "verfiySSL": {
+                            "description": "Validate SSL/TLS certs on client request",
+                            "type": "boolean"
+                        },
+                        "protocol": {
+                            "description": "Request protocol",
+                            "type": "string",
+                            "enum": ["http", "https"]
+                        },
+                        "port": {
+                            "description": "Optional http(s) host port, default 80/443",
+                            "type": "string"
+                        },
+                        "root": {
+                            "description": "The Redfish resources target path",
+                            "type": "string"
+                        },
+                        "uri": {
+                            "description": "The Redfish endpoint address",
+                            "type": "string"
+                        }
+                    },
+                    "required": ["host", "user", "password", "root", "uri", "protocol"]
+                }
+            },
+            "required": ["nodeId", "service", "config"]
+        }
+    }
+}
--- a/data/views/node.2.0.json
+++ b/data/views/node.2.0.json
@@ -32,10 +32,12 @@
         <% } else { %>
         "info": <%- JSON.stringify(value.info) %>,
         <% } %>
+        "targets": [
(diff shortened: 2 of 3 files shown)


yyscamper:
@anhou also submitted a PR (https://github.com/RackHD/on-http/pull/381) to fix node view, but had different solution. The API 2.0 style prefers link rather than raw identifier, so I'd prefer @anhou 's fix.

@brianparry

srinia6:
@jlongever : yes, we wanted to make all properties a reference link.  But didn't realize about "elementEndpoints".

jlongever:
@yyscamper How does @anhou's PR solve the problem that `elementEndpoints` targets are not a resource that can be referenced?  e.g. `/api/2.0/nodes/StorageElement1` is an invalid request.

jlongever:
test this please

anhou:
@jlongever I also didn't notice if the reference link is valid, just fix the syntax issue, I will close my PR. I‘m thinking, in previous implementation of relationType is for RackHD managed nodes, which is also assigned a mongoId(or other identifier), so maybe either a node is created for 'ComputeElement0' has a Id, or this info is not included in 'targets', just let them in catalogs?

jlongever:
@anhou there was a requirement to include endpoint elements as node relations for composed systems.

jlongever:
@anhou Also- generating new nodes for elements seems like an overkill.   
Typically (at least following Redfish conventions) there are designated `ref` properties for a resource.  Seems like 2.0 API applies refs links as the the resource value.

static/schemas/obms/redfish-obm-service.json
```diff
@@ -0,0 +1,57 @@
...
+            "properties": {
+                "nodeId": {
+                    "type": "string"
+                },
+                "service": {
+                    "type": "string"
+                },
+                "config": {
+                    "type": "object",
+                    "properties": {
```
yyscamper: Suggest adding an "additionalProperties: false" in order to prevent those extra and useless properties from being added. And also, if we add new setting in future, this will catch the error if developer forgots the schema definition of new setting

anhou:
@jlongever Thanks for clarification! agree 'generating new nodes for elements seems like an overkill.' so although it's a requirements, personally, I prefer to include endpoint elements in catalogs.  Or maybe we have to extend original thoughts of 'targets' (at least for me) that it's for RackHD manged nodes. just some thoughts :) not a comment for this PR
yeah, redfish/odata convention is well designed. API 2.0 could refer to it
:+1: for me

yyscamper:
@jlongever : How to distinguish the relation node is RackHD generated node or from elementEndpoints?  I think a raw identifier is hard to distinguish.

jlongever:
@yyscamper They can be distinguished by the Chassis (enclosure) node relation.    i.e. enclosed by this enclosure target with these elements.

yyscamper:
@jlongever : enclosure relation also applies for RackHD generated nodes, as the discovery-workflow will auto generate enclosure nodes. So I think this is still not enough for distinguishability.

jlongever:
@yyscamper there might not be any relations since no Systems have been composed yet so the enclosure relation tasks won't set those up.

I don't understand your concern - a compute node will have an `enclosedBy` target and `elementEndpoint` targets.   It's implied those `elementEndpoint` targets are from the `enclosedBy` target because the Chassis (enclosure node) resource provides the element catalog.

yyscamper:
@jlongever : I didn't realize the difference of relationType.  I understand it right now. Anyway, To convert the relation target to link should still be the right direction for API 2.0 though we cannot resolve all challenge now. That's not the scope of your PR. We can move the discussion to otherwhere.

For this PR is :+1:  with a nice-to-have suggestion to add "additionProperties: false"

PR #387, opened 2016-07-26
Fix ODR-800: RackHD Redfish v1 - bad link returned in /redfish/v1/Sys…

zyoung51:
…tems/<nodeid>/LogServices/SEL/Entries/NN


anhou:
:+1:

PR #389, opened 2016-07-27
YAML return code corrections

srinia6:
Update YAML to reflect correct return codes. Also, make the return code consistent across the API's as follows:
GETs/PATCH - 200
PUT/POST - 201
DELETE - 204 [No content]
With an exception of PUT - /workflows/{identifier}/action, /nodes/{identifier}/workflows/action and POST /tags/{tagName}/nodes/workflows that returns a 202.

 ALL unit-tests pass locally as this PR has unit-test updates as well.

@benbp @keedya @BillyAbildgaard @brianparry @dalebremner @RackHD/corecommitters @zyoung51 @jlongever

### DIFF
--- a/lib/api/2.0/files.js
+++ b/lib/api/2.0/files.js
@@ -11,7 +11,7 @@ var filesGet = controller(function(req, res) {
     return fileService.getFile(req, res, req.swagger.params.fileidentifier.value);
 });
 
-var filesPut = controller(function(req) {
+var filesPut = controller({success: 201}, function(req) {
     return fileService.putFile(req, req.swagger.params.fileidentifier.value);
 });
 
--- a/lib/api/2.0/lookups.js
+++ b/lib/api/2.0/lookups.js
@@ -88,7 +88,7 @@ var lookupsPatchById = controller(function(req) {
  *       "error": "Not Found"
  *      }
  */
-var lookupsDelById = controller(function (req) {
+var lookupsDelById = controller({success: 204}, function (req) {
     return waterline.lookups.destroyOneById(req.swagger.params.id.value);
 });
 
--- a/lib/api/2.0/nodes.js
+++ b/lib/api/2.0/nodes.js
@@ -26,7 +26,7 @@ var nodesPatchById = controller(function(req) {
     return nodes.patchNodeById(req.swagger.params.identifier.value, req.body);
 });
 
-var nodesDelById = controller(function(req) {
+var nodesDelById = controller({success: 204}, function(req) {
     return nodes.delNodeById(req.swagger.params.identifier.value);
 });
 
@@ -106,7 +106,7 @@ var nodesGetTagsById = controller(function(req) {
     return nodes.getTagsById(req.swagger.params.identifier.value);
 });
 
-var nodesDelTagById = controller(function(req) {
+var nodesDelTagById = controller({success: 204}, function(req) {
     return nodes.removeTagsById(req.swagger.params.identifier.value,
                                 req.swagger.params.tagName.value);
 });
@@ -116,7 +116,7 @@ var nodesPatchTagById = controller(function(req) {
                              req.body.tags);
 });
 
-var nodesMasterDelTagById = controller(function(req) {
+var nodesMasterDelTagById = controller({success: 204}, function(req) {
     return nodes.masterDelTagById(req.swagger.params.tagName.value);
 });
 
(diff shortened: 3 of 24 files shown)


srinia6:
Functional Test PR link - https://github.com/RackHD/RackHD/pull/342

yyscamper:
I have some concern about the 204 that returned after delete request. I did some search and found the 204 means "No Content", it is reasonable if there is no data in response body. But nearly all RackHD DELETE request will response the deleted data, so actually there is content in the response, if so, I think `200` will be more reasonable.

Below is the a part of 204 description in w3c.org (https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.5):
```
The 204 response MUST NOT include a message-body, and thus is always terminated by the first empty line after the header fields.
```

Add @benbp @zyoung51 @jlongever @WangWinson

WangWinson:
Agree with @yyscamper , if we return the entity when DELETE it. see method definition from [w3c.org](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

>
9.7 DELETE
>
The DELETE method requests that the origin server delete the resource identified by the Request-URI. This method MAY be overridden by human intervention (or other means) on the origin server. The client cannot be guaranteed that the operation has been carried out, even if the status code returned from the origin server indicates that the action has been completed successfully. However, the server SHOULD NOT indicate success unless, at the time the response is given, it intends to delete the resource or move it to an inaccessible location.
>
A successful response SHOULD be **200 (OK)** if the response **includes an entity** describing the status, 202 (Accepted) if the action has not yet been enacted, or **204 (No Content)** if the action has been enacted but the response does **not include an entity**.
>
If the request passes through a cache and the Request-URI identifies one or more currently cached entities, those entries SHOULD be treated as stale. Responses to this method are not cacheable.

srinia6:
@yyscamper @WangWinson Thank you for the input. I understand that previously we did return data along with 200[OK] status code, and in API 2.0, two of the API's returns 204 on DELETE. 

My intention with this PR was to have a consistent response across. And as the PR itself mentions that 204 would return  (No Content), with this change, you will not see any returned data. 

I will wait for some inputs on this - whether to return 200/204 on a DELETE. Also there are many links that advocate that DELETE could return a 204 than a universal 200. Also, 200 is returned by GET/PATCH. Do we want the DELETE to return a 200 as well?

keedya:
:+1: I agree with @srinia6

heckj:
👍

yyscamper:
@srinia6 : This is the word that copied from you attached REST API tutorial:
```
DELETE is pretty easy to understand. It is used to **delete** a resource identified by a URI.

On successful deletion, return HTTP status 200 (OK) along with a response body, perhaps the representation of the deleted item (often demands too much bandwidth), or a wrapped response (see Return Values below). Either that or return HTTP status 204 (NO CONTENT) with no response body. In other words, a 204 status with no body, or the JSEND-style response and HTTP status 200 are the recommended responses.
```
It still recommends return 204 if no data in response body; 200 if there is data in response body.

If you are going to remove the data from response body, I am OK with it. If so the 204 is reasonable.

Anyway, I don't have much experience on web service data before, all my comment is from web. We can wait more people's comment.

@heckj : are you OK with 204 since you have approve it.

And what's your comment @zyoung51 @jlongever @benbp @anhou ?

WangWinson:
I agree that 204 is more suitable for DELETE as long as there is really **no content** in response. 
Can we ensure that all DELETE method will not return any data in these APIs ?  If so, I'm OK

zyoung51:
@yyscamper I doubt there is an absolute right answer here (perhaps only a chance for more bikeshedding @benbp).  

Assuming the deletion of the resource completed before issuing the response, a 204 and no data is best.  A 200 would make a consumer question if the resource was really deleted simply because we still have data to reply with.  A 202 would make a consumer question if the resource was deleted because we are telling them it isn't deleted yet.  A 204 conveys that the resource is gone and it is so gone that we can not provide data about it anymore.

jlongever:
+1

heckj:
yeah, I'm totally good with a `204` response here

srinia6:
test this please

srinia6:
The latest build failure is due to change in return codes.  And those are fixed by this PR - https://github.com/RackHD/RackHD/pull/342

benbp:
👍

PR #393, opened 2016-07-29
Check kargs object type before parsing

jlongever:
- Checks `kargs` option type is object before parsing to valid kernel boot string.  If non-object (i.e. string) will just be passed as-is.

@RackHD/corecommitters @uppalk1 @tannoa2 @zyoung51 @keedya


keedya:
:+1:

heckj:
👍

benbp:
:+1:

PR #395, opened 2016-08-02
apply update regardless of node being found

heckj:
# WIP 

from @jeamland - debugging thread related to `/lookups` API not accepting an update


heckj:
@benbp @brianparry FYI - from slack conversation with Benno

heckj:
Looks like our tests are very specific to the 1.1 API set - req.swagger.query vs. req.query

```
1) Http.Services.Api.Profiles setLookup setLookup should add IP lookup entry for new node:
     TypeError: Cannot read property 'query' of undefined
      at ProfileApiService.setLookup (lib/services/profiles-api-service.js:79:32)
      at Context.<anonymous> (spec/lib/services/profiles-api-service-spec.js:93:38)
  2) Http.Services.Api.Profiles setLookup setLookup should add IP lookup entry and proxy for new node:
     TypeError: Cannot read property 'query' of undefined
      at ProfileApiService.setLookup (lib/services/profiles-api-service.js:79:32)
      at Context.<anonymous> (spec/lib/services/profiles-api-service-spec.js:107:38)
  3) Http.Services.Api.Profiles setLookup setLookup does not add IP lookup entry for existing node:
     TypeError: Cannot read property 'query' of undefined
      at ProfileApiService.setLookup (lib/services/profiles-api-service.js:79:32)
      at Context.<anonymous> (spec/lib/services/profiles-api-service-spec.js:122:38)
  4) Http.Services.Api.Profiles setLookup setLookup does not lookup node on missing required query string:
     TypeError: Cannot read property 'query' of undefined
      at ProfileApiService.setLookup (lib/services/profiles-api-service.js:79:32)
      at Context.<anonymous> (spec/lib/services/profiles-api-service-spec.js:133:38)
```

lib/services/profiles-api-service.js
```diff
@@ -76,16 +76,14 @@ function profileApiServiceFactory(
     };
 
     ProfileApiService.prototype.setLookup = function(req) {
-        var query = req.query;
+        var query = req.swagger.query;
```
brianparry: @heckj I think we want to check whether the swagger object is defined in the request so that the profiles service will work with both the 1.1 and 2.0 APIs.  I think this will fix your test failure as well.
```
var query = req.swagger ? req.swagger.query : req.query;
```

heckj:
this flies in the face of some specific usage that was embedded in test cases - that a node wouldn't update if it was found (https://github.com/RackHD/on-http/blob/master/spec/lib/services/profiles-api-service-spec.js#L115-L128). @jlongever set this up originally - what would we be screwing up if we changed this logic to allow updates regardless of a node being found? Originally came from https://github.com/RackHD/on-http/pull/157 based on blame/history...

anhou:
@heckj In my plan, 'query.macs & query.ips' will replace 'query.mac & qeury.ip' ,  temporally, be compatible inmy submitted PR  https://github.com/RackHD/on-http/pull/411.  no matter node discovered or not.  lookup ip and mac will be updated.  on-dhcp-proxy also do like this https://github.com/RackHD/on-dhcp-proxy/blob/master/lib/message-handler.js#L61. @jlongever  

Anyway, :+1:  for me

benbp:
test this please

heckj:
I'm updating the unit tests to match the code change - another commit pending...

heckj:
er, happy to close this in favor of @anhou solution in https://github.com/RackHD/on-http/pull/411 - probably a better longer term solution than this update hack

panpan0000:
test this please

PR #399, opened 2016-08-03
partial doc

heckj:
partial doc fix for for https://github.com/RackHD/RackHD/issues/363

### DIFF
--- a/static/monorail.yml
+++ b/static/monorail.yml
@@ -650,34 +650,12 @@ paths:
           schema:
             $ref: '#/definitions/Error'
 
-  /workflows/library:
-    get:
-      summary: |
-        List all workflows available to run
-      description: |
-        List all workflows available to run
-      tags:
-        - workflow
-        - get
-      responses:
-        200:
-          description: |
-            List all workflows available to run
-          schema:
-            type: array
-            items:
-              type: object
-        default:
-          description: Unexpected error
-          schema:
-            $ref: '#/definitions/Error'
-
   /workflows/library/{injectableName}:
     get:
       summary: |
-        List all workflows available to run
+        Get a workflow definition
       description: |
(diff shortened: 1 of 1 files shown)


heckj:
@benbp This match the workflow update you made a while back in this respect?

static/monorail.yml
```diff
@@ -650,34 +650,12 @@ paths:
...
-            $ref: '#/definitions/Error'
-
   /workflows/library/{injectableName}:
     get:
       summary: |
-        List all workflows available to run
+        Get a workflow definition
       description: |
-        List all workflows available to run
+        Get a workflow definition. To list all available workflows, use the '*' as a wildcard for the name.
```
yyscamper: I think using the `*` wildcard is not a recommend, we need to fix the code so that /workflows/library can return all list workflow definitions.

heckj: actually it is - @benbp made changes several months ago, but it didn't propagate to this or the documentation, so this is the doc catchup. The 2.0 API made it work as you would normally expect, without a wildcard, but the 1.1 API had limited based on the URI structuring and path rendering that made that infeasible.

yyscamper: Make sense, considering the API 1.1 will be deprecated soon, it's not worthy to do some API level change. So I am OK to keep the wildcard for API 1.1 now.

yyscamper:
:+1:

static/monorail.yml
benbp: Yep, it was intended to be a short lived hack that would go away with the 1.1 API :)

benbp: It was some express framework issue where the route ordering in the code wasn't letting it work otherwise.

benbp:
👍

PR #363, opened 2016-07-12
Change Graph.InstallEsx to Graph.InstallESXi

anhou:
resolve RackHD/RackHD#218

@RackHD/corecommitters

PR #402, opened 2016-08-04
Issue#363: Fix the API response of /api/1.1/workflows/libary

yyscamper:
Resolves https://github.com/RackHD/RackHD/issues/363

To align the 1.1 & 2.0 APIs and fix the mismatching of getting workflow definition & task definition.

- Use `/api/1.1/workflows/library` to get all workflow definitions rather than `/api/1.1/workflows/library/*`

- Use `/api/1.1/workflows/tasks/library` to get all task definitions rather than `/api/1.1/workflows/tasks/library/*`

- Add `/api/1.1/workflows/tasks/library/:injectableName` to get a specific task definition. This is to align the API `/api/1.1/workflows/library/:injectableName` to get a specific graph definition.

@RackHD/corecommitters @codenrhoden @heckj

### DIFF
--- a/lib/api/1.1/northbound/workflows.js
+++ b/lib/api/1.1/northbound/workflows.js
@@ -102,20 +102,51 @@ function workflowsRouterFactory (
     }));
 
     /**
+     * @api {get} /api/1.1/workflows/tasks/library/:injectableName GET /library/:injectableName
+     * @apiVersion 1.1.0
+     * @apiDescription get a specific task definition from the library.
+     * @apiParam {String} injectableName of task definition
+     * @apiName tasks-library-item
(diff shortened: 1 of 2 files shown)


yyscamper:
The smoke test need some update to reflect this change. If you agree on this solution, I will then try to update the smoke-test.

jlongever:
@yyscamper Is it necessary to maintain 1.1 changes seeing how it will be soon superseded by 2.0? Could we point issue #363 at the 2.0 API?

yyscamper:
@jlongever : I am OK either way. I still had the concern that it  is not a good time to change the 1.1 API at present.

PR #404, opened 2016-08-04
coreos install: indicate complete after write to disk

codenrhoden:
Right now, the CoreOS install task says its finished once the
cloud-config file is downloaded, but this really before the install
even kicks off. Add a line to the install script to hit the
completionUri right before the reboot so we can trigger off that
instead.

### DIFF
--- a/data/templates/install-coreos.sh
+++ b/data/templates/install-coreos.sh
@@ -2,4 +2,5 @@
 set -e
 curl -O http://<%=server%>:<%=port%>/api/current/templates/pxe-cloud-config.yml
 sudo coreos-install -d <%=installDisk%> -c pxe-cloud-config.yml -b <%=repo%>
+wget --spider http://<%=server%>:<%=port%>/api/current/templates/<%=completionUri%>
 sudo reboot


data/templates/install-coreos.sh
```diff
@@ -2,4 +2,5 @@
 set -e
 curl -O http://<%=server%>:<%=port%>/api/current/templates/pxe-cloud-config.yml
 sudo coreos-install -d <%=installDisk%> -c pxe-cloud-config.yml -b <%=repo%>
+wget --spider http://<%=server%>:<%=port%>/api/current/templates/<%=completionUri%>
```
heckj: not sure you want to use `--spider` here unless you're really trying to spider through that directory - the end result if you passed in a `completionUri` that represented a directory would be replicating that entire directory structure down (probably inadvertently) to the local disk at the root of the CoreOS setup.

I'm also not sure if the CoreOS install includes `wget`, although I know it includes curl. You might just instead have this be:

`curl -s http://<%=server%>:<%=port%>/api/current/templates/<%=completionUri%> >> /dev/null` 

so you quite out the logging for curl and explicitly throw away the results

codenrhoden: Thanks for the feedback!

I used `--spider` because that's what I saw here: https://github.com/RackHD/on-http/blob/master/data/templates/centos-ks#L216

Truth be told, there are lots of different ways I saw that being done. some with wget, some with curl, and all different variations of options.  However, `--spider` seemed like not a bad one, and I don't think it would download a directory. The help says `       --spider                    don't download anything`.  It just sees if the file is there.  But it sure might crawl a directory. I don't see any examples of completionUri being a directory in the code though.

I'm testing with "CoreOS stable (1068.8.0)", and it has wget. not sure about older versions.

codenrhoden: Happy to change it, though.  curl is more ubiquitous, and your method will work just fine! Let me know what you think.

yyscamper: @codenrhoden : @cgx027 just defined a special notification API to replace previous wget or curl. And we are planing to replace all completionUri with the new notification API.

Since you happened to touch these code, so could you please update the code to use the notification API?

yyscamper: The task definition (https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/install-coreos.js#L18) still need be changed, otherwise the install-coreos will still finish after the config file is downloaded. 

Since both completionUri and notification API mechanism co-exist before we completely migrate to notification API. So at present, if you want to use the notification API, you need change the `completionUri` in task definition to something arbitrary (usually "renasar-ansible.pub") that prevents the completionUri from being work.

Add @lanchongyizu for awareness since he is the owner to migrate all install-os to use new notification API.

yyscamper: @codenrhoden : Right now, either use notification API or schemaUri is OK for me.

codenrhoden: @yyscamper: just to address your most recent comment -- this PR is cross-referenced with https://github.com/RackHD/on-tasks/issues/307, so I did change that as well. :)

yyscamper:
:+1:

codenrhoden:
@yyscamper Thanks for the 👍 . I did have someone else point out in Slack there was the new notification API. I didn't switch to it because I already had this queued up, and (embarrasingly) I have some docker images that are built from an older fork, and I want be able to cherry-pick these changes into it.

However, I am also happy to take on a follow-on PR to switch this stuff to the new API. Would that be okay? It seems like you're now okay with this change and the corresponding one in `on-tasks`, is that right?

heckj:
@codenrhoden actually - hold off on the notification API (that was Felix BTW)

I think we may have identified a flaw in the design in the past day that would only lead to frustration if you tried to actively use it now. @cgx027 and @yyscamper are poking at some possible solutions, so it may make a lot more sense to follow up and fix this up to be consistent later

codenrhoden:
@heckj: Roger that. Then I *think* the only outstanding issue is whether you want me to switch from `wget` to `curl`. If it's fine as-is, I think both PRs are good. (yay for less work!)

heckj:
Good by me - you answered my question on the `--spider` option, so no concerned there. We'll get another @RackHD/corecommitters  to do a review in here and be good to go!

anhou:
:+1:

yyscamper:
jenkins: test this please

PR #407, opened 2016-08-05
Set routes current and common to 2.0 by default

keedya:
@amymullins @heckj @RackHD/corecommitters @dalebremner @srinia6 @brianparry

### DIFF
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -220,7 +220,7 @@ function httpServiceFactory(
         );
 
         //re-route common and current
-        var versionPath = configuration.get('versionBase') || '2.0';
+        var versionPath = configuration.get('versionBase', '2.0');
         app.use(rewriter('/api/current/*', '/api/' + versionPath + '/$1'));
         app.use(rewriter('/api/common/*', '/api/' + versionPath + '/$1'));
 
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -220,7 +220,7 @@ function httpServiceFactory(
         );
 
         //re-route common and current
-        var versionPath = configuration.get('versionBase') || '1.1';
+        var versionPath = configuration.get('versionBase') || '2.0';
         app.use(rewriter('/api/current/*', '/api/' + versionPath + '/$1'));
         app.use(rewriter('/api/common/*', '/api/' + versionPath + '/$1'));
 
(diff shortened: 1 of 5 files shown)


keedya:
@cgx027, I had to modify unit test to make sure it works regardless to the api version
Please take a look at this PR to make sure I have it all covered.

dalebremner:
+1

lib/services/http-service.js
```diff
@@ -220,7 +220,7 @@ function httpServiceFactory(
         );
 
         //re-route common and current
-        var versionPath = configuration.get('versionBase') || '1.1';
+        var versionPath = configuration.get('versionBase') || '2.0';
```
benbp: I recommend switching to use the standard way we do configurations defaults (which obviously happened after the original code here was written):

`configuration.get('versionBase', '2.0')`

benbp:
👍

spec/lib/services/notification-api-service-spec.js
```diff
@@ -36,86 +37,42 @@ describe('Http.Api.Notification', function () {
...
-            .post(
-                '/api/current/notification?taskId='
-                + notificationMessage.taskId 
-                + '&data='
-                + notificationMessage.data)
-            .set('Content-Type', 'application/json')
-            .expect('Content-Type', /^application\/json/)
-            .expect(201, notificationMessage);
+            return notificationApiService.postNotification(notificationMessage)
+                        .then(function (resp) {
```
yyscamper: indent problem

spec/lib/services/notification-api-service-spec.js
```diff
@@ -36,86 +37,42 @@ describe('Http.Api.Notification', function () {
...
+                            console.log(resp);
```
yyscamper: Avoid log some useless info

yyscamper:
@keedya : Please revert all file permissions from 755 to 644

lib/services/http-service.js
yyscamper: Also my recommendation, this is the common way we handle the default value before.

cgx027:
Hi @keedya , the modification looks good to me. There are background talks between @benbp @heckj @yyscamper and myself on refactoring the notification api, so that I will need to push another PR to change the design in coming days. I do not see a reason to block this PR anyway.

yyscamper:
:+1: with some minor suggestion.

anhou:
:+1:  with comments fix

PR #410, opened 2016-08-09
notification API design refactor

cgx027:
Notification API design refactor:
* Route set to /notification
* In most cases, a notification will send to a node, so the API will accept a nodeId parameter, either in query string or request body, both POST /notification?nodeId=<nodeId> and POST /notification –d ‘{“nodeId”: ”<nodeId>”}’ will work.
  * The notification will be routed by AMQP with key “notification.<nodeId>”
* If the notification is send to other identities, say a workflow, the request will look like POST /notification?workflowId=<workflowId> or POST /notification –d ‘{“workflowId”: ”< workflowId >”}’. 
  * The notification will be routed by AMQP with key “notification.<workflowId >”
  * I will not code for this and leave it till the need comes.
* If no Id are specified in the parameter, the notification will be route by AMQP with key ‘notification’. This is kind of like a broadcast notification.
* All parameters including the Id are optional, and are defined and interpreted by the notification sender and receiver.  This design is meant to leave flexibility to the applications that want to send/receive notification
  * In the Isilon case where a progress needs to be sent, suggested parameters would look like:
    * Progress: the progress to 
    * Message: the additional message to send. 

This PR depends on: 
* https://github.com/RackHD/on-core/pull/191
* https://github.com/RackHD/on-tasks/pull/311

### DIFF
--- a/lib/api/1.1/southbound/notification.js
+++ b/lib/api/1.1/southbound/notification.js
@@ -33,16 +33,7 @@ function notificationRouterFactory (
 
     router.post('/notification/', rest(function (req) {
         var message = _.defaults(req.query || {}, req.body || {});
-
-        if (_.has(message, 'nodeId')) {
-            return notificationApiService.postNodeNotification(message);
-        }
-        // Add other cases here if to support more notification types
-
-        // This will be a broadcast notification if no id (like nodeId) is specified
-        else {
-            return notificationApiService.postBroadcastNotification(message);
-        }
+        return notificationApiService.postNotification(message);
     }, {renderOptions: {success: 201}}));
 
     return router;
--- a/lib/api/1.1/southbound/notification.js
+++ b/lib/api/1.1/southbound/notification.js
@@ -33,7 +33,16 @@ function notificationRouterFactory (
 
     router.post('/notification/', rest(function (req) {
         var message = _.defaults(req.query || {}, req.body || {});
-        return notificationApiService.postNotification(message);
+
+        if (_.has(message, 'nodeId')) {
+            return notificationApiService.postNodeNotification(message);
+        }
+        // Add other cases here if to support more notification types
+
+        // This will be a broadcast notification if no id (like nodeId) is specified
+        else {
+            return notificationApiService.postBroadcastNotification(message);
+        }
     }, {renderOptions: {success: 201}}));
 
     return router;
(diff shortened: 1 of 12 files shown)


cgx027:
Remove WIP flag

cgx027:
@RackHD/corecommitters @yyscamper @anhou @WangWinson

lib/api/1.1/southbound/notification.js
```diff
@@ -33,7 +33,16 @@ function notificationRouterFactory (
 
     router.post('/notification/', rest(function (req) {
         var message = _.defaults(req.query || {}, req.body || {});
-        return notificationApiService.postNotification(message);
+
+        if (_.has(message, 'nodeId')) {
```
WangWinson: Though the logic looks simple here. 
Will it be better to move it inside `notificationApiService` ? and expose a general `postNotification` function here ?

static/monorail-2.0-sb.yaml
```diff
@@ -127,18 +127,19 @@ paths:
...
+        - name: nodeId
           in: query
           description: |
-            task instance identifier
+            Node instance identifier
           required: false
           type: string
-        - name: data
-          in: query
+        - name: notification_body
```
WangWinson: `notification_data` better ?

static/monorail-2.0-sb.yaml
```diff
@@ -127,18 +127,19 @@ paths:
...
+          in: body
           description: |
-            notification data
+            task instance identifier
```
WangWinson: This description need to be updated

lib/api/2.0/notification.js
```diff
@@ -8,8 +8,17 @@ var notificationApiService = injector.get('Http.Services.Api.Notification');
 var _ = injector.get('_');    // jshint ignore:line
 
 var notificationPost = controller({success: 201}, function(req) {
-    var message = _.defaults(req.swagger.query || {}, req.body || {});
-    return notificationApiService.postNotification(message);
+    var message = _.defaults(req.swagger.query || {}, req.query || {}, req.body || {});
+
+    if (_.has(message, 'nodeId')) {
+            return notificationApiService.postNodeNotification(message);
+        }
```
yyscamper: indent

static/monorail-2.0-sb.yaml
yyscamper: Even may be use the `data` for the sake of simplicity?

cgx027: OK, use data instead.

lib/api/1.1/southbound/notification.js
cgx027: Sounds good, @WangWinson , updated.

WangWinson:
:+1:

yyscamper:
:+1:

yyscamper:
jenkins: test this please

PR #411, opened 2016-08-09
Get profiles with macs and ips

anhou:
PR for proposal https://groups.google.com/forum/#!topic/rackhd/I81Tl2qkQMI
related on-dhcp-proxy PR  https://github.com/RackHD/on-dhcp-proxy/pull/52

changes:
1. support geting profiles with macs and ips instead of 'macs' and 'mac & ip' for static ip address for both 1.1 and 2.0 GET /profiles route
2. remove duplicated profile rendering for new node 'node.newRecord'
3. refine naming of 'renderProfileFromTaskOrNode' to 'getProfileFromTaskOrNode'
4. treat 'no active workflow and boot settings' as  expected information.
5. remove error.zt to be consistent with @benbp PR https://github.com/RackHD/on-http/pull/376  and also be consistent with 2.0 GET /profiles API
6. fix 2.0 profile service's 'self.res.xxxx' to 'res.xxxx'

@RackHD/corecommitters  @benno  @jlongever

### DIFF
--- a/lib/api/1.1/southbound/profiles.js
+++ b/lib/api/1.1/southbound/profiles.js
@@ -58,7 +58,7 @@ function profilesRouterFactory (
                     });
                 });
             } else {
-                presenter(req, res).renderProfile('redirect.ipxe', { isLookUpNotChecked: true });
+                presenter(req, res).renderProfile('redirect.ipxe', { ignoreLookup: true });
             }
         })
         .catch(function (err) {
--- a/lib/api/1.1/southbound/profiles.js
+++ b/lib/api/1.1/southbound/profiles.js
@@ -52,17 +52,13 @@ function profilesRouterFactory (
 
                 return profileApiService.getNode(macAddresses, options)
                 .then(function (node) {
-                    if (node.newRecord) {
-                        presenter(req, res).renderProfile('redirect.ipxe');
-                    } else {
-                        return profileApiService.renderProfileFromTaskOrNode(node, 'compute')
-                        .then(function (render) {
-                            presenter(req, res).renderProfile(render.profile, render.options, 200, render.context);
-                        });
-                    }
+                    return profileApiService.getProfileFromTaskOrNode(node, 'compute')
+                    .then(function (render) {
+                        presenter(req, res).renderProfile(render.profile, render.options, 200, render.context);
+                    });
                 });
             } else {
-                presenter(req, res).renderProfile('redirect.ipxe');
+                presenter(req, res).renderProfile('redirect.ipxe', { isLookUpNotChecked: true });
             }
         })
         .catch(function (err) {
@@ -88,7 +84,7 @@ function profilesRouterFactory (
 
         profileApiService.getNode(macAddresses, options)
         .then(function(node) {
-            return profileApiService.renderProfileFromTaskOrNode(node, 'switch');
+            return profileApiService.getProfileFromTaskOrNode(node, 'switch');
         })
         .then(function(render) {
             presenter(req, res).renderProfile(
--- a/lib/services/profiles-api-service.js
+++ b/lib/services/profiles-api-service.js
@@ -76,7 +76,7 @@ function profileApiServiceFactory(
     };
 
     ProfileApiService.prototype.setLookup = function(req) {
-        var query = req.query;
+        var query = req.swagger ? req.swagger.query : req.query;
         if (query.mac && query.ip) {
             return waterline.nodes.findByIdentifier(this.getMacs(query.mac))
             .then(function (node) {
@@ -103,7 +103,7 @@ function profileApiServiceFactory(
             var macAddresses = _.flattenDeep([query.macs]);
             var ipAddresses = _.flattenDeep([query.ips]);
             return Promise.map(ipAddresses, function(ip, index) {
-                if (ip) {
+                if (ip && macAddresses[index]) {
                     return lookupService.setIpAddress(ip, macAddresses[index]);
                 }
             });
@@ -344,7 +344,7 @@ function profileApiServiceFactory(
 
         var promises = [
             swaggerService.makeRenderableOptions(req, res, graphContext,
-                    profile.isLookUpNotChecked),
+                    profile.ignoreLookup),
             profiles.get(profile.profile, true, scope)
         ];
 
@@ -380,7 +380,7 @@ function profileApiServiceFactory(
                                 });
                         });
                 } else {
-                    return { profile: 'redirect.ipxe', isLookUpNotChecked: true };
+                    return { profile: 'redirect.ipxe', ignoreLookup: true };
                 }
             })
             .catch(function (err) {
--- a/lib/services/profiles-api-service.js
+++ b/lib/services/profiles-api-service.js
@@ -98,6 +98,17 @@ function profileApiServiceFactory(
                 }
             });
         }
+
+        if (query.macs && query.ips) {
+            var macAddresses = _.flattenDeep([query.macs]);
+            var ipAddresses = _.flattenDeep([query.ips]);
+            return Promise.map(ipAddresses, function(ip, index) {
+                if (ip) {
+                    return lookupService.setIpAddress(ip, macAddresses[index]);
+                }
+            });
+        }
+
         return Promise.resolve();
     };
 
@@ -251,7 +262,7 @@ function profileApiServiceFactory(
         throw err;
     };
 
-    ProfileApiService.prototype.renderProfileFromTaskOrNode = function(node) {
+    ProfileApiService.prototype.getProfileFromTaskOrNode = function(node) {
         var self = this;
         var defaultProfile;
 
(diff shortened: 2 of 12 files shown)


lib/services/profiles-api-service.js
```diff
@@ -98,6 +98,17 @@ function profileApiServiceFactory(
                 }
             });
         }
+
+        if (query.macs && query.ips) {
+            var macAddresses = _.flattenDeep([query.macs]);
+            var ipAddresses = _.flattenDeep([query.ips]);
+            return Promise.map(ipAddresses, function(ip, index) {
+                if (ip) {
+                    return lookupService.setIpAddress(ip, macAddresses[index]);
```
benbp: Does this call succeed if `macAddresses[n]` is null (I think it will, just double checking)?

lib/services/profiles-api-service.js
```diff
@@ -315,11 +326,12 @@ function profileApiServiceFactory(
                                 'Unable to retrieve valid node bootSettings', node.type);
                     }
                 } else {
-                    return self._handleProfileRenderError(
-                        'Unable to locate active workflow or there are no bootSettings',
-                        node.type,
-                        400
-                    );
+                    return {
+                        profile: 'ipxe-info.ipxe',
```
benbp: I think this could end up sending an ipxe script to a switch? Or maybe this code path already had that problem?

benbp:
A couple minor questions/potential issues. 👍 otherwise

lib/services/profiles-api-service.js
anhou: yes, if mac address is null, it will break in ipxe profile, then to chainload, double checking is also fine, it's added

lib/services/profiles-api-service.js
anhou: it keeps the code logic with before. not sure about original switch purpose . from original code logic  perspective, I think it's fine here, if any concerns, let me update

lib/services/profiles-api-service.js
```diff
@@ -98,6 +98,17 @@ function profileApiServiceFactory(
                 }
             });
         }
```
heckj: I think you'll need the same update I made on line 79:

`var query = req.swagger ? req.swagger.query : req.query;`

to get the query properly from both the 1.1 and 2.0 API for this function

spec/lib/api/2.0/profiles-spec.js
```diff
@@ -233,19 +233,14 @@ describe('Http.Api.Profiles', function () {
...
-            var errorMsg = "Error: Unable to locate active workflow or there are no bootSettings";
-
             return helper.request().get('/api/2.0/profiles')
                 .query({ macs: '00:00:de:ad:be:ef' })
-                .expect(400)
-                .then(function(resp) {
-                    expect(resp.body.message).to.equal(errorMsg);
-                });
+                .expect(200);
         });
```
heckj: can you add in the tests which invoke the `/profiles` API here with the "macs" and "ips" structure to validate how that could be called?

heckj:
Would like to see a small addition to this so that the details of how to call the updated `/profiles` API mechanism with "macs" and "ips" is included in documentation somewhere so that it's clear how to integrate with it when you're someone trying to integrate our code to a separate DHCP setup (aka what @jeamland is doing)

brianparry:
@anhou static/monorail-2.0.yaml needs to be updated to include the `ips` query parameter.

anhou:
@heckj @brianparry @benbp Thanks for comments, I'll address all comments.
@heckj Yes,  I'll refine or refactor the docs http://rackhd.readthedocs.io/en/latest/rackhd/dhcp_server.html#dhcp-server-setup, and it's a little out of date,  I also need some validations in real/virtual environments. before updating docs. and also address all kinds of cases that RackHD could support such as how to intergrate, it will be listed in a table.  due to busy for tftp scaling tests.  maybe no time to finish it this sprint together with this PR.   And after docs are updated, I'll publish to google groups to let more folks know like @jeamland

By the way, Just realized Benno is @jeamland,   not @Benno  :)

lib/api/1.1/southbound/profiles.js
```diff
@@ -52,17 +52,13 @@ function profilesRouterFactory (
...
-                        });
-                    }
+                    return profileApiService.getProfileFromTaskOrNode(node, 'compute')
+                    .then(function (render) {
+                        presenter(req, res).renderProfile(render.profile, render.options, 200, render.context);
+                    });
                 });
             } else {
-                presenter(req, res).renderProfile('redirect.ipxe');
+                presenter(req, res).renderProfile('redirect.ipxe', { isLookUpNotChecked: true });
```
yyscamper: A little suggestion, how about renaming the variable to `ignoreLookup` or `ignoreLookupFailure`?

anhou: sounds more straightforward, I'll update

spec/lib/api/2.0/profiles-spec.js
anhou: updated. it seems the lines are mismatch with original comments position. the logics is in setLookup, and the unittests cases are there, anyway update 'macs' and 'ips' in unittests to avoid confusion

lib/services/profiles-api-service.js
anhou: updated

spec/lib/services/profiles-api-service-spec.js
```diff
@@ -84,6 +84,20 @@ describe("Http.Services.Api.Profiles", function () {
             }
         };
 
+        var profileReq = {
+            query: {
+                'ips': ['ip1', 'ip2'],
```
yyscamper: Suggest to use valid IP/Mac even in unit-test, we my add some validation logic in future to screen out invalid IP/Mac address.

spec/lib/services/profiles-api-service-spec.js
```diff
@@ -138,6 +152,24 @@ describe("Http.Services.Api.Profiles", function () {
             });
         });
 
+        it("setLookup should set lookup when macs and ips exists in query", function() {
+            this.sandbox.stub(lookupService, 'setIpAddress').resolves();
+
+            return profileApiService.setLookup(profileReq)
+            .then(function(result) {
+                expect(lookupService.setIpAddress).to.be.calledTwice;
```
yyscamper: It's not enough to only checked the callCount, suggest adding some assertion logic for callArgs, for example:
```
expect(lookupService.setIpAddress.firstCall).to.be.calledWith('ip1', 'mac1');
expect(lookupService.setIpAddress.secondCall).to.be.calledWith('ip2', 'mac2');
```

lib/services/profiles-api-service.js
```diff
@@ -98,6 +98,17 @@ function profileApiServiceFactory(
...
+
+        if (query.macs && query.ips) {
```
yyscamper: Need to combine the logic to handle the query pair `mac/ip` and `macs/ips` or we can completely remove the support for one of the query pair.

Recently I am studying the previous API proxy design, the previous code seems have some bug, so I'm acceptable to refactor this after we fully understand or fix the API proxy design.

yyscamper:
add some minor suggestion, but PR is :+1:

anhou:
test this please

heckj:
@anhou seems latest packages are having some trouble with discovery since this went in. @ytjohn was hitting an issue earlier today:

console output from node
```
PXE 1.0.0+ (15a8) -- Open Source Network Boot Firmware -- http://ipxe.org
Features: DNS HTTP TFTP VLAN EFI Menu
MonoRail iPXE
net0: 00:8c:fa:ee:b7:d2 using NII on NII-PCI02:00.0 (open)
  [Link:down, TX:0 TXE:0 RX:0 RXE:0]
  [Link status: Unknown (http://ipxe.org/1a086194)]
Configuring (net0 00:8c:fa:ee:b7:d2)... ok
net0: 10.101.201.4/255.255.255.224 gw 10.101.201.1
Next server: 10.100.16.6
Filename: http://10.100.16.6:9080/api/current/profiles
http://10.100.16.6:9080/api/current/profiles... ok
profiles : 2801 bytes [script]

MonoRail Boilerplate iPXE...
net0: 10.101.201.4/255.255.255.224 gw 10.101.201.1
MonoRail Boilerplate iPXE Completed.

RackHD: NIC0 MAC: 00:8c:fa:ee:b7:d2
RackHD: NIC0 IP: 10.101.201.4
RackHD: Chainloading next profile
http://10.100.16.6:9080/api/current/profiles?macs=00%3A8c%3Afa%3Aee%3Ab7%3Ad2&ips=10.101.201.4.... No such file or directory (http://ipxe.org/2d0c618e)
Failed to download profile, retrying in 3 seconds
RackHD: Chainloading next profile
```

And the resulting logs from `on-http`:

```
==> /var/log/rackhd/on-http.log <==
[debug] [2016-08-18T17:54:04.663Z] [on-http] [Http.Server] [9dd3e0] http: GET 404 1492.834 - 7defe9d1-d203-442b-b642-8fd18c819e43 - /api/current/profiles?macs=00%3A8c%3Afa%3Aee%3Ab7%3Ad2&ips=10.101.201.4
 -> /lib/services/http-service.js:336
ipAddress: 10.101.201.4
```

Looks like the new pattern is reporting a 404 and not chain-loading properly, which is breaking the discovery process from booting up the discovery microkernel.

ytjohn:
This is probably a non issue. Some of the graph definitions were missing and had to be recreated by starting on-taskgraph.   I will check again on 20160817 versions.

anhou:
@heckj @ytjohn I think this PR changes wouldn't bring in this issue . I guess it failed at https://github.com/RackHD/on-http/blob/master/lib/services/profiles-api-service.js#L313 due to graph problems, maybe just like what you've found. If it's an issue after check, let me know or report a bug to me :)

PR #414, opened 2016-08-09
Toggle error view stack trace based on minLogLevel

dalebremner:
@RackHD/corecommitters @keedya @BillyAbildgaard @srinia6

### DIFF
--- a/lib/fittings/error_handler.js
+++ b/lib/fittings/error_handler.js
@@ -6,6 +6,8 @@ var http = require('http');
 module.exports = function create() {
     var injector = require('../../index.js').injector;
     var swaggerService = injector.get('Http.Services.Swagger');
+    var conf = injector.get('Services.Configuration');
+    var constants = injector.get('Constants');
     return function error_handler(context, next) {    // jshint ignore:line
         if (!util.isError(context.error)) { return next(); }
 
@@ -23,9 +25,15 @@ module.exports = function create() {
                 res.locals.errorStatus = err.status;
             }
 
-            var stackDump = err.stack.replace(/\n/g, '');
             message = message.replace(/\n/g, '');
-            res.body = {message: message, status: res.locals.errorStatus, stack: stackDump};
+            res.body = {message: message, status: res.locals.errorStatus};
+            var minLogLevel = conf.get('minLogLevel');
+            if (minLogLevel === constants.Logging.Levels.debug) {
+                if (err.stack !== undefined) {
+                    var stackDump = err.stack.split(/\n/);
+                    res.body.stack = stackDump;
+                }
+            }
             swaggerService.renderer(context.request, context.response, 'error.2.0.json', next);
 
         } else {
--- a/data/views/error.2.0.json
+++ b/data/views/error.2.0.json
@@ -1,5 +1,12 @@
 {
     "message": "<%=message%>",
-    "status": "<%=status%>",
-    "stack": "<%=stack%>"
+    "status": "<%=status%>"
+    <% if (hasOwnProperty("stack")) { %>
+    ,"stack": [
+    <% stack.forEach(function ( item, i, arr){ %>
+    "<%=item %>"
+    <%= ( arr.length > 0 && i < arr.length-1 ) ? ',': '' %>
+    <% }); %>
+    ]
+    <% } %>
 }


heckj:
nice add @dalebremner, I like the pattern.

Might be worth a mention explicitly in the docs as well - maybe a little tweak/additional letting folks know about this upcoming side effect at https://github.com/RackHD/docs/blame/master/docs/rackhd/configuration.rst#L167

heckj:
👍

dalebremner:
will do @heckj

lib/fittings/error_handler.js
```diff
@@ -6,6 +6,8 @@ var http = require('http');
 module.exports = function create() {
     var injector = require('../../index.js').injector;
     var swaggerService = injector.get('Http.Services.Swagger');
+    var nconf = injector.get('nconf');
```
benbp: You should the `Services.Configuration` dependency here instead of `nconf` directly.

lib/fittings/error_handler.js
```diff
@@ -23,9 +25,15 @@ module.exports = function create() {
...
             }
 
-            var stackDump = err.stack.replace(/\n/g, '');
             message = message.replace(/\n/g, '');
-            res.body = {message: message, status: res.locals.errorStatus, stack: stackDump};
+            res.body = {message: message, status: res.locals.errorStatus};
+            var minLogLevel = nconf.get('minLogLevel');
+            if (minLogLevel === constants.Logging.Levels.debug) {
+                if (err.stack !== undefined) {
+                    var stackDump = err.stack.replace(/\n/g, '');
```
benbp: Won't this just make it harder to reconstruct the stack trace in a readable way?

yyscamper: Second on this.

Below is an example of error output:
```
{
    "message": "Could not find node with identifier &#34;57a7fe2dd84b11337a3293a6&#34;",
    "stack": "NotFoundError: Could not find node with identifier &#34;57a7fe2dd84b11337a3293a6&#34;    at /home/yuanf/src/on-core/lib/common/model.js:292:27    at bound (domain.js:287:14)    at runBound (domain.js:300:12)    at tryCatcher (/home/yuanf/src/on-core/node_modules/waterline/node_modules/bluebird/js/main/util.js:26:23)    at Promise._settlePromiseFromHandler (/home/yuanf/src/on-core/node_modules/waterline/node_modules/bluebird/js/main/promise.js:503:31)    at Promise._settlePromiseAt (/home/yuanf/src/on-core/node_modules/waterline/node_modules/bluebird/js/main/promise.js:577:18)    at Async._drainQueue (/home/yuanf/src/on-core/node_modules/waterline/node_modules/bluebird/js/main/async.js:128:12)    at Async._drainQueues (/home/yuanf/src/on-core/node_modules/waterline/node_modules/bluebird/js/main/async.js:133:10)    at Immediate.Async.drainQueues [as _onImmediate] (/home/yuanf/src/on-core/node_modules/waterline/node_modules/bluebird/js/main/async.js:15:14)    at processImmediate [as _immediateCallback] (timers.js:383:17)",
    "status": "404"
}
```

I don't think the stack is the information that should export to user even in debug level if we don't convert it to be more user-friendly.  Also, we'd better to decode the message and stack string to remove  those characters like "&#34;"

anhou: I also think 'stack' trace should not export to API result. either from security or user-friendly perspective.

lib/fittings/error_handler.js
dalebremner: Thanks @benbp ! all set...

keedya:
:+1:

lib/fittings/error_handler.js
dalebremner: @benbp the view fails if there are newlines present...

brianparry: @dalebremner In that case, could we change the template to expect `stack` to be an array of strings and use a forEach to render them all on separate lines?

dalebremner: Thanks for the great idea @brianparry ! Made the change...

brianparry:
:+1:

benbp:
👍 though I agree with @anhou and @yyscamper that we probably don't need/want to return stack traces through the API at all, but rather if debugging you can look them up in logs instead.

That said, on a separate note, maybe it would be a good idea for us to start keeping track of unique request IDs and returning those, so if someone gets a 500 error, we have a unique ID to search the logs with to find the exact message that corresponds to the problematic request? @RackHD/corecommitters @heckj

PR #415, opened 2016-08-10
[redfish] Resolve chassis related ODR

zyoung51:
ODR-795: no blade nodes are contained in the chassis list
- Add blade nodes to chassis list
- Add Contains and ContainedBy member lists

ODR-796: duplicate entries for each enclosure using nodeid and
    chassis serialno
- Remove listing by serial number

ODR-793: ChassisType should reflect Enclosure, Rackmount, or
    Blade as opposed to 'Other'

### DIFF
--- a/lib/api/redfish-1.0/chassis.js
+++ b/lib/api/redfish-1.0/chassis.js
@@ -10,33 +10,42 @@ var Promise = injector.get('Promise'); // jshint ignore:line
 var Errors = injector.get('Errors');
 var taskProtocol = injector.get('Protocol.Task');
 var controller = injector.get('Http.Services.Swagger').controller;
+var env = injector.get('Services.Environment');
+var nodeApi = injector.get('Http.Services.Api.Nodes');
+
+/*
+ * This mapping is used by the Redfish Chassis logic to determine what 
+ * sensors should be displayed in the chassis object representing the
+ * enclosure vs. what sensors should be displayed in the chassis object
+ * representing the blade element.  Entities that are specific to the
+ * blade are filtered out of the enclosure by default.
+ *
+ * This value can be overridden on a sku basis via the redfish.sensorFilter
+ * value.
+ */
+var EntityIdFilterDefaults = [
+    '3', // processor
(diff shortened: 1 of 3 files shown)


lib/api/redfish-1.0/chassis.js
```diff
@@ -10,49 +10,36 @@ var Promise = injector.get('Promise'); // jshint ignore:line
 var Errors = injector.get('Errors');
 var taskProtocol = injector.get('Protocol.Task');
 var controller = injector.get('Http.Services.Swagger').controller;
+var env = injector.get('Services.Environment');
+var nodeApi = injector.get('Http.Services.Api.Nodes');
+
+var EntityIdFilterDefaults = [ '3', '7', '8', '66' ];
```
yyscamper: What's the meaning of these numbers? Please add some comment here.

lib/api/redfish-1.0/chassis.js
```diff
@@ -10,49 +10,36 @@ var Promise = injector.get('Promise'); // jshint ignore:line
...
-            }).then(function(node) {
-                if(!node) {
-                    throw new Errors.NotFoundError('Cound not find node with id ' + identifier);
-                }
+            return nodeApi.getNodeById(identifier)
+            .then(function(node) {
                 return _.filter(node.relations, function(relation) {
                     return relation.relationType === 'encloses';
                 }).map(function(relation) {
                     return relation.targets[0];
```
yyscamper: An enclosure may contains multiple nodes, so relation.targets is an array, and each target is on equal footing and can be in any random order. So what't the reason here only pick the first target?

zyoung51: This is returning a reference to any compute node in the enclosure.  Each target is on equal footing, so the first is picked.

yyscamper:
+1

amymullins:
test this please

brianparry:
test this please

amymullins:
test this please

amymullins:
test this please

PR #424, opened 2016-08-16
Added schema to users route

dalebremner:
@RackHD/corecommitters @srinia6 @BillyAbildgaard @keedya

### DIFF
--- a/static/schemas/2.0/user.2.0.json
+++ b/static/schemas/2.0/user.2.0.json
@@ -1,0 +1,32 @@
+{
+    "title": "User.2.0",
+    "definitions": {
+        "PartialUser": {
+            "description": "A RackHD local user",
+            "type": "object",
+            "properties": {
+                "username": {
+                    "description": "The username of the user",
+                    "type": "string"
+                },
+                "password": {
+                    "description": "The password of the user",
+                    "type": "string",
+                    "minLength": 6,
+                    "maxLength": 50
+                },
+                "role": {
+                    "description": "The role of the user",
+                    "type": "string"
+                }
+            },
+            "additionalProperties": false
+        },
+        "User": {
+            "allOf": [
+                { "$ref": "#/definitions/PartialUser" },
+                { "required": [ "username", "password" ] }
+            ]
+        }
+    }
+}
--- a/static/monorail-2.0.yaml
+++ b/static/monorail-2.0.yaml
@@ -3163,6 +3163,7 @@ paths:
     post:
       x-privileges: [ 'Write' ]
       x-authentication-type: [ 'jwt', 'anonymous' ]
+      x-swagger-schema: user.2.0.json#/definitions/User
       operationId: addUser
       summary: |
         Post a new user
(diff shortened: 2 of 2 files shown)


static/schemas/2.0/user.2.0.json
```diff
@@ -0,0 +1,30 @@
...
+            "properties": {
+                "username": {
+                    "description": "The username of the user",
+                    "type": "string"
+                },
+                "password": {
+                    "description": "The password of the user",
+                    "type": "string"
+                },
+                "role": {
```
yyscamper: The `role` is to specify the read/write right? If so, change the role to `enum` will be better.

static/schemas/2.0/user.2.0.json
```diff
@@ -0,0 +1,30 @@
...
+            "description": "A RackHD local user",
+            "type": "object",
+            "properties": {
+                "username": {
+                    "description": "The username of the user",
+                    "type": "string"
+                },
+                "password": {
+                    "description": "The password of the user",
+                    "type": "string"
```
yyscamper: Recommend to add more conditions for `username` and `password`, these conditions will be played as validation code and keep the code & data clean.  Such as use `minLength` to specify the minimum characters for that filed, and use `patter` to screen some invalid characters.

static/schemas/2.0/user.2.0.json
```diff
@@ -0,0 +1,30 @@
...
+                    "description": "The role of the user",
+                    "type": "string"
+                }
+            },
+            "additionalProperties": false
+        },
+        "User": {
+            "allOf": [
+                { "$ref": "#/definitions/PartialUser" },
+                { "required": [ "username", "password" ] }
```
yyscamper: Not a big problem here, your code still works well here, but I suggest moving the `required` to within `PartialUser`, as they are strongly coupled with the PartialUser properties.

dalebremner: @yyscamper PartialUser is used for PATCH, so the user must be able to specify a variable number of properties using PartialUser.

static/schemas/2.0/user.2.0.json
dalebremner: @yyscamper good idea. Will add the enum...

static/schemas/2.0/user.2.0.json
dalebremner: @yyscamper agreed. I'll add more validation

static/schemas/2.0/user.2.0.json
yyscamper: You are right, I didn't realized the PATH

static/schemas/2.0/user.2.0.json
brianparry: @dalebremner @yyscamper @keedya Long term plan is to allow roles to be added, removed or modified, so the set of enumerated values could change at run-time.  "Read", "Write", etc are privileges not roles and it makes sense to enumerate them in the schema for the roles route when it is available.

yyscamper: If `role` can be updated or removed then I agree it's not suitable to make it as enum in schema.

yyscamper:
:+1:

brianparry:
Looks good :+1:

PR #427, opened 2016-08-18
Add unit-testing for template files

yyscamper:
As I found we put a lot logic into the template file, but it lacks the unit-testing to ensure these logic works as expected. So this PR is to add unit-testing for template files.

I specially created a helper class in [spec/template-ut-helper.js](https://github.com/yyscamper/on-http/blob/e1a5281ab24ddfc67b3ca2e621ada990171fddad/spec/template-ut-helper.js) (and corresponding unit-testing [spec/template-ut-helper-spec.js](https://github.com/yyscamper/on-http/blob/e1a5281ab24ddfc67b3ca2e621ada990171fddad/spec/template-ut-helper-spec.js)) to assist template unit-testing, this class defines a lot chainable methods to obtain the similar syntax just like the style we do unit-testing for API using supertest. 

I also choose some existing template files and write the test cases for them, these test cases can be considered as example to guide the design for other template files. 

Meanwhile, during the design, I also found some template error, so I fix it take this opportunity.

**Just like previous PRs for task-schema, though this PR includes a lot change, but most of  them are similliar**

@RackHD/corecommitters @iceiilin @WangWinson @pengz1

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "istanbul": "^0.3.5",
     "jsdoc": "^3.3.0-alpha13",
     "jshint": "^2.5.11",
+    "jsonlint": "^1.6.2",
     "mocha": "^2.1.0",
     "nock": "~3.5.0",
     "rewire": "^2.5.1",
--- a/spec/data/profiles/install-centos-ipxe-spec.js
+++ b/spec/data/profiles/install-centos-ipxe-spec.js
@@ -1,0 +1,20 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var jsonlint = require('jsonlint');
+var fs = require('fs');
+var path = require('path');
+
+describe('install-centos.ipxe', function() {
+    var subject = 'data/profiles/install-centos.ipxe';
+    var context = jsonlint.parse(
+        fs.readFileSync(path.resolve(__dirname, '../samples/os-common-options.json')).toString()
+    );
+
+    it('should succeed to render', function() {
+        helper.startTemplateTest(subject)
+            .render(context)
+            .expectSuccess();
+    });
+});
--- a/spec/data/profiles/install-esx-ipxe-spec.js
+++ b/spec/data/profiles/install-esx-ipxe-spec.js
@@ -1,0 +1,24 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var jsonlint = require('jsonlint');
+var fs = require('fs');
+var path = require('path');
+
+describe('install-esx.ipxe', function() {
+    var subject = 'data/profiles/install-esx.ipxe';
+    var context = jsonlint.parse(
+        fs.readFileSync(path.resolve(__dirname, '../samples/os-common-options.json')).toString()
+    );
+    context = _.defaults({
+        mbootFile: 'http://172.31.128.1:9080/esxi/6.0/mboot.c32',
+        esxBootConfigTemplateUri: 'http://172.31.128.1:9080/api/current/templates/esx-boot-cfg'
+    }, context);
+
+    it('should succeed to render', function() {
+        helper.startTemplateTest(subject)
+            .render(context)
+            .expectSuccess();
+    });
+});
--- a/spec/data/samples/os-common-options.json
+++ b/spec/data/samples/os-common-options.json
@@ -1,0 +1,113 @@
+{
+    "version": "1.0",
+    "repo": "http://172.31.128.1:9080/os/repo",
+    "rootPassword": "MyPassword",
+    "rootPlainPassword": "MyPassword",
+    "rootEncryptedPassword": "****Pass****",
+    "hostname": "rackhd-node",
+    "domain": "example.com",
+    "ntpServers": ["ntp0.com", "ntp1.com"],
+    "comport": "com1",
+    "users": [
+        {
+            "name": "rackhd1",
+            "password": "123456",
+            "encryptedPassword": "***1***",
+            "uid": 1010,
+            "sshKey": "mysshkey1"
+        },
+        {
+            "name": "rackhd2",
+            "password": "123456",
+            "encryptedPassword": "***2***"
+        }
+    ],
+    "rootSshKey": "mysshkey2",
+    "dnsServers": ["172.12.88.91", "192.168.20.77"],
+    "networkDevices": [
+        {
+            "device": "eth0",
+            "ipv4": {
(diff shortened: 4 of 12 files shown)


changev:
Jenkins: test this please

srinia6:
@yyscamper The views are currently tested as part of the API calls via unit-tests for API. They would fail if the views fail, right? Adding unit-test for views seems redundant here. ( as they just displays content differently)

brianparry:
@yyscamper @srinia6 Yeah - I think this needs some discussion.  While I think unit testing of scripts under the /templates directory may have value in some cases.  I think this change may make changing things like view templates too cumbersome.

yyscamper:
@srinia6 @brianparry : 

In general, these is sure some duplication, but isn't much and still is vauleable.

The motivation for adding template unit-testing is:
- Too much logic in template but lack unit-testing to ensure its quality, for example the kickstart for nearly all OS installation, Actually, there were some bugs that previously reported were just due to template syntax error.
- Template file testing sometimes very cumbersome . Even for a simple fix for kickstart file, we need to trigger OS installation; And for a API view change, we need to restart on-http. Users are reluctant to test while facing the test difficulty. So if the template unit-testing is ready, template error can be detected earlier and in an easier way.

As for the duplication testing about API view, surely there is some duplication, but isn't much. They have different focus: Template unit-testing focuses on whether the template syntax is OK and whether generate correct output; However API unit-tesing foucses on whether the API router works well and reports correct status & repsonse. If we combine all template unit-testing into API testing, the API testing will become very complex, and sometimes (if strictly speaking) it beyond the meaning of `unit`.

Meanwhile, current API unit-tesing doesn't have a good coverage for all template cases, for example, As I commented in this PR (https://github.com/RackHD/on-http/pull/428/files#r75672124), I don't find test cases to test inline task definition, I can also list other cases that don't take every case into consideration. Actually I found some and fixed them in my PR, see below picture. 

image

As for whether the view template will become too cumbersome in future, I don't think so. If you look at the unit-testing implementation for all views template, that's pretty simple and straightforward, and I think that's will be a burden for developer.

For further API design, If unit-testing for view template passes, I think we have gain the confidence that at least view template is OK. If these is further bug in API testing, we can switch the focus to other areas and save us some time.

WangWinson:
:+1: 
Although, view template already has some test. I think Felix's test has a different focus , it's simply verify that the template script in syntax level.  If there is concern on apply this to view template, we may not set it as a force requirement ?

srinia6:
@yyscamper and @brianparry I am not convinced yet to have on unit-test for views. You have justified for templates but not for views[unless you are interpreting both the same way ]. Views are purely for displaying, and you will not see any display if the API's didn't finish successfully, so that makes me wonder if we should have a unit-test for views to begin with? Also, views are purely for displaying the API route contents only.

keedya:
Hi @yyscamper  some of our templates are shell scripts and others are python and java scripts. Form my understanding, adding unit tests to templates should be me more focused on testing functions in these scripts rather then testing the rendering engine. 
I think this is a great effort to evaluate templates rendering but I won't consider it unit tests for templates.
It is a :-1:  for me

srinia6:
I agree with @keedya

yyscamper:
@keedya, @srinia6 : This is not to test the rendering engine, this is to test our template file has correct template syntax. For example, a  template value is enclosed by `<% %>`, but sometimes we may miss the latter `%>`, this is an syntax error. Currently we can only catch such error only if that template is used by some active task, but why not catch such simple error in very early stage? But that's just the benefit of my PR.

Some bugs (some of customer) I had solved is just due to the template syntax, if we write some simple unit-test code for these template files, these bugs can be prevented in early time, why not do this?

yyscamper:
@srinia6 @keedya : Actually, If we have method to validate the template script function, I will do this (For example, We have unit-test for get_driveId.js right now) . But these bash-scripts or python-script can only be run in microkernel, I don't have good way to handle it. So As a compromise, I can at least verify whether the script syntax is OK or not, including template rendering syntax, and even the python/bash syntax.

brianparry:
@yyscamper:  One significant problem with this approach is that DB renderable objects (templates, profiles, and views) are designed with flexibility in mind.  A user can customize templates using the API at run-time.  These tests assume that templates are entirely static and will break if the API is used to modify them.  

Furthermore, unit tests should only need to change when requirements change and should not depend on the implementation of the unit under test.  It looks like these tests will require modification any time a view, profile or template is modified.  I think there is some value in testing templates that render to Python or Javascript, but I don't think the rendering itself should be the focus.  Instead we should be testing the function of these scripts as we would any other piece of code.  Things like kickstart files should be covered by functional tests that perform OS installs.  I think we need to discuss with the @RackHD/corecommitters team, but for now this gets a :-1:  from me.

anhou:
@yyscamper @brianparry @keedya @srinia6 @WangWinson  I think we need to discuss this in the coming CC meeting. @RackHD/corecommitters

WangWinson:
We have very sufficient reason to enable this. But my English is poor ... 
@anhou good idea to discuss on meeting

yyscamper:
@brianparry : The template unit-testing only covers those RackHD built-in templates, those user customized ones are out of our scope. And also that's will not be a big concern, as user will not run unit-testing in product environment.

For TDD driven development, the unit-testing should be written before code implementation, I agree usually the unit-test need not change if no requirement change in TDD model. However: 
Firstly, we are not purely TDD driven at present; 
Secondly, the requirement change varies on different point of view. If the view point is RackHD REST API, then those nodes-api-service, workflows-api-service is purely internal code, so we should remove the unit-testing on nodes-api-service and workflow-api-service? But actually at present, we sometimes need to update unit-testing when the service interface function changes. But if we change our view point to every module, then our current back-and-forth unit-testing change will make sense. The same for view template, if we change the view point to what the view templates will output to us, then adding unit-testing for view template will make sense.

As for the flexibility, the template unit-test helper have considered this. I defined two kinds of assertion function: one is Render-Result-Cared (`expect`, `expectFile`, `expectJson`), another is Render-Result-Careless (`expectSuccess`, `expectFailure`, `expectFormat`). If you want flexibility, you can use those Render-Result-Careless assertion functions.

I have sent you mail that list some bugs I collected caused by template error. You will see even only to check the template syntax (Regardless of render result), a lot bugs can be prevented.

yyscamper:
And my thought on testing is: If we can catch the error in unit-testing, then don't move it to functional testing. The earlier we can catch the bug, the better it is.

Issue #429, opened 2016-08-22
RackHD fails to post workflows to ORA with tag in API 2.0


lhw4d4:
Use tag to post workflows in /api/2.0: 
```
curl -X POST -H 'Content-Type: application/json' -d '{"name": "Graph.PowerOff.Node"}'  
http://{ORA IP}:8080/api/2.0/tags/{tagname}/workflows
```
It will fail and return "Cannot POST /api/2.0/tags/firstgroup/workflows".
And if use the following format to post workflow:
```
curl -X POST -H 'Content-Type: application/json' 
http://{ORA IP}:8080/api/2.0/tags/{tagname}/workflows?name=Graph.PowerOff.Node
```
It will fail and return "validation errors"

yyscamper:
I have found the root cause, the call argument order isn't correct: https://github.com/RackHD/on-http/blob/49456164c0b14843b32606e75661b93c96f3ef00/lib/api/2.0/tags.js#L60

I am working on a fix.

PR #430, opened 2016-08-22
Fix issue that fails to POST workflow via tag

yyscamper:
Resolves https://github.com/RackHD/on-http/issues/429

Root cause:
The argument order for setNodeWorkflow is not correct.

@RackHD/corecommitters @lhw4d4 @srinia6

### DIFF
--- a/spec/lib/api/2.0/tags-spec.js
+++ b/spec/lib/api/2.0/tags-spec.js
@@ -223,13 +223,13 @@ describe('Http.Api.Tags', function () {
                     });
             });
 
-            it('should 400 if tag name is not existing', function() {
+            it('should 404 if tag name is not existing', function() {
                 sandbox.stub(nodesApi, 'getNodesByTag').withArgs('tag001')
-                    .rejects(new Error('fail to find tag with name tag001'));
+                    .rejects(new Errors.NotFoundError('fail to find tag with name tag001'));
                 return helper.request()
                     .post('/api/2.0/tags/tag001/nodes/workflows')
                     .send({ name: 'Graph.test', options: { test: true} })
-                    .expect(400);
+                    .expect(404);
             });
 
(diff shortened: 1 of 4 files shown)


yyscamper:
test please

yyscamper:
test this please

benbp:
👍

anhou:
:+1:

spec/lib/api/2.0/tags-spec.js
```diff
@@ -154,5 +157,108 @@ describe('Http.Api.Tags', function () {
...
+                    .send({ name: 'Graph.test', options: { test: true} })
+                    .expect('Content-Type', /^application\/json/)
+                    .expect(202)
+                    .then(function() {
+                        expect(nodesApi.getNodesByTag).to.be.calledWith('tag001');
+                        expect(nodesApi.setNodeWorkflow).to.not.be.called;
+                    });
+            });
+
+            it('should 400 if tag name is not existing', function() {
```
zyoung51: Should this be a 404 and not a 400?

yyscamper: Actually, this is somewhat beyond the scope for the PR.

The response status depends on what's kind of Error  the `nodesApi.getNodesByTag` returns, if it is returns `NotFoundError`, then the error code is 404, otherwise is 400. If you look at the implementation of `nodesApi.getNodesByTag`, actually it never reports `NotFoundError` if tag name is not existing, instead, it returns empty array.

Since a lot other APIs call `nodesApi.getNodesByTag` as well, sometimes expect empty array, sometimes expect `NotFoundError`, so I'd suggest fixing it in another PR.

At least for this PR, I can change to stub the `nodesApi.getNodesByTag` to return `NotFoundError` and expect 404.

PR #431, opened 2016-08-22
Fix swagger yaml schemes based on RackHD config

rolandpoulter:
This fixes an issue with the swagger-ui failing when `https` is enabled for RackHD. The issue is that our swagger yaml files specify both `http` and `https` as schemes. They should reflect how RackHD is configured.

In order to make the swagger yaml schemes reflect what is in the RackHD configuration. The yaml files are preloaded, modified and saved to a temporary location. Then the patched yaml is used to load the swagger server.

@pscharla


changev:
test this please

lib/services/http-service.js
```diff
@@ -145,26 +146,43 @@ function httpServiceFactory(
     HttpService.prototype.createSwagger = function() {
         var app = this.app;
         var endpoint = this.endpoint;
+
+        function loadYamlConfig(yamlPath) {
+            var config = yaml.safeLoad(fs.readFileSync(yamlPath, 'utf8'));
```
yyscamper: Change to use promise version of readFile, as `createSwagger` is using promise, and so that the read file error can be passed out of `createSwagger` via promise rejection.

rolandpoulter: Don't have to change to the asynchronous version could add a try catch that returns a rejected promise.

```
  try {
    // ...
  } catch (err) {
    return Promise.reject(err)
  }
```

pscharla:
@RackHD/corecommitters This is blocking for OnRack team right now. Looks like we need one more thumbs up before getting this PR merged.

pscharla:
test this please

pscharla:
test this please

yyscamper:
:+1:

yyscamper:
jenkins : test this please

lib/services/http-service.js
yyscamper: Agree.

pscharla:
@RackHD/corecommitters This one is ready for merge

PR #432, opened 2016-08-23
added rack node type and node relation management

VulpesArtificem:
adds a rack node type and capabilities to add and remove 'contains'/'containedBy' relationships as per
https://github.com/RackHD/RackHD/wiki/proposal-passively-add-remove-a-rack-node
and
https://github.com/RackHD/RackHD/wiki/proposal-neighborhood-physical-topology
depends on https://github.com/RackHD/on-core/pull/196
@RackHD/corecommitters @pscharla

### DIFF
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -280,6 +280,11 @@ function nodeApiServiceFactory(
                     }
                 };
                 return workflowApiService.createAndRunGraph(configuration);
+            } else if (node.type === Constants.NodeTypes.Rack) {
+                return eventsProtocol.publishNodeEvent(node, 'added')
+                .then(function() {
+                    return node;
+                });
             }
             return node;
         });
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -47,23 +47,27 @@ function nodeApiServiceFactory(
      * @param  {String}     type
      * @return {Promise}    array of target nodes
      */
-    NodeApiService.prototype._findTargetNodes = function(node, type) {
-        if (!_.has(node, 'relations')) {
+    NodeApiService.prototype._findTargetNodes = function(relations, type) {
+        return this._needTargetNodes(relations, type)
+        .catch(function (err) {
+            logger.warning("Error getting target node with type " + type,
+                { error: err });
+            return [];
+        });
+    };
+
+    NodeApiService.prototype._needTargetNodes = function(relations, type) {
+        if (!relations) {
             return Promise.resolve([]);
         }
 
-        var relation = _.find(node.relations, { relationType: type });
+        var relation = _.find(relations, { relationType: type });
         if (!relation || !_.has(relation, 'targets') ) {
             return Promise.resolve([]);
         }
 
         return Promise.map(relation.targets, function (targetNodeId) {
-            return waterline.nodes.needByIdentifier(targetNodeId)
-            .catch(function (err) {
-                logger.warning("Error getting target node with type " + type,
-                    { error: err });
-                return;
-            });
+            return waterline.nodes.needByIdentifier(targetNodeId);
         });
     };
 
@@ -71,20 +75,15 @@ function nodeApiServiceFactory(
      * Remove the relations to original node in target node. If the node is invalid
      * or doesn't have required relation, this function doesn't need to update the
      * node info and ignore silently with Promise.resolve(). If the relation field
-     * is empty after removing, delete this field.
+     * is empty after removing, delete this field. If the node to be updated is a component
+     * of a node delete it
      * @param  {Object}     node node whose relation needs to be updated
      * @param  {String}     type relation type that needs to be updated
      * @param  {String}     targetId id in relation type that needs to be deleted
      * @return {Promise}    node after removing relation
      */
-    NodeApiService.prototype._removeRelation = function(node, type, targetId) {
-
-        if (!node) {
-            return Promise.resolve();
-        }
-
-        // Remove the relationship between node and downstram node
-        if (!_.has(node, 'relations')) {
+    NodeApiService.prototype._removeRelation = function removeRelation(node, type, targets) {
+        if (!node || !type || !_.has(node, 'relations')) {
             return Promise.resolve();
         }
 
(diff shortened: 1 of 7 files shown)


changev:
test this please

static/monorail-2.0.yaml
```diff
@@ -2321,7 +2321,106 @@ paths:
...
+          in: path
+          description: |
+            Mac addresses and unique aliases to identify the node by, |
+            expect a string or an array of strings.
+          required: true
+          type: string
+      tags: [ "/api/2.0" ]
+      responses:
+        200:
+          description: array of all
```
anhou: array of all xxxxx ?

static/monorail-2.0.yaml
```diff
@@ -2321,7 +2321,106 @@ paths:
...
+      x-authentication-type: [ 'jwt' ]
+      summary: |
+        List of all relations for a node or if there are none an empty object
+      description: |
+        List of all relations for a node or if there are none an empty object
+      parameters:
+        - name: identifier
+          in: path
+          description: |
+            Mac addresses and unique aliases to identify the node by, |
```
anhou: maybe it could be simple and be consistent with other routes like https://github.com/RackHD/on-http/blob/master/static/monorail-2.0.yaml#L2037

DavidjohnBlodgett:
@VulpesArtificem looks like we have some open questions from @anhou

static/monorail-2.0.yaml
VulpesArtificem: Thanks; I cleaned them up a bit accordingly.

pscharla:
@RackHD/corecommitters Please review
@anhou thanks for review. Did you get your concerns addressed?

static/schemas/2.0/relations.2.0.json
```diff
@@ -0,0 +1,23 @@
...
+                "contains": {
+                    "description": "an array of nodes contained by the rack",
+                    "type": "array"
+                },
+                "containedBy": {
+                    "description": "the rack containing the node",
+                    "type": "array"
+                }
+            },
+            "anyOf": [
```
yyscamper: Whether user can input both `containedBy` and `contains`? If not, need change `anyOf` to `oneOf`.

static/schemas/2.0/relations.2.0.json
```diff
@@ -0,0 +1,23 @@
+{
+    "title": "Relations.2.0",
+    "definitions": {
+        "Relation": {
+            "description": "A key value set of relations",
+            "type": "object",
+            "properties": {
+                "contains": {
+                    "description": "an array of nodes contained by the rack",
+                    "type": "array"
```
yyscamper: Please specify the format of array item, for example:
```
"type": "array",
"items": {
    "type": "string",
    "minLength": 1
},
"minItems": 1,
"uniqueItems": true
```

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -501,6 +516,148 @@ describe("Http.Services.Api.Nodes", function () {
...
+            };
+            updateByIdentifier.resolves();
+        });
+
+        it("should do nothing if arguments are missing", function() {
+            var argList = [rackNode, "contains", [computeNode]];
+            _.forEach(argList, function(arg, index) {
+                var argsCopy = [].concat(argList);
+                argsCopy[index] = undefined;
+                nodeApiService._addRelation(argsCopy[0], argsCopy[1], argsCopy[2])
```
yyscamper: This is promise, but it doesn't be returned, so this test case will never fail. You need change the `_.forEach` to `return Promose.each`

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -501,6 +516,148 @@ describe("Http.Services.Api.Nodes", function () {
...
+                rackNode, 'contains', [computeNode.id, computeNode2.id]
+            ).then(function() {
+                expect(updateByIdentifier).to.be.calledWithExactly(rackNode.id, {relations:
+                    [{relationType: 'contains', targets: [computeNode.id, computeNode2.id]}]
+                });
+            }).then(function() {
+                rackNode.relations = [{relationType: 'contains', targets: [computeNode.id]}];
+                return nodeApiService._addRelation(rackNode, 'contains', [computeNode2.id]);
+            }).then(function() {
+                expect(updateByIdentifier).to.be.calledWithExactly(rackNode.id, {relations:
```
yyscamper: updateByIdentifier need be reset before the 2nd `_addRelation`. Otherwise, it only reflects the result of first call of `_addRelation`.

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -501,6 +516,148 @@ describe("Http.Services.Api.Nodes", function () {
...
+    });
+
+
+    describe("getNodeRelations", function() {
+        it("should return the relations field of the requested node", function() {
+            computeNode.relations = [{relationType: "enclosedBy", targets:[enclosureNode]}];
+            waterline.nodes.needByIdentifier.resolves(computeNode);
+            return nodeApiService.getNodeRelations(computeNode.id)
+            .then(function(relations) {
+                expect(relations).to.equal(computeNode.relations);
```
yyscamper: As `relations` is array, so use `deep.equal` rather than `equal`

static/monorail-2.0.yaml
```diff
@@ -2321,7 +2321,115 @@ paths:
...
+        - name: identifier
+          in: path
+          description: |
+              The node identifier
+          required: true
+          type: string
+      tags: [ "/api/2.0" ]
+      responses:
+        200:
+          description: Successfully retrieved the node's reltions
```
yyscamper: typo: reltions -> relations

static/monorail-2.0.yaml
```diff
@@ -2321,7 +2321,115 @@ paths:
...
+            $ref: '#/definitions/Error'
+    delete:
+      x-swagger-schema: relations.2.0.json#/definitions/Relation
+      operationId: nodesDelRelations
+      x-privileges: [ 'Write' ]
+      x-authentication-type: [ 'jwt' ]
+      summary: |
+        Remove relations from nodes
+      description: |
+        Edit the relations fields of a nodes to remove specific relationships
```
yyscamper: typo: `a nodes` -> `a node`

static/monorail-2.0.yaml
```diff
@@ -2321,7 +2321,115 @@ paths:
...
+          description: Successfully removed the relations 
+        404:
+          description: The node with the identifier was not found.
+          schema:
+            $ref: '#/definitions/Error'
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    patch:
```
yyscamper: Since this is to add relations, so use POST instead of PATCH. If doesn't support duplicated relation, then use PUT.

Maybe you think in the view point of node, this is to PATCH node information, but since you have use a new resource name `/nodes/{identifier}/relations`, so the HTTP verb need works on the `relations`.

lib/services/nodes-api-service.js
```diff
@@ -47,44 +47,43 @@ function nodeApiServiceFactory(
      * @param  {String}     type
```
yyscamper: The param in line 46 need be changed.

lib/services/nodes-api-service.js
```diff
@@ -47,44 +47,43 @@ function nodeApiServiceFactory(
...
     /**
      * Remove the relations to original node in target node. If the node is invalid
      * or doesn't have required relation, this function doesn't need to update the
      * node info and ignore silently with Promise.resolve(). If the relation field
-     * is empty after removing, delete this field.
+     * is empty after removing, delete this field. If the node to be updated is a component
+     * of a node delete it
      * @param  {Object}     node node whose relation needs to be updated
      * @param  {String}     type relation type that needs to be updated
      * @param  {String}     targetId id in relation type that needs to be deleted
```
yyscamper: need update according the new API arguments.

lib/services/nodes-api-service.js
```diff
@@ -297,6 +282,68 @@ function nodeApiServiceFactory(
...
+                result.push(relationType);
+            }, [])
+            );
+        })
+        .then(function(targetRelations) {
+            var parentNode = this.parentNode;
+            handler = handler.bind(self);
+            targetRelations = _.chunk(targetRelations, 2); //divide the array into subArray chunks
+                                                           //of [[targetNodes], relationType]
+            return Promise.all(_.transform(targetRelations, function(result, relationSet) {
```
yyscamper: Change `Promise.all` to `Promise.each`? As the relations is mutual and may be complex. The `Promise.all` will run all promises at same time, so there may be some conflict if two promises are handling the two nodes which have mutual-relations. But `Promise.each` will run the promise one by one, so no conflict.

yyscamper:
:+1: after my comments are addressed or clarified.

static/schemas/2.0/relations.2.0.json
```diff
@@ -0,0 +1,23 @@
+{
+    "title": "Relations.2.0",
+    "definitions": {
+        "Relation": {
+            "description": "A key value set of relations",
+            "type": "object",
+            "properties": {
+                "contains": {
```
VulpesArtificem: The scope of this PR is just manual rack management and I don't know if we want to also allow access to the other relation types at this point or if that's in the scope of this PR

benbp:
👍

PR #438, opened 2016-08-24
Pagination support for nodes, workflows, and pollers routes

brianparry:
This PR adds support for pagination using $skip and $top query parameters. 
@RackHD/corecommitters @keedya @geoff-reid @dalebremner @BillyAbildgaard @srinia6 @rolandpoulter

### DIFF
--- a/lib/api/2.0/pollers.js
+++ b/lib/api/2.0/pollers.js
@@ -4,6 +4,7 @@
 
 var injector = require('../../../index').injector;
 var controller = injector.get('Http.Services.Swagger').controller;
+var addLinks = injector.get('Http.Services.Swagger').addLinksHeader;
 var pollers = injector.get('Http.Services.Api.Pollers');
 var Errors = injector.get('Errors');
 
@@ -48,8 +49,16 @@ var pollersLibByIdGet = controller(function(req) {
  * @apiGroup pollers
  * @apiSuccess {json} pollers list of pollers or an empty object if there are none.
  */
-var pollersGet = controller(function(req) {
-    return pollers.getPollers(req.query);
+var pollersGet = controller(function(req, res) {
+    var options = {
+        skip: req.swagger.query.$skip,
+        limit: req.swagger.query.$top
+    };
+
+    return pollers.getPollers(req.query, options)
+    .tap(function(pollers) {
+        return addLinks(req, res, 'workitems', req.query);
+    });
 });
 
 /**
--- a/lib/fittings/error_handler.js
+++ b/lib/fittings/error_handler.js
@@ -32,7 +32,7 @@ module.exports = function create() {
                 uuid: res.locals.uuid
             };
             var minLogLevel = conf.get('minLogLevel');
-            if (minLogLevel === constants.Logging.Levels.debug) {
+            if (/*minLogLevel === constants.Logging.Levels.debug*/true) {
                 if (err.stack !== undefined) {
                     var stackDump = err.stack.split(/\n/);
                     res.body.stack = stackDump;
--- a/lib/fittings/error_handler.js
+++ b/lib/fittings/error_handler.js
@@ -32,7 +32,7 @@ module.exports = function create() {
                 uuid: res.locals.uuid
             };
             var minLogLevel = conf.get('minLogLevel');
-            if (/*minLogLevel === constants.Logging.Levels.debug*/true) {
+            if (minLogLevel === constants.Logging.Levels.debug) {
                 if (err.stack !== undefined) {
                     var stackDump = err.stack.split(/\n/);
                     res.body.stack = stackDump;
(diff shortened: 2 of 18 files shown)


lib/api/2.0/pollers.js
```diff
@@ -48,8 +49,16 @@ var pollersLibByIdGet = controller(function(req) {
...
  */
-var pollersGet = controller(function(req) {
-    return pollers.getPollers(req.query);
+var pollersGet = controller(function(req, res) {
+    var options = {
+        skip: req.swagger.query.$skip,
+        limit: req.swagger.query.$top
+    };
+
+    return pollers.getPollers(req.query, options)
```
anhou: for'req.query' maybe it should be  'var query = req.swagger ? req.swagger.query : req.query'? and extend to all query inputs?

lib/fittings/error_handler.js
```diff
@@ -32,7 +32,7 @@ module.exports = function create() {
                 uuid: res.locals.uuid
             };
             var minLogLevel = conf.get('minLogLevel');
-            if (minLogLevel === constants.Logging.Levels.debug) {
+            if (/*minLogLevel === constants.Logging.Levels.debug*/true) {
```
anhou: debug sentences :)  change back

lib/services/nodes-api-service.js
```diff
@@ -235,8 +235,16 @@ function nodeApiServiceFactory(
      * @returns {Promise}
      */
 
-    NodeApiService.prototype.getAllNodes = function(query) {
-        return waterline.nodes.find(query).populate('obms');
+    NodeApiService.prototype.getAllNodes = function(query, options) {
+        options = options || {};
+        return Promise.try(function() {
+            var query = waterline.nodes.find(query);
```
anhou: 'queryNodes' are more descriptive than 'query'

lib/services/nodes-api-service.js
```diff
@@ -235,8 +235,16 @@ function nodeApiServiceFactory(
...
+
+            if (options.skip) query.skip(options.skip);
+            if (options.limit) query.limit(options.limit);
+
+            return query.populate('obms');
```
anhou: I notice the function https://github.com/RackHD/on-http/blob/master/lib/services/nodes-api-service.js#L238  it could be refined here or there

lib/services/pollers-api-service.js
```diff
@@ -102,8 +102,18 @@ function pollerApiServiceFactory(
...
     };
 
-    PollerApiService.prototype.getPollers = function(query) {
-        return waterline.workitems.find(query);
+    PollerApiService.prototype.getPollers = function(query, options) {
+        options = options || {};
+
+        return Promise.try(function() {
+            console.log('query', query);
+            query = waterline.workitems.find(query);
```
anhou: the same for 'query = ' it could be more descriptive.

lib/services/swagger-api-service.js
```diff
@@ -154,66 +154,87 @@ function swaggerFactory(
...
+            // Set appropriate HTTP status.
+            if (_.isEmpty(res.body) && req.swagger.options.send204OnEmpty) {
+                res.status(204);
+            } else {
+                res.status(res.locals.errorStatus || req.swagger.options.success || 200);
+            }
+        })
+        .then(res.send.bind(res))
+        .catch(function(err) {
+            console.log(err.stack);
```
anhou: using logger?

lib/api/2.0/pollers.js
brianparry: @anhou req.swagger.query contains all "in query" swagger parameters that are explicitly defined in the swagger parameters object.  req.query contains query parameters that are not explicitly listed.  The `var query = req.swagger ? req.swagger.query : req.query` convention is generally only used in services that need to support both 1.1 and 2.0 APIs.  In this case, I want to extract $skip and $top from the swagger query because they are explicitly defined parameters and I want to pass req.query to `getPollers` to allow filtering of pollers using a user-defined query.

lib/fittings/error_handler.js
brianparry: Yep :)

lib/services/swagger-api-service.js
brianparry: This is just debug code that I forgot to remove before submitting the PR.  I'll delete it.

brianparry:
test this please

brianparry:
test this please

keedya:
:+1:

lib/services/nodes-api-service.js
brianparry: @anhou I'm not sure I understand what your asking.

benbp:
Do waterline find queries always return items in a consistent order? If not, probably good to sort by createdAt or something so that top doesn't return duplicates and skip isn't random. 

:+1: if it does though

brianparry:
Thanks @benbp.  At the moment find queries are sorted in "natural" order which is their order on disk.  The order can change if documents are added, deleted, or modified in a way that requires more space on disk.

PR #439, opened 2016-08-25
add a rack added event when a rack node is posted

VulpesArtificem:
for https://github.com/RackHD/RackHD/wiki/proposal-rack-events-notification

@RackHD/corecommitters @pscharla

### DIFF
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -240,8 +240,11 @@ function nodeApiServiceFactory(
     };
 
     NodeApiService.prototype.postNode = function(body) {
-        return waterline.nodes.create(body)
-        .then(function(node) {
+        return Promise.resolve().then(function() {
+            return waterline.nodes.create(body);
+        }).tap(function(node) {
+            return eventsProtocol.publishNodeEvent(node, 'added');
+        }).then(function(node) {
             if(node.type === Constants.NodeTypes.Switch &&
                node.snmpSettings && node.autoDiscover) {
                 return workflowApiService.createAndRunGraph(
@@ -280,11 +283,6 @@ function nodeApiServiceFactory(
                     }
                 };
                 return workflowApiService.createAndRunGraph(configuration);
-            } else if (node.type === Constants.NodeTypes.Rack) {
-                return eventsProtocol.publishNodeEvent(node, 'added')
-                .then(function() {
-                    return node;
-                });
             }
             return node;
         });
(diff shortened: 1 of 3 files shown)


heckj:
:+1:

lib/services/nodes-api-service.js
```diff
@@ -280,6 +280,11 @@ function nodeApiServiceFactory(
                     }
                 };
                 return workflowApiService.createAndRunGraph(configuration);
+            } else if (node.type === Constants.NodeTypes.Rack) {
+                return eventsProtocol.publishNodeEvent(node, 'added')
```
anhou: maybe it's worth to extend to other types of nodes, not only for rack nodes.  if you think it's fine,  https://github.com/RackHD/on-http/blob/master/lib/services/nodes-api-service.js#L284  https://github.com/RackHD/on-http/blob/master/lib/services/profiles-api-service.js#L224  http://rackhd.readthedocs.io/en/latest/rackhd/event_notification.html#node-events  need to  be updated for publish events and update docs.   I also find https://github.com/RackHD/on-http/blob/master/lib/services/profiles-api-service.js#L216  createNodeAndRunDiscovery()  maybe should be located at nodes-api-service.js, not profiles-api-service.js. it could be refined. Sorry for mentioning so much, and out of your original purpose.

VulpesArtificem: I don't know if the event should be triggered for other types of nodes since there's already a 'discovered' event which should be triggered for most node types. "rack" nodes don't have a discovery process though hence this "added" event. 

I think that createNodeAndRunDiscovery() is a method of profiles-api-service just for internal use?

anhou: I think it's fine to add 'added' to other types of nodes events. it's a pre-discovered state, otherwise, it will be a little difficult for documents,  or confuse users in docs.
Forget createNodeAndRunDiscovery(), let me refine it when get a chance :）

VulpesArtificem: I can do that, the 'added' event will only trigger for nodes created through posts to the api though.

VulpesArtificem: I can do that, but the "added" event will only represent/trigger for nodes added through posts to the api.

anhou: Thanks @VulpesArtificem

anhou:
:+1:

pscharla:
@RackHD/corecommitters Looks like this one is ready to be merged

benbp:
👍

Issue #444, opened 2016-08-29
Only Southbound API displayed on swagger-ui


wangzhizheng:
When I opened the swagger-ui and found only Southbound API is displayed.
![pasted image at 2016_08_29 03_50 pm](https://cloud.githubusercontent.com/assets/10055093/18046090/2777ad92-6e11-11e6-9ba0-2ed24a6d0b7e.png)

yyscamper:
Discussed with @wangzhizheng, the programmatic system is installed from debian package, the version is:
```
root@RackHD:/var/renasar/on-http# cat commitstring.txt
137bb9d.2016-08-25 15:38:01 -0700.Merge pull request #439 from VulpesArtificem/rack-event
``` 

I guess the swagger-ui installation fails while building debian package.

wangzhizheng:
When I disabled my firewall and restart the on-http service, all api displayed on the swagger-ui.
I think it is caused by my firewall configuration.

brianparry:
@yyscamper @wangzhizheng This issue has been observed a number of different times under different circumstances.  We definitely should investigate it.  I think I'd start with removing the southbound swagger file altogether since it is not used.

BillyAbildgaard:
link to story in the Stryker backlog:

https://www.pivotaltracker.com/story/show/129367841

heckj:
I've seen this in the past too (probably the mention that @brianparry was referencing). Wasn't able to get a consistent reproduction of the issue, and noted that when I restarted the services, this sometimes disappeared. My hypothesis is that swagger-ui doesn't like to represent multiple endpoints in its documentation - only shows one (north or south), in a non-deterministic fashion

PR #445, opened 2016-08-29
Fix issue with Swagger UI showing only southbound routes.

brianparry:
* Chain swagger registration promises to avoid race condition.

Resolves https://github.com/RackHD/on-http/issues/444

### DIFF
--- a/lib/services/http-service.js
+++ b/lib/services/http-service.js
@@ -169,27 +169,32 @@ function httpServiceFactory(
 
         var swaggerCreateAsync = Promise.promisify(swagger.create);
 
-        var apiV2 = swaggerCreateAsync(swaggerConfig).then(function(swaggerExpress) {
-            app.use(require('restify-links')());
-            return swaggerExpress.register(app);
+        return Promise.try(function() {
+            return swaggerCreateAsync(swaggerConfig)
+            .then(function(swaggerExpress) {
+                app.use(require('restify-links')());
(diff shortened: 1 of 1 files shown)


benbp:
Out of curiosity, why does doing them in sequence fix it rather than just settling on resolution of all promises?

brianparry:
@benbp The problem is caused by the fact that we are assigning two swagger files to the same basepath.  Swagger middleware will return a swagger document corresponding to the first config to be registered. If monorail-2.0-sb.yaml is registered first, you'll see only those routes in the UI.  This is really just a workaround until we get to this story: https://www.pivotaltracker.com/n/projects/1496834/stories/129367887

heckj:
👍

BillyAbildgaard:
good job finding this bug, Brian!

+1

yyscamper:
Good finding, :+1:

lib/services/http-service.js
```diff
@@ -169,27 +169,32 @@ function httpServiceFactory(
...
+            .then(function(swaggerExpress) {
+                app.use(require('restify-links')());
+                return swaggerExpress.register(app);
+            });
+        })
+        .then(function() {
+            return swaggerCreateAsync(sbSwaggerConfig)
+            .then(function(swaggerExpress) {
+                return swaggerExpress.register(app);
+            })
```
yyscamper: Cheer, The houndci recoveries ?!

PR #454, opened 2016-09-02
Add roles route to 2.0 apis

keedya:
depends on https://github.com/RackHD/on-core/pull/208

- Update users view and remove privilege property

@brianparry @zyoung51 @amymullins @srinia6 @dalebremner @BillyAbildgaard @geoff-reid

### DIFF
--- a/lib/services/account-api-service.js
+++ b/lib/services/account-api-service.js
@@ -59,17 +59,27 @@ function accountApiServiceFactory(
             return Promise.all(promises);
         }).then(function() {
             // Enforce a subset of hardcoded privileges:
+            var operator = {
+                role: 'Operator',
+                privileges: ['Login', 'ConfigureComponents', 'ConfigureSelf']
+            };
+            var admin = {
+                role: 'Administrator',
+                privileges: ['Read', 'Write', 'Login', 'ConfigureManager',
+                             'ConfigureUsers', 'ConfigureComponents',
+                             'ConfigureSelf']
+            };
+            var readOnly = {
+                role: 'ReadOnly',
+                privileges: ['Read', 'Login', 'ConfigureSelf']
+            };
             return Promise.all([
-                // 2.0 API's fixed privileges
-                acl.addRoleParents('Administrator', ['Read', 'Write']),
-                acl.addRoleParents('ReadOnly', ['Read']),
-
-                // Redfish 1.0 fixed privileges
-                acl.addRoleParents('Administrator', ['Login','ConfigureManager','ConfigureUsers']),
-                acl.addRoleParents('Administrator', ['ConfigureComponents','ConfigureSelf']),
-                acl.addRoleParents('Operator', ['Login','ConfigureComponents','ConfigureSelf']),
-                acl.addRoleParents('ReadOnly', ['Login', 'ConfigureSelf'])
-            ]);
+                self.createRole(readOnly),
+                self.createRole(operator),
+                self.createRole(admin)
+            ]).catch(function () {
+                //mask error here
+            });
         });
     };
 
@@ -186,6 +196,106 @@ function accountApiServiceFactory(
             return {
                 username: name
             };
+        });
+    };
+
+    AccountApiService.prototype.createRole = function (newrole) {
+        var self = this;
+        var acl = this.acl;
+        return Promise.try(function () {
+            assert.string(newrole.role);
+            return waterline.roles.create({
+                role: newrole.role,
+                privileges: newrole.privileges
+            });
+        })
+        .then(function (entry) {
+            return acl.addRoleParents(entry.role, entry.privileges);
+        })
+        .then(function () {
+            return self.getRoleByName(newrole.role);
+        })
+        .catch(function(err) {
+            if (err.message.match(/A record with that `role` already exists/)) {
+                err = new Errors.BaseError('role already exists');
+                err.status = 409;
+            }
+            throw err;
+        });
+    };
+
+    AccountApiService.prototype.getRoleByName = function (name) {
+        var self = this;
+        return Promise.try(function () {
+            assert.string(name);
+            return waterline.roles.findOne({ role: name });
+        })
+        .then(function (roleFound) {
+            if (!roleFound) {
+                throw new Errors.NotFoundError();
+            }
+            return Promise.props({
+                role: roleFound.role,
+                privileges: self.aclMethod('_rolesParents', [roleFound.role])
+            });
+        });
+    };
+
+    AccountApiService.prototype.listRoles = function () {
+        var self = this;
+        return Promise.map(waterline.roles.find({}), function (role) {
+            return Promise.props({
+                role: role.role,
+                privileges: self.aclMethod('_rolesParents', [role.role])
+            });
+        });
+    };
+
+    AccountApiService.prototype.removeRoleByName = function (name) {
+        var acl = this.acl;
+        return Promise.try(function () {
+            assert.string(name);
+        })
+        .then(function () {
+            return waterline.roles.findOne({ role: name });
+        })
+        .then(function (entry) {
+            if (!entry) {
+                throw new Errors.NotFoundError();
+            }
+            return Promise.all([
+                acl.removeRoleParents(entry.role),
+                waterline.roles.destroy({ role: name })
+            ]);
+        })
+        .then(function () {
+            return {
+                role: name
+            };
+        });
+    };
+
+    AccountApiService.prototype.modifyRoleByRoleName = function (name, info) {
+        var self = this;
+        var acl = this.acl;
+        return Promise.try(function () {
+            assert.string(name);
+            return waterline.roles.findOne({ role: name });
+        })
+        .then(function (entry) {
+            if (!entry) {
+                throw new Errors.NotFoundError();
+            }
+            return acl.removeRoleParents(entry.role);
+        })
+        .then(function () {
+            return waterline.roles.update({ role: name }, info.privileges);
+        })
+        .then(function () {
+            return acl.addRoleParents(name, info.privileges);
+        })
+        .then(function () {
+            return self.getRoleByName(name);
         });
     };
 
--- a/spec/lib/api/2.0/roles2.0-spec.js
+++ b/spec/lib/api/2.0/roles2.0-spec.js
@@ -1,0 +1,68 @@
+// Copyright 2016, EMC, Inc.
+/* jshint node:true */
+
+'use strict';
+
+describe('Http.Api.Roles', function () {
+    var roleObj = {
+        role: 'admin-role',
+        privileges: ['write', 'read']
+    };
+
+    var newRole = {
+        role: 'read-role',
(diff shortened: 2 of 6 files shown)


lib/services/account-api-service.js
```diff
@@ -189,5 +197,108 @@ function accountApiServiceFactory(
...
+
+    AccountApiService.prototype.getRoleByName = function (name) {
+        var self = this;
+        return Promise.try(function () {
+            assert.string(name);
+            return waterline.roles.findOne({ role: name });
+        })
+        .then(function (roleFound) {
+            if (!roleFound) {
+                throw new Errors.NotFoundError();
```
yyscamper: Add some message for the error.

lib/services/account-api-service.js
```diff
@@ -59,17 +59,25 @@ function accountApiServiceFactory(
...
-                acl.addRoleParents('Administrator', ['Login','ConfigureManager','ConfigureUsers']),
-                acl.addRoleParents('Administrator', ['ConfigureComponents','ConfigureSelf']),
-                acl.addRoleParents('Operator', ['Login','ConfigureComponents','ConfigureSelf']),
-                acl.addRoleParents('ReadOnly', ['Login', 'ConfigureSelf'])
-            ]);
+                self.createRole(readOnly),
+                self.createRole(operator),
+                self.createRole(admin)
+            ]).catch(function () {
+                //mask error here
```
yyscamper: Please fix this before removing the `WIP` flag.

lib/services/account-api-service.js
```diff
@@ -189,5 +197,108 @@ function accountApiServiceFactory(
...
+        var acl = this.acl;
+        return Promise.try(function () {
+            assert.string(name);
+        })
+        .then(function () {
+            return waterline.roles.findOne({ role: name });
+        })
+        .then(function (entry) {
+            if (!entry) {
+                throw new Errors.NotFoundError();
```
yyscamper: Add some message for t he threw error.

spec/lib/api/2.0/roles2.0-spec.js
```diff
@@ -0,0 +1,70 @@
...
+    var accountService;
+    var waterline;
+
+    before('start HTTP server', function () {
+        var self = this;
+        this.timeout(5000);
+        this.sandbox = sinon.sandbox.create();
+        return helper.startServer([])
+        .then(function () {
+            console.log('stub accountSevice');
```
yyscamper: remove this before removing "WIP" flag

spec/lib/services/account-api-service-spec.js
```diff
@@ -57,98 +70,193 @@ describe("Account API Service", function() {
...
-                expect(user.username).to.be.a('string');
+    describe('Users services', function () {
+
+        it('should list users', function () {
+            waterline.localusers.find.resolves([user1, user2]);
+            return accountService.listUsers()
+            .then(function (users) {
+                expect(waterline.localusers.find).to.have.been.called;
+                expect(users.length).to.equal(2);
+                return Promise.map(users, function (user) {
```
yyscamper: Actually the `users` is already an object, you don't need to use Promise.map here.

spec/lib/services/account-api-service-spec.js
```diff
@@ -57,98 +70,193 @@ describe("Account API Service", function() {
...
+                    expect(user.username).to.be.a('string');
+                    expect(user.password).to.be.an('undefined');
+                    expect(user.role).to.be.a('string');
+                    expect(user.privileges).to.be.instanceof(Array);
```
yyscamper: For those user's property, we'd better check their content rather than only the type.

spec/lib/services/account-api-service-spec.js
```diff
@@ -57,98 +70,193 @@ describe("Account API Service", function() {
...
-            expect(user.privileges).to.be.instanceof(Array);
-            expect(user.privileges[0]).to.equal('ConfigureUsers');
+        it('should recreate a user', function () {
+            waterline.localusers.findOne
+                .onFirstCall().resolves(user1)
+                .onSecondCall().resolves(user1);
+            waterline.localusers.create.resolves(user1);
+            return accountService.createUser(user1)
+            .then(function () {
+                expect(false);
```
yyscamper: This will not work as expected

spec/lib/services/account-api-service-spec.js
```diff
@@ -57,98 +70,193 @@ describe("Account API Service", function() {
...
+        var role1 = {
+            role: 'role1',
+            privileges: ['read']
+        };
+
+        it('should not create a new role if role already exist', function () {
+            waterline.roles.findOne.resolves(role1);
+            return accountService.createRole(role1)
+            .then(function () {
+                return new Error("expected a failure");
```
yyscamper: This will not work as expected.

spec/lib/services/account-api-service-spec.js
```diff
@@ -57,98 +70,193 @@ describe("Account API Service", function() {
...
+                expect(res[0]).to.have.property('role').that.equal(role1.role);
+                expect(res[0]).to.have.property('privileges').that.deep.equal(role1.privileges);
+            });
+        });
+
+        it('should return not found if role does not exist', function () {
+            waterline.roles.findOne.resolves();
+            return accountService.getRoleByName(role1.role)
+            .then(function (res) {
+                return new Error("expected a failure");
```
yyscamper: This will not work as expected.

static/monorail-2.0.yaml
```diff
@@ -3493,6 +3493,172 @@ paths:
...
+            type: object
+        403:
+          description: Forbidden
+          schema:
+            type: object
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    post:
```
yyscamper: Whether the role can be duplicated? If not, change the `POST` to `PUT`

spec/lib/services/account-api-service-spec.js
keedya: @yyscamper this is not part of this PR. it is existing code

spec/lib/services/account-api-service-spec.js
keedya: @yyscamper same here

spec/lib/services/account-api-service-spec.js
keedya: @yyscamper this is not part of my PR

lib/services/account-api-service.js
keedya: @yyscamper this is done on purpose to mask the error. This function needs to run on startup. This part will create the essential roles (Administrator, ReadOnly and operator) if they don't exist

spec/lib/services/account-api-service-spec.js
yyscamper: @keedya : the existing code is not always correct. If we found some minor bug and can fix it with little effort, I strongly recommend to do this. Otherwise, we may forget it, and it's not worth to open a PR even an issue to track this.

spec/lib/services/account-api-service-spec.js
yyscamper: Just feel free to change existing code if it is not reasonable.

lib/services/account-api-service.js
yyscamper: If it is purposely, that's OK.

lib/services/account-api-service.js
```diff
@@ -189,5 +199,107 @@ function accountApiServiceFactory(
...
+                role: newrole.role,
+                privileges: newrole.privileges
+            });
+        })
+        .then(function (entry) {
+            return acl.addRoleParents(entry.role, entry.privileges);
+        })
+        .then(function () {
+            return self.getRoleByName(newrole.role);
+        });
```
brianparry: @keedya We may want to add a `.catch` here so we respond with a 409 in the event of a duplicate key error.  Otherwise, I think we'll return a 500 if two REST clients attempt to add a role concurrently.  In fact, we might not need the initial `findOne` query, we might be able to rely on the error thrown when a Mongo index violation occurs.

Maybe something like:
```
.catch(function(err) {
    if (err.message.match(/E11000 duplicate key error index:/) {
        var err = new Errors.BaseError('role already exists');
        err.status = 409;
        throw err;
    }
    throw err;
})
```

lib/services/account-api-service.js
brianparry: We could use `waterline.roles.needByIdentifier` and remove the `!roleFound` condition altogether.  I'm fine with this either way though.

static/monorail-2.0.yaml
```diff
@@ -3493,6 +3493,172 @@ paths:
...
+            type: object
+        403:
+          description: Forbidden
+          schema:
+            type: object
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    put:
```
brianparry: @keedya Shouldn't this be POST?

static/monorail-2.0.yaml
```diff
@@ -3493,6 +3493,172 @@ paths:
...
+          in: body
+          description: The role information
+          required: true
+          schema:
+            $ref: '#/definitions/role_obj'
+      tags: [ "/api/2.0" ]
+      responses:
+        200:
+          description: |
+            Successfully modified the user
```
brianparry: Minor: user should be role

static/monorail-2.0.yaml
```diff
@@ -3493,6 +3493,172 @@ paths:
...
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    delete:
+      x-privileges: [ 'Write' ]
+      x-authentication-type: [ 'jwt' ]
+      operationId: removeRole
+      summary: |
+        Delete a user
```
brianparry: Minor: user should be role

static/monorail-2.0.yaml
```diff
@@ -3493,6 +3493,172 @@ paths:
...
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    get:
+      x-privileges: [ 'Write' ]
+      x-authentication-type: [ 'jwt' ]
+      operationId: getRole
+      summary: |
+        Get a user
```
brianparry: Minor: user should be role

keedya:
@RackHD/corecommitters @brianparry @dalebremner @yyscamper @srinia6 @BillyAbildgaard @geoff-reid 

Addressed @brianparry  and @yyscamper comments

For the comments outside the scope of this PR, I prefer to address them in a separate PR

brianparry:
Thanks @keedya :+1:

yyscamper:
:+1: With a suggestion to fix the existing code's minor bug.

keedya:
@yyscamper I am planning for some changes in users that will require changes in the unit tests as well.
Therefore the effort will be captured in a different PR.
(Changes includes making username a unique index in mongo)

rackhdmirror:
Build finished.

PR #456, opened 2016-09-08
Increase the buffer size of child process in bootstrap.js

iceiilin:
currently the max buffer size of child process in bootstrap.js is 1000*1024, but the stdout of flashing bios task is 1131830. from the child process document, "largest amount of data (in bytes) allowed on stdout or stderr - if exceeded child process is killed". The original setting prevents the flashing bios from finishing and would result in the successive catalogs task hanging.
Increase the max buffer size to 5MB


yyscamper:
:+1:

heckj:
:+1:

cgx027:
:+1:

PR #459, opened 2016-09-09
Replace completionUri with notification

lanchongyizu:
* replace completionUri with notification in OS kickstart files
* additional works include:
	- esx: move the install-os completion into esx-ks %post, and remove postLookup
	- windows: move the install-os completion into winpe-kickstart.ps1

### DIFF
--- a/data/templates/autounattend_windows_2012_r2_SMC_6028U-TR4+.xml
+++ b/data/templates/autounattend_windows_2012_r2_SMC_6028U-TR4+.xml
@@ -84,7 +84,7 @@
             </AutoLogon>
             <FirstLogonCommands>
 	       <SynchronousCommand wcm:action="add">
-                    <CommandLine>powershell -Command "Invoke-WebRequest -uri http://<%=server%>:<%=port%>/api/current/templates/renasar-ansible.pub"</CommandLine>
+                    <CommandLine>powershell -Command "curl -Method POST -ContentType 'application/json' http://<%=server%>:<%=port%>/api/current/notification?nodeId=<%=nodeId%> -Outfile windows-install.log"</CommandLine>
                     <Description>flag the end of the setup</Description>
                     <Order>1</Order>
                     <RequiresUserInput>true</RequiresUserInput>
--- a/data/templates/autounattend_windows_2012_r2_unknown.xml
+++ b/data/templates/autounattend_windows_2012_r2_unknown.xml
@@ -84,7 +84,7 @@
             </AutoLogon>
             <FirstLogonCommands>
 	       <SynchronousCommand wcm:action="add">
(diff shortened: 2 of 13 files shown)


rackhdmirror:

lanchongyizu:
@RackHD/corecommitters @zyoung51 @cgx027 @uppalk1

data/templates/autounattend_windows_2012_r2_SMC_6028U-TR4+.xml
```diff
@@ -84,7 +84,7 @@
             </AutoLogon>
             <FirstLogonCommands>
 	       <SynchronousCommand wcm:action="add">
-                    <CommandLine>powershell -Command "Invoke-WebRequest -uri http://<%=server%>:<%=port%>/api/current/templates/renasar-ansible.pub"</CommandLine>
+                    <CommandLine>powershell -Command "curl -Method POST -ContentType 'application/json' http://<%=server%>:<%=port%>/api/current/notification?nodeId=<%=nodeId%> -Outfile windows-install.log"</CommandLine>
```
cgx027: may consider replace -Method with a simple -X to keep consistent with other places.

cgx027:
:+1:

uppalk1:
@lanchongyizu Did you consider checking the unattend_server2012.xml and post-windows-install.ps1 and test them with your changes ? I believe Azzam @tannoa2 created these files as part of the Windows OS install workflow.

yyscamper:
:+1:

lanchongyizu:
@uppalk1 I have tested them already.

keedya:
@lanchongyizou is maxpayload for Windows server is been tested with ur changes?
The reason I am  asking is if you set static Ip, post install script will fail and notification api u added will never be called. This is an existing issue that @uppalk1 was working on.

keedya:
@lanchongyizu post install script fails because of lookup up IP to MAC address due to set static ip during the kick start script of Windows installation

data/templates/autounattend_windows_2012_r2_SMC_6028U-TR4+.xml
lanchongyizu: curl in Windows is different from that in Linux, and -X is not supported in Windows.

lanchongyizu:
@keedya The notification API doesn't depend on lookup, so setting static IP isn't an issue now. And I have tested it too.

yyscamper:
@keedya, the intention of notification URL is just want to get rid of the dependency of lookup. @lanchongyizu have tested the windows installation and seems everything works fine until now.

keedya:
@yyscamper post install script depends on lookup and it is the script where u call the notification Api. We tested this code in our lab and I am confident  that static ip will fail to finish the Windows installation.

lanchongyizu:
@keedya This PR doesn't cover static IP now, and we could follow up the problem later. Thanks.

PR #475, opened 2016-09-26
seperate static file of bootstrap and os repos from api.server into file.server

lanchongyizu:
* seperate static file of bootstrap and os repos from api.server into file.server

@RackHD/corecommitters @iceiilin @cgx027 @panpan0000 @benbp 

https://github.com/RackHD/on-tasks/pull/347
https://github.com/RackHD/on-taskgraph/pull/173

### DIFF
--- a/lib/services/common-api-presenter.js
+++ b/lib/services/common-api-presenter.js
@@ -237,6 +237,16 @@ function CommonApiPresenterFactory(
                                                         config.apiServerPort ||
                                                         80;
         var serverUri = 'http://' + apiServerAddress + ':' + apiServerPort;
+        var fileServerUri;
+        if (_.has(config, 'fileServerAddress') &&
+            _.has(config, 'fileServerPort') &&
+            _.has(config, 'fileServerPath')) {
+            fileServerUri = 'http://' + config.fileServerAddress + ':' +
+                            config.fileServerPort + config.fileServerPath;
+            fileServerUri = _.trimRight(fileServerUri, '/');
+        } else {
+            fileServerUri = serverUri;
+        }
         var baseUri = serverUri + '/api/current';
 
         graphContext = graphContext || {};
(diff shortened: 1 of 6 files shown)


rackhdmirror:

lib/services/swagger-api-service.js
```diff
@@ -306,7 +306,17 @@ function swaggerFactory(
             config.get('apiServerAddress'),
             config.get('apiServerPort')
         );
-        var baseUri = util.format('%s/%s', apiServer, req.swagger.operation.api.basePath);
+        var baseUri = util.format('%s%s', apiServer, req.swagger.operation.api.basePath);
```
iceiilin: Is it necessary to remove '/' here?

lanchongyizu: Since `req.swagger.operation.api.basePath` already has '/', `%s/%s` will contain '//'.

lib/services/common-api-presenter.js
```diff
@@ -237,6 +237,16 @@ function CommonApiPresenterFactory(
                                                         config.apiServerPort ||
                                                         80;
         var serverUri = 'http://' + apiServerAddress + ':' + apiServerPort;
+        var fileServerUri;
+        if (_.has(config, 'fileServerAddress') &&
+            _.has(config, 'fileServerPort') &&
+            _.has(config, 'fileServerPath')) {
```
cgx027: I would suggest not making fileServerPath a mandatory field. If user meant to use the root directory of the file server, they would got a chance of not specifying the fileServerPath. 

This is a non blocking anyway.

lib/services/swagger-api-service.js
```diff
@@ -306,7 +306,17 @@ function swaggerFactory(
...
+        var fileServerUri;
+        if (config.get('fileServerAddress') !== undefined &&
+            config.get('fileServerPort') !== undefined &&
+            config.get('fileServerPath') !== undefined) {
```
cgx027: Same as above

iceiilin:
great! :+1:

PR #476, opened 2016-09-27
[redfish] Resolve ODR-938

zyoung51:
Always use the chassis type defined in the sku configuration over the
chassis type defined in the dmi.


tannoa2:
+1

PR #480, opened 2016-09-29
Added admin and readonly to user.roles schema

lacarb:
This is for ODR-833 https://hwjiraprd01.corp.emc.com/browse/ODR-833
This restricts the creation of users to two roles, Administrator and ReadOnly. Previously, anything could be in the roles field. Presently, these are the only two roles acceptable for 2.0.


lacarb:
@benbp @yyscamper @anhou @brianparry

rackhdmirror:

PR #483, opened 2016-10-04
[redfish] Resolve ODR-920 and ODR-927

zyoung51:
- Update Redfish to use new OBM models
- Resolve issues where /Managers was returning 200 with an error
  payload when it should 404


rackhdmirror:
Build finished.

PR #494, opened 2016-10-14
fixing last reboot in ks

BillyAbildgaard:
fixing esxi 5.5 so it passes in jenkins continuous-os-installs
@geoff-reid


rackhdmirror:
Build finished.

PR #496, opened 2016-10-17
Set prefix as repo address to simplify ESXi boot.cfg design

sunnyqianzhang:
To address Young, Zachary's comment in PR (https://github.com/RackHD/on-http/pull/38), we can set the prefix as repo address in esx-boot-cfg file so that we don't need long "https://........" appear everywhere.
@RackHD/corecommitters @iceiilin @pengz1

### DIFF
--- a/data/templates/esx-boot-cfg
+++ b/data/templates/esx-boot-cfg
@@ -1,6 +1,6 @@
 bootstate=0
 title=Loading ESXi installer
-prefix=<%=repo%>/
+prefix=<%=repo%>
 kernel=<%=tbootFile%>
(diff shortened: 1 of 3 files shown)


rackhdmirror:
Build finished.

sunnyqianzhang:
This PR need cowork with the following PR to take effective: https://github.com/RackHD/on-tasks/pull/357

pengz1:
+1

data/templates/esx-boot-cfg
```diff
@@ -1,5 +1,6 @@
 bootstate=0
 title=Loading ESXi installer
+prefix=<%=repo%>/
```
anhou: doesn't need this '/'?

rackhdmirror:
Build finished.

data/templates/esx-boot-cfg
sunnyqianzhang: Double confirmed and it works without this '/', thanks @anhou

PR #497, opened 2016-10-17
Ztip esx kickstart update

nucklehead:
## **I have been Working on a project here at Dell EMC using RackHD and here are few issues I have encountered with the ESX Kickstarter.**

**Here are the changes i am proposing**
- The section performing this action `Search the networkDevices and set the first device (if defined) up.` is not really needed because the installer will by default set the interface where it is pxe booting from with dhcp if the network is not passed. The relevant network configuration happens in %firtboot script.
  Also this assumes the first interface in the list is the interface connect to RackHD which is not necessary true 
- The payload to update the lookups with vmk information only requires mac address and node id. The ip will be updated from the pollers right after I believe. If the vmk is set for dhcp there will be no ip to pass.

```
BODY="{"
   BODY=$BODY"\"macAddress\": \"$mac\","         
-  BODY=$BODY"\"ipAddress\": \"$2\","       
   BODY=$BODY"\"node\": \"<%=nodeId%>\""         
   BODY=$BODY"}"
```

Switch $2 to $mac because in case of dhcp we do not have ip yet

```
/api/current/lookups?q=$2 
/api/current/lookups?q=$mac
```
- The -i was added just to ignore case
  `grep -i <%=n.device%>
  `
- The `rackhdCallbackScript`variable is no longer in context once `Task.Os.Install.ESXi` is marked as completed. So The wget will fail in %firstboot.
  The file is downloaded in the %post script instead to the datastore and used to send first notification. Then moved to /etc/rc.local.d/local.sh in the %firstboot
  `mv /vmfs/volumes/datastore1/<%=rackhdCallbackScript%> /etc/rc.local.d/local.sh
  `
  This is the section to download the callback script in the datastore.
  `wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%> -O /vmfs/volumes/datastore1/<%=rackhdCallbackScript%>`
  We can then use it to send the notification

```
chmod +x /vmfs/volumes/datastore1/<%=rackhdCallbackScript%>
/vmfs/volumes/datastore1/<%=rackhdCallbackScript%>
```


rackhdmirror:

nucklehead:
@mtannous this is the PR for the esx-ks. I will create the one for centos-ks soon.

anhou:
@yyscamper @pengz1  could help review more details on this

lanchongyizu:
:-1:

data/templates/esx-ks
```diff
@@ -236,7 +214,9 @@ cp /var/log/esxi_install.log "/vmfs/volumes/datastore1/firstboot-esxi_install.lo
 #
 # NOTE: this script will execute right away as a result of writing it to local.sh
 # along with executing on every subsequent boot
-wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%> -O /etc/rc.local.d/local.sh
+# rackhdCallbackScript is no longer in context once "Task.Os.Install.ESXi" is marked as completed
+
```
nucklehead: Yes but the wget itself will fail because the esx.rackhdcallback file containsvariables to be rendered as well.

data/templates/esx-ks
```diff
@@ -7,28 +7,6 @@ accepteula
...
-      network --bootproto=static --device=<%=n.device%> <%=ipopts%>
-    <% } else { %>
-      network --bootproto=dhcp --device=<%=n.device%>
-    <% } %>
-    <% return false; %>
-  <% }); %>
-  <% if (need_default) { %>
-    network --bootproto=dhcp --device=vmnic0
-  <% } %>
-<% } %>
```
nucklehead: wouldn't the first boot section take care of that? Even if dhcp is set for installation, once firstboot runs the ip will be set to static as desired.

yyscamper:
jenkins: ok to test

rackhdmirror:
Build finished.

nucklehead:
can you please test this please

tannoa2:
test this please

data/templates/esx-ks
lanchongyizu: But `server` and `port` is in context then, so `wget` wouldn't fail in your comment here.

brianparry:
@nucklehead This PR has been open for a long time.  Are you still looking to have this merged?  If so, could you please rebase with the master branch.

nucklehead:
@brianparry hey sorry. Yeah I ended up losing track on this moving form  ztip to vayager. This is still valid. I will rebase and push later today.

anhou:
some problems for this PR, and it caused regression test failure, I confirmed with @lanchongyizu  about these problems
1. call /notifications URL is essential, cannot be removed.
2. postlookup() is out of date, currently, we don't need this function, don't need to add it back to previous status
3. rackhdCallbackScript will be rendered once esx-ks is downloaded.

@nucklehead  Because this PR will cause OS installation regression test failure, I'll revert this PR, when you have more time, you could address these problems above and submit another PR, If you don't have more time, we could help pick useful things in this PR and submit another PR.

nucklehead:
@anhou . Ok thanks. Sorry for this issue. It was working at the time of the PR. I will check to see if all these changes are still needed with the latest rackhd code and fix whats needed.

anhou:
@nucklehead OK, Thanks!

PR #499, opened 2016-10-18
fix issue#430 about templates api in version 1.1

iceiilin:
fix https://github.com/RackHD/RackHD/issues/430

templates.get() interface requires scope as a parameter. If it is not specified, this interface only returns the templates with "global" scope.

In templates/library API, add "scope" as a parameter when getting templates.
In templates/library/:id, add "?scope=xxx" to get a template for a specific scope, which is align with the behavior in API 2.0. If user doesn't use this query, the default scope is global.

@cgx027 @changev

### DIFF
--- a/lib/api/1.1/northbound/templates.js
+++ b/lib/api/1.1/northbound/templates.js
@@ -4,7 +4,6 @@
 
 var di = require('di');
 var express = require('express');
-var parser = require('body-parser');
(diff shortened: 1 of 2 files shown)


rackhdmirror:
Build finished.

gpaulos:
+1

PR #500, opened 2016-10-18
Remove postLookup() logic, and fix the failure of installing Esxi when setting static IP.

lanchongyizu:
The solution is
- wget callbackScript into directory /tmp in %pre
- then move it to /vmfs/volumes/datastore1 in %post
- then copy the content into /etc/rc.local.d/local.sh

@RackHD/corecommitters @panpan0000 @keedya

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -33,33 +33,6 @@ reboot
 
 %firstboot --interpreter=busybox
 
-postLookup () {
-  echo "Attempting postLookup operation on $1" >> /vmfs/volumes/datastore1/firstboot.log
-  mac=`esxcli --debug --formatter=csv network ip interface list | grep $1 | awk -F, '{print $3}'`
-  if [ "<%=version%>" == "5.5" ]; then
-    mac=`esxcli --debug --formatter=csv network ip interface list | grep $1 | awk -F, '{print $2}'`
-  fi
-  BODY="{"
-  BODY=$BODY"\"macAddress\": \"$mac\","
-  BODY=$BODY"\"ipAddress\": \"$2\","
-  BODY=$BODY"\"node\": \"<%=nodeId%>\""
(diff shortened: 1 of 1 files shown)


rackhdmirror:

keedya:
@lanchongyizu I m not sure ssh validation will pass in this case. I probably missed something here?

yyscamper:
@lanchongyizu : could you please explain why this can solve the static IP issue?

lanchongyizu:
@keedya For now, ssh validation fails for all OSes, so it is not caused by this PR.
@yyscamper Because callbackScript is `wget` in `%pre` where static IP is not set then.

panpan0000:
:+0: :-)  looks good.  just questions:  (1) so in %pre, the "/vmfs/volumes/datastore1" has not been established or is not accessible ?    (2) can you help: why wget works in %pre  but not in %firstboot ?

lanchongyizu:
@panpan0000 
(1) yes, it's not established
(2) `wget` in `%firstboot` will fail since setting static IP makes `lookups` fail  when pulling callbackScript.

keedya:
@lanchongyizu it  is :-1: for me. If lookup table is not updated, ssh validation will fail.

lanchongyizu:
@keedya When the last notification is posted, static IP will be in ARP cache, so it is in lookup table.

yyscamper:
My comment purely depends on the test result. I eagerly to remove the "ugly" code to POST lookup in ESXi kickstart file, if the testing is OK, this is definitely :+1:

lanchongyizu:
Since the issue https://github.com/RackHD/RackHD/issues/467 exists, we can't remove postLookups() currently.

lanchongyizu:
This has been solved by https://github.com/RackHD/on-http/pull/531, so close the PR.

PR #504, opened 2016-10-20
Add alerts when SSDP services are added or removed.

brianparry:
* Add node-cache dependency.
* Handlers for cache delete, set, and expire events
* Set cache entries based on M-SEARCH response.

@RackHD/corecommitters


rackhdmirror:
Build finished.

keedya:
:+1:

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

benbp:
I don't have an issue with adding the cache dependency here, but just an FYI we do already include https://www.npmjs.com/package/lru-cache in our dependencies, albeit it has some slightly different behavior (maxAge vs. ttl, dispose callbacks but no way to trigger callbacks on `set` like node-cache does).

brianparry:
@benbp I attempted to switch to lru-cache but I had issues with it prematurely expiring entries.  If you don't object, I'd like to stick with node-cache for now.

benbp:
+1

PR #505, opened 2016-10-20
Ssh obm service

srinia6:
This PR Jenkins:depends on https://github.com/RackHD/on-core/pull/222 and Jenkins:depends on RackHD/on-tasks#361

In this PR:

```
Move sshSettings from nodes.
Create /ibms and /ibms/{identifier} like the obms .
Backward compatible with 1.1 like obms.
Schema/view for ibms.
unit-test fix
```

@RackHD/corecommitters @dalebremner @brianparry @BillyAbildgaard @geoff-reid

### DIFF
--- a/data/views/node.2.0.json
+++ b/data/views/node.2.0.json
@@ -50,5 +50,19 @@
     <% } %>
     <% } %>
     "type": "<%=type%>",
-    "workflows": "<%=basepath %>/nodes/<%=id%>/workflows" 
+    "workflows": "<%=basepath %>/nodes/<%=id%>/workflows",
+    <% if (ibms === null) { %>
+    "ibms": null,
+    <% } else { %>
+    "ibms":[
+    <% ibms.forEach(function (value, i, arr){ %>
+    {
+        "service": "<%=value.service%>",
+        "ref":"<%=basepath %>/ibms/<%=value.id%>"
+    }
+    <%= ( arr.length > 0 && i < arr.length-1 ) ? ',': '' %>
+    <% }); %>
+    ]
+    <% } %>
+
 }
--- a/lib/api/1.1/northbound/nodes.js
+++ b/lib/api/1.1/northbound/nodes.js
@@ -21,7 +21,7 @@ di.annotate(nodesRouterFactory, new di.Inject(
         '_',
         'Assert',
         'Errors'
-)
+    )
 );
 
 function nodesRouterFactory (
@@ -46,7 +46,7 @@ function nodesRouterFactory (
      * @param {object} node - node model instance
      * @return {object} Promise that fulfills to the rendered node
      */
-    function _renderNodeObmSettings(node) {
+    function _renderNodeMgmtSettings(node) {
         // This function exists to allow backward compatibility of the 1.1 API
         // with 2.0 API OBM and SSH resources.
         assert.object(node, 'node should be a node model instance.');
@@ -68,10 +68,10 @@ function nodesRouterFactory (
                     password: 'REDACTED'
                 };
             } else {
-                delete node['sshSettings'];
+                delete node.sshSettings;
             }
             return node;
-        })
+        });
     }
 
     /**
@@ -87,7 +87,7 @@ function nodesRouterFactory (
         return nodeApiService.getAllNodes(req.query)
         .then(function (nodes){
             return Promise.map(nodes, function(node) {
-                return _renderNodeObmSettings(node);
+                return _renderNodeMgmtSettings(node);
             });
         }); 
     }));
@@ -154,7 +154,7 @@ function nodesRouterFactory (
     router.get('/nodes/:identifier', rest(function (req) {
         return nodeApiService.getNodeById(req.params.identifier)
         .then(function (node){
-            return _renderNodeObmSettings(node);
+            return _renderNodeMgmtSettings(node);
         });
     }));
 
@@ -198,7 +198,7 @@ function nodesRouterFactory (
         })
         .then(function() {
             // lastly render obmSettings into the node.
-            return _renderNodeObmSettings(node);
+            return _renderNodeMgmtSettings(node);
         });
     }));
 
@@ -247,7 +247,7 @@ function nodesRouterFactory (
     router.get('/nodes/:identifier/obm', rest(function (req) {
         return nodeApiService.getNodeObmById(req.params.identifier)
         .then(function (node){
-            return _renderNodeObmSettings(node)
+            return _renderNodeMgmtSettings(node)
             .then(function (retNode){
                 return retNode.obmSettings;
             });
@@ -298,7 +298,7 @@ function nodesRouterFactory (
             });
         })
         .then(function() {
-            return _renderNodeObmSettings(node);
+            return _renderNodeMgmtSettings(node);
         });
     }));
 
@@ -344,7 +344,7 @@ function nodesRouterFactory (
     router.get('/nodes/:identifier/ssh', rest(function (req) {
         return nodeApiService.getNodeById(req.params.identifier)
         .then(function(node) {
-            return _renderNodeObmSettings(node)
+            return _renderNodeMgmtSettings(node)
             .then(function (retNode){
                 return retNode.sshSettings;
             });
@@ -390,7 +390,7 @@ function nodesRouterFactory (
             return waterline.ibms.upsertByNode(req.params.identifier, newBody);
         })
         .then(function() {
-            return _renderNodeObmSettings(node);
+            return _renderNodeMgmtSettings(node);
         });
     }, {
         renderOptions: { success: 201 }
--- a/lib/api/1.1/northbound/nodes.js
+++ b/lib/api/1.1/northbound/nodes.js
@@ -19,8 +19,9 @@ di.annotate(nodesRouterFactory, new di.Inject(
         'Logger',
         'Promise',
         '_',
-        'Assert'
-    )
+        'Assert',
+        'Errors'
+)
 );
 
 function nodesRouterFactory (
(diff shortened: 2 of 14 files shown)


rackhdmirror:

lib/api/1.1/northbound/nodes.js
```diff
@@ -34,30 +35,43 @@ function nodesRouterFactory (
...
+        .spread(function(obms, sshSettings) {
             node.obmSettings = obms;
+            if (sshSettings) {
+                node.sshSettings = {
+                    host: sshSettings.config.host,
+                    user: sshSettings.config.user,
+                    password: 'REDACTED'
+                };
+            } else {
+                delete node['sshSettings'];
```
srinia6: ignoring this.

srinia6:
@RackHD/corecommitters - the unit test failures are due to on-core dependency.

srinia6:
FYI the unit-test failure is due to on-core dependency.

lib/api/1.1/northbound/nodes.js
anhou: it's recommended style also keep consistent with other codes

lib/api/2.0/ibms.js
```diff
@@ -0,0 +1,118 @@
...
+// GET /api/2.0/ibms/identifier
+var ibmsGetById = controller(function(req) {
+    return waterline.ibms.needByIdentifier(req.swagger.params.identifier.value);
+});
+
+// PATCH /api/2.0/ibms/identifier
+var ibmsPatchById = controller( function(req) {
+    var ibmId = req.swagger.params.identifier.value;
+    var values = req.swagger.params.body.value;
+    return waterline.ibms.needByIdentifier(ibmId)
```
anhou: duplicated codes, it also exists in some other places for ibm and obm. two places  will be maintained, and easy to miss something.  it needs some optimization,  or move the common codes to /lib/services/xxxx. maybe it also could be waterline.[xxx].needByxxxx().

srinia6: @anhou : I can create a ibm service and move the code for Patch and Delete. But the obm calls the obms waterline collection and ibms calls its own collection. Trying to think if thats what you meant? And not sure what you mean if one place changes ? They are 2 different collections that inherit from base class under on-core/models.

brianparry: @anhou @srinia6 The duplicated code is in `ibmsGet`, `ibmsPut`, `ibmsGetById`, and `ibmsGetAll`.  These are all basically one-liners, does it really make sense to create another layer of abstraction for this?  The two verbs with some substance to them (patch and delete) are not common anyway.  They could be moved to a service, but that does not avoid duplication.

lib/api/2.0/ibms.js
```diff
@@ -0,0 +1,118 @@
...
+            .tap(function() {
+                /* Publish events of nodes got beofre update */
+                _.forEach(oldNodes, function(oldNode) {
+                    if (oldNode) {
+                        /* asynchronous, don't wait promise return for performance*/
+                        return waterline.nodes.getNodeById(oldNode.id)
+                        .then(function(newNode) {
+                            return eventsProtocol.publishNodeAttrEvent(oldNode, newNode, 'ibms');
+                        })
+                        .catch(function (error) {
```
brianparry: @srinia6 It's probably worth appending `error.message` to your base error message.

lib/api/2.0/ibms.js
```diff
@@ -0,0 +1,118 @@
...
+});
+
+// PATCH /api/2.0/ibms/identifier
+var ibmsPatchById = controller( function(req) {
+    var ibmId = req.swagger.params.identifier.value;
+    var values = req.swagger.params.body.value;
+    return waterline.ibms.needByIdentifier(ibmId)
+        .then(function(oldIbm) {
+            /* Get nodes that need to publish events */
+            if (oldIbm.node && !values.nodeId) {
```
brianparry: srinia6 Is this logic needed?  Couldn't we just compare the old node to the new node and alert if anything has changed?  Or maybe even just always alert?

lib/api/2.0/ibms.js
anhou: some of the codes are copied from /lib/services/obm-api-services.js   the code logic is the same. I think don't need another abstract layer to service api, if ibm and obm really almost the similar in document operation. maybe just a map/dict could reduce the duplication.  maintain obm/ibm in /lib/services/bm-api-service.js or lib/services/nm-api-service.js(network management)  all the document openration like waterline.obms.xxxx could be waterline.nm[type].xxx, just a method I could think of to de-duplication. you decide use it or not.

lib/api/2.0/ibms.js
anhou: I think this part of codes are copied from /lib/service/obm-api-service.js  they are written and tested by me several months ago. the  logic could be referenced from the unit test cases. the behavior is documented in RackHD/docs in 'event notification' section about obms.assigned/unassigned/updated

changev:
test this please

changev:
test this please

changev:
@srinia6 a strange issue.
When get description by github api
It get
```
This PR Jenkins:depends on https://github.com/RackHD/on-core/pull/222 and Jenkins:depends on RackHD/on-tasks#361
```
RackHD/on-tasks#361 but not a URL.

I add sub line to correct this
jenkins: depends on https://github.com/RackHD/on-tasks/pull/361

test this please

lib/api/1.1/northbound/nodes.js
```diff
@@ -34,28 +35,41 @@ function nodesRouterFactory (
...
-        return Promise.map(waterline.obms.find({node: node.id}), function(obm) {
-           return _.pick(obm.toJSON(), 'service', 'config');
-        })
-        .then(function(obms) {
+        node = _.omit(node.toJSON(), ['obms', 'ibms']);
+
+        var obms = Promise.map(waterline.obms.find({node: node.id}), function(obm) {
+            return _.pick(obm.toJSON(), 'service', 'config');
+        });
+        var sshSettings = waterline.ibms.findByNode(node.id, 'ssh-ibm-service');
```
benbp: Since this is only for a 1.1 API backwards compatibility edge case, I'm not going to raise an issue. But in any 2.0 controller I would object to hardcoding for specific IBM service types, especially just for doing 'REDACT' which shouldn't be handled at this layer anyways.

lib/api/2.0/ibms.js
brianparry: Thanks @anhou

yyscamper:
@srinia6 : this is a reminder for you. When I worked on the issue https://github.com/RackHD/RackHD/issues/483, I found the obm setting was not added while manually adding node, so the node's discovery will fail. The solution is in my PR https://github.com/RackHD/on-http/pull/524.

After the ibm is merged, and when you plan to move the `snmpSettings` into ibm, don't forget this issue.

static/schemas/ibms/ssh-ibm-service.json
```diff
@@ -0,0 +1,36 @@
...
+                    "type": "string"
+                },
+                "service": {
+                    "type": "string"
+                },
+                "config": {
+                    "type": "object",
+                    "properties": {
+                        "host": {
+                            "description": "BMC address",
```
yyscamper: This is not BMC address!

static/monorail-2.0.yaml
```diff
@@ -617,6 +625,271 @@ paths:
       - filesRead
       x-view: file.2.0.json
     x-swagger-router-controller: files
+  /ibms:
+    get:
+      description: 'Get a list of all In Band Management settings that have been associated
+
+        with nodes.
```
yyscamper: This multiple lines style is very strange for me. this line should be put together with line630, the same for a lot below lines.

data/views/node.2.0.json
```diff
@@ -50,5 +50,19 @@
     <% } %>
     <% } %>
     "type": "<%=type%>",
-    "workflows": "<%=basepath %>/nodes/<%=id%>/workflows" 
+    "workflows": "<%=basepath %>/nodes/<%=id%>/workflows",
+    <% if (ibms === null) { %>
```
yyscamper: A  minor suggestion, put the ibm together with obm will help people to compare the difference between them.

lib/api/1.1/northbound/nodes.js
yyscamper: Besides, the function name is `_renderNodeObmSettings` but the function covers both obm & ibm (ssh), so this doesn't match.

srinia6: @benbp : Thanks ben! @yyscamper : I will rename the function.

static/monorail-2.0.yaml
srinia6: @yyscamper Because this patch has been sitting in review for ~14 days, it needs to be rebased with current master, that got changed since then! This will go away after rebasing.

srinia6:
@yyscamper  : Looking at your PR, that is no way related to this PR. https://github.com/RackHD/on-http/pull/524 What issue are you talking about?

lib/api/1.1/northbound/nodes.js
```diff
@@ -19,8 +19,9 @@ di.annotate(nodesRouterFactory, new di.Inject(
         'Logger',
         'Promise',
         '_',
-        'Assert'
-    )
+        'Assert',
+        'Errors'
+)
```
srinia6: ok

data/views/node.2.0.json
srinia6: Since this has been tested with these changes, i plan to keep them as is for now. Will change them in the snmp story.

lib/api/1.1/northbound/nodes.js
srinia6: done

lib/api/2.0/ibms.js
srinia6: @anhou - we have created another story to handle slightly similar code in this story - https://www.pivotaltracker.com/story/show/133740825

lib/api/1.1/northbound/nodes.js
srinia6: done

static/schemas/ibms/ssh-ibm-service.json
srinia6: OK

static/monorail-2.0.yaml
srinia6: done

srinia6:
FYI - Jenkins build and CI failures are due to unavailability of on-core models.

PR #509, opened 2016-10-21
IME-63 To allow Setting up two network devices. These devices can be …

manfrednde:
…static or DHCP

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -34,6 +34,7 @@ rootpw <%=rootPlainPassword%>
 reboot
 
 %firstboot --interpreter=busybox
+
 postLookup () {
   echo "Attempting postLookup operation on $1" >> /vmfs/volumes/datastore1/firstboot.log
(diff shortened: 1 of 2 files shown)


rackhdmirror:

manfrednde:
The changes will allow user to set up network devices with static or dynamic ip address. it will also allow to create switch devices.
@RackHD/corecommitters @jimturnquist @johren

data/templates/esx-ks
```diff
@@ -171,6 +183,9 @@ cp /var/log/esxi_install.log "/vmfs/volumes/datastore1/firstboot-esxi_install.lo
       esxcli network ip interface remove -i <%=vmkname%>
       esxcli network ip interface add -i <%=vmkname%> -p <%=n.device%>
       esxcli network ip interface ipv4 set -i <%=vmkname%> -t dhcp
+      sleep 60
```
yyscamper: Could you please explain why there is a delay here?

manfrednde: To allow the interface to dhcp and gets an IP address before we can query that IP to post lookup.

PR #510, opened 2016-10-24
Add callback script to ubuntu installation workflow

keedya:
Dependency on PR https://github.com/RackHD/on-tasks/pull/366 https://github.com/RackHD/on-taskgraph/pull/178

### DIFF
--- a/data/templates/install-ubuntu/ubuntu-preseed
+++ b/data/templates/install-ubuntu/ubuntu-preseed
@@ -195,4 +195,6 @@ in-target wget http://<%=server%>:<%=port%>/api/current/templates/post-install-u
 in-target chmod +x ./post-install.sh; \
 in-target ./post-install.sh; \
 in-target wget http://<%=server%>:<%=port%>/api/current/templates/ubuntu-sources -O /etc/apt/sources.list; \
+in-target wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%> -O /etc/rc2.d/S99callback; \
+in-target chmod +x /etc/rc2.d/S99callback; \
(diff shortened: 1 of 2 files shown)


rackhdmirror:
Build finished.

keedya:
@RackHD/corecommitters @geoff-reid @tannoa2 @uppalk1

data/templates/install-ubuntu/ubuntu-preseed
```diff
@@ -195,4 +195,6 @@ in-target wget http://<%=server%>:<%=port%>/api/current/templates/post-install-u
 in-target chmod +x ./post-install.sh; \
 in-target ./post-install.sh; \
 in-target wget http://<%=server%>:<%=port%>/api/current/templates/ubuntu-sources -O /etc/apt/sources.list; \
+in-target wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%> -O /etc/rc2.d/S99callback; \
```
yyscamper: the name "S99callback" is a little odd to me, any special meaning? or we have to do like this?

yyscamper: I just searched my ubuntu machine, it also contains such kind of S99/S70 prefix, so this is OK for me.

anhou: rc2.d is specific to run level, all file under rc2.d is a symlink to ./init.d. maybe it could be optimized here?

keedya:
@anhou, the callback script is temporary and it will be deleted upon running after the first boot. 
The rules for startup scripts is as u mentioned but in this case, We only need it once

PR #511, opened 2016-10-25
Return original swagger-validator error messages to client

benbp:
Resolves https://github.com/RackHD/RackHD/issues/388
Resolves https://github.com/RackHD/RackHD/issues/445

The built-in swagger_validator fitting that we use as part of the swagger-express-mw package returns a detailed array of error messages, but we don't actually return that information to the user to help with debugging basic validation errors.

This change updates the API to return data resembling what was proposed in https://github.com/RackHD/RackHD/issues/388. I'm open to suggestions about how we want to structure this, since there's a lot of data (like error codes) that I'm sanitizing out of the final return value.

Example 1:
```
{
    "UUID": "e5786dfe-1cfe-4dd2-b94e-e3192f958f96",
    "errors": [
        "Invalid Content-Type (application/octet-stream).  These are supported: application/json"
    ],
    "message": "Validation errors",
    "status": "400"
}
```

UPDATE: This now also bubbles up JSON schema violation errors, in order to address https://github.com/RackHD/RackHD/issues/445:

```
{
    "UUID": "6fb7e671-fb4f-4de0-911f-a121a03ff574",
    "errors": [
        "Missing required property: service",
        "Missing required property: nodeId"
    ],
    "message": "Validation errors",
    "status": "400"
}
```

Also, the test files that existed for our swagger fittings were named wrong, so they weren't picked up by the mocha test runner. I made some minor changes to get the tests working again as well, mostly by restricting the scope of the unit tests to not rely on the swagger-api-service renderer code pathway, which is tested elsewhere.

@brianparry I feel like I'm missing a spot somewhere since I'm updating the error view. Is there another location or api client change that should go along with this addition to the return data structure?

@RackHD/corecommitters @pscharla @dalebremner


rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

lib/fittings/error_handler.js
```diff
@@ -3,8 +3,8 @@
 var util = require('util');
 var http = require('http');
 
-module.exports = function create() {
-    var injector = require('../../index.js').injector;
+module.exports = function create(_, __, injector) {
```
yyscamper: one lodash `_` and two lodashes `__`! why? Couldn't use other meaningful name?

yyscamper:
@benbp : I am curious whether the detail error reporting can fix this issue as well? https://github.com/RackHD/RackHD/issues/445

PengTian0:
Jenkins: test this please

benbp:
@yyscamper it seems like it would, if the errors are coming from the swagger_validator fitting (which seems like they are). I'll do a test with the API calls you've detailed and let you know.

lib/fittings/error_handler.js
benbp: I'm using the underscores as placeholders for unused arguments that get passed into this function by bagpipes. I need to be able to pass in the injector during unit tests somehow, but also need the code to treat injector as undefined in production. The general convention I've seen is to use underscores for unused function arguments, though I agree the namespace collision with lodash is bad, so I've switched it to double and triple underscore.

benbp:
@yyscamper updated, see the new PR description for an example of accurate JSON schema error reporting.

yyscamper:
@benbp : Thanks for your feedback.

brianparry:
Thanks @benbp!  I just noticed this PR now.

To answer your question, we'll also need to update the swagger definition: https://github.com/RackHD/on-http/blob/master/static/monorail-2.0.yaml#L14.  This definition is used to build client libraries.

benbp:
@brianparry thanks, will do.

PR #518, opened 2016-11-01
enhance the debian init script to support ubuntu 16.04

PengTian0:
1. This PR removed the "--force".
    Because ubuntu 16.04 doesn't support the option.
2. This PR added time interval after starting and stopping service. 
    Because sometimes the service won't start but will later when running the command "service start" 

@yyscamper @anhou @panpan0000


PengTian0:
Jenkins: test this please

PengTian0:
Jenkins: test this please

PR #519, opened 2016-11-01
Fixed centos-ks for 6.5 max payload install

geoff-reid:
@keedya @tannoa2 @uppalk1 @RackHD/corecommitters

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -45,14 +45,14 @@ clearpart --all --drives=<%=installDisk%>
         <% if(partition.mountPoint !== undefined && partition.size !== undefined) { %>
             <% if(partition.fsType !== undefined) { %>
                 <% if(partition.size === 'auto') { %>
-                    partition <%=partition.mountPoint%> --grow --fstype=<%=partition.fsType%>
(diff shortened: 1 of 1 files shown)


geoff-reid:
test this please

data/templates/centos-ks
```diff
@@ -45,14 +45,14 @@ clearpart --all --drives=<%=installDisk%>
         <% if(partition.mountPoint !== undefined && partition.size !== undefined) { %>
             <% if(partition.fsType !== undefined) { %>
                 <% if(partition.size === 'auto') { %>
-                    partition <%=partition.mountPoint%> --grow --fstype=<%=partition.fsType%>
+                    partition <%=partition.mountPoint%> --size=250 --grow --fstype=<%=partition.fsType%>
```
yyscamper: @geoff-reid : could you please teach me why you set the minimum size to 250? What's the problem you are trying to resolve?

yyscamper: Add @anhou to review, since @anhou is the author of this code.

geoff-reid: --size is required for centOS 6, the install will fail if --size is not specified.  Since --grow is also specified it will configure the partition to use all available space with 250 being the min.

PR #521, opened 2016-11-03
return empty object when access api without token or with error token

PengTian0:
I tried to access API  without token or with a wrong token and just got an empty object.
Cause:
The returned info object is a list.
But the code uses it as a dictionary which contains "message".


iceiilin:
:+1:

PengTian0:
@cgx027 @yyscamper @anhou @panpan0000

lanchongyizu:
:+1: after fix houndci problems.

PengTian0:
Jenkins: test this please

PR #523, opened 2016-11-04
Add taskId rendering feature and implement centos installation progre…

pengz1:
Jenkins depends on: https://github.com/RackHD/on-core/pull/227, https://github.com/RackHD/on-tasks/pull/370, https://github.com/RackHD/on-core/pull/227, https://github.com/RackHD/on-taskgraph/pull/189

This PR includes changes for CentOS installation progress feature, major points include:
1. Embed notification API in ipxe/ks files
2. Enable taskId rendering in profile/templates
3. Create new notification API to get notification from iPXE files.

### DIFF
--- a/data/profiles/install-centos.ipxe
+++ b/data/profiles/install-centos.ipxe
@@ -3,7 +3,10 @@ set base-url <%=repo%>/images/pxeboot
 set params initrd=initrd.img ks=<%=installScriptUri%> hostname=<%=hostname%> ksdevice=bootif BOOTIF=01-${netX/mac} console=<%=comport%>,115200n8 console=tty0
 kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
 initrd ${base-url}/initrd.img
+
+# Below two lines are designed for progress info, downloading action itself is meaningless.
 initrd http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=steps%>
 imgfree <%=steps%>
+
 boot || prompt --key 0x197e --timeout 2000 Press F12 to investigate || exit shell
 
--- a/data/profiles/install-centos.ipxe
+++ b/data/profiles/install-centos.ipxe
@@ -3,5 +3,7 @@ set base-url <%=repo%>/images/pxeboot
 set params initrd=initrd.img ks=<%=installScriptUri%> hostname=<%=hostname%> ksdevice=bootif BOOTIF=01-${netX/mac} console=<%=comport%>,115200n8 console=tty0
 kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
 initrd ${base-url}/initrd.img
+initrd http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=steps%>
+imgfree <%=steps%>
 boot || prompt --key 0x197e --timeout 2000 Press F12 to investigate || exit shell
 
--- a/data/profiles/install-centos.ipxe
+++ b/data/profiles/install-centos.ipxe
@@ -5,7 +5,7 @@ kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
 initrd ${base-url}/initrd.img
 
 # Below two lines are designed for progress info, downloading action itself is meaningless.
-imgfetch --name fakedimage http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=totalSteps%>
+imgfetch --name fakedimage http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=totalSteps%>/2
 imgfree fakedimage
 
 boot || prompt --key 0x197e --timeout 2000 Press F12 to investigate || exit shell
--- a/data/profiles/install-centos.ipxe
+++ b/data/profiles/install-centos.ipxe
@@ -5,8 +5,8 @@ kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
 initrd ${base-url}/initrd.img
 
 # Below two lines are designed for progress info, downloading action itself is meaningless.
-initrd http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=steps%>
-imgfree <%=steps%>
+imgfetch --name fakedimage http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=totalSteps%>
+imgfree fakedimage
 
 boot || prompt --key 0x197e --timeout 2000 Press F12 to investigate || exit shell
 
--- a/data/profiles/install-centos.ipxe
+++ b/data/profiles/install-centos.ipxe
@@ -4,8 +4,7 @@ set params initrd=initrd.img ks=<%=installScriptUri%> hostname=<%=hostname%> ksd
 kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
 initrd ${base-url}/initrd.img
 
-# Below two lines are designed for progress info, downloading action itself is meaningless.
-imgfetch --name fakedimage http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=totalSteps%>/2
+imgfetch --name fakedimage http://<%=server%>:<%=port%>/api/current/notification/progress?taskId=<%=taskId%>&totalSteps=<%=totalSteps%>&currentStep=2
 imgfree fakedimage
 
 boot || prompt --key 0x197e --timeout 2000 Press F12 to investigate || exit shell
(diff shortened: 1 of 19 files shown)


pengz1:
@RackHD/corecommitters @iceiilin @cgx027

pengz1:
test this please

data/profiles/install-centos.ipxe
```diff
@@ -3,5 +3,10 @@ set base-url <%=repo%>/images/pxeboot
 set params initrd=initrd.img ks=<%=installScriptUri%> hostname=<%=hostname%> ksdevice=bootif BOOTIF=01-${netX/mac} console=<%=comport%>,115200n8 console=tty0
 kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
 initrd ${base-url}/initrd.img
+
+# Below two lines are designed for progress info, downloading action itself is meaningless.
+initrd http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=steps%>
```
cgx027: I understand that's no CURL/wget tool for accessing an URL in iPXE environment, but use ```imgfetch``` instead of ```initrd``` to access url and using ```--name uselessoutput``` may help make the code more readable.

data/profiles/install-centos.ipxe
```diff
@@ -3,5 +3,10 @@ set base-url <%=repo%>/images/pxeboot
 set params initrd=initrd.img ks=<%=installScriptUri%> hostname=<%=hostname%> ksdevice=bootif BOOTIF=01-${netX/mac} console=<%=comport%>,115200n8 console=tty0
 kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
 initrd ${base-url}/initrd.img
+
+# Below two lines are designed for progress info, downloading action itself is meaningless.
```
cgx027: May help readers if adding comments here, like: ```There are no curl or wget tools for accessing an URL so initrd/fetchimg is used instead```

data/profiles/install-centos.ipxe
```diff
@@ -3,5 +3,10 @@ set base-url <%=repo%>/images/pxeboot
...
+initrd http://<%=server%>:<%=port%>/api/current/notification/<%=taskId%>/<%=steps%>
+imgfree <%=steps%>
```
cgx027: Adding comments like ```ipxe will run the fetched image and hang. Using imgfree to remove the useless output``` may help readers.

data/templates/centos-ks
```diff
@@ -99,6 +101,7 @@ net-tools
 %end
 
 %pre
+/usr/bin/curl -X POST -H 'Content-Type:application/json' -d "{\"taskId\": \"<%=taskId%>\", \"progress\": { \"description\": \"Installer started, starting OS installtion\", \"totalSteps\": \"<%=steps%>\", \"currentStep\": 3}}" http://<%=server%>:<%=port%>/api/current/notification;
```
cgx027: line tooooo long?

data/profiles/install-centos.ipxe
iceiilin: would it avoid confusion if renaming "steps" to some phrase like "totalSteps"? It needs the variable in on-tasks to change as well.

lib/services/profiles-api-service.js
```diff
@@ -285,15 +286,22 @@ function profileApiServiceFactory(
                     }
                 })
                 .then(function(profile) {
-                    return [profile, taskProtocol.requestProperties(node.id)];
+                    return [
+                        profile, 
+                        taskProtocol.requestProperties(node.id), 
+                        taskProtocol.activeTaskExists(node.id)
```
iceiilin: There is already a same AMQP request before calling this function (https://github.com/pengz1/on-http/blob/47bdcb17e75e8a9fcf647154ba5523985a90dcc8/lib/services/profiles-api-service.js#L117). To avoid another chance of AMQP resource consumption (it might help in @scale environment), would it be possible to leverage the result got before?

lib/api/2.0/notification.js
```diff
@@ -12,6 +12,31 @@ var notificationPost = controller({success: 201}, function(req) {
...
 
+/**
+ * @api {get} /api/2.0/notification/:taskId/:steps
+ * @apiDescription notification get api, used to notify RackHD ipxe initrd downloading finished
+ *  :taskId: active taskId
+ *  :steps: task steps count
+ * @apiName notification-get
+ * @apiGroup notification
+ */
+var notificationIdGet = controller(function(req, res){
```
cgx027: Would suggest change function name from ```notificationIdGet``` to a simpler ```notificationGet``` or ```notifyProgressByTaskId```(since this function will only be used for sending progress notifications) or ```notifyProgressByGet```  or some other names that will reflect more precisely on it's function.

lib/api/2.0/notification.js
```diff
@@ -12,6 +12,31 @@ var notificationPost = controller({success: 201}, function(req) {
...
+    return notificationApiService.postNotification({
+        taskId: req.swagger.params.taskId.value,
+        progress: {
+            //percentage: "20%",
+            totalSteps: _.parseInt(req.swagger.params.steps.value) || 2,
+            currentStep: 2,
```
cgx027: The same magic number used another time, suggest to define it as a variable or move to constant service.

static/monorail-2.0-sb.yaml
```diff
@@ -156,6 +156,43 @@ paths:
           description: Unexpected error
           schema:
             $ref: '#/definitions/Error'
+  /notification/{taskId}/{steps}:
+    x-swagger-router-controller: notification
+    get:
+      operationId: notificationIdGet
+      summary: |
+        get notification via identifier
```
cgx027: maybe consider ```send notification by HTTP get with task ID```

lib/api/2.0/notification.js
```diff
@@ -12,6 +12,31 @@ var notificationPost = controller({success: 201}, function(req) {
...
+            description: "iPXE initrd download done, starting initiating installer"
```
cgx027: Although this API by now is only used by posting initrd notification, I would suggest to make it a more generic API. By the URL ```/api/2.0/notification/:taskId/:steps``` has no indication that this API is magically binded to initrd notifcation. 

Consider pass the ```initrd``` information as a parameter to the API so the handler.

lib/services/notification-api-service.js
```diff
@@ -80,35 +80,54 @@ function NotificationApiServiceFactory(
...
             if (!_.isEmpty(task) && _.has(task, 'graphId')){
                 return waterline.graphobjects.findOne({instanceId: task.graphId})
                 .then(function(graphObject){
-                    //TODO: workflow progress percentage should be designed
+                    if (_.has(progress, "totalSteps") &&
+                        _.has(progress, "currentStep")) {
+                        progress.progressRate = progress.currentStep +
+                            '/' + progress.totalSteps;
+                        progress = _.omit(progress, ['totalSteps', 'currentStep']);
+                    } else if(!_.has(progress, "percentage")) {
```
cgx027: maybe like this:

```
if(!validProgressObj()) {
//at here handle no valid progress data error
}

//at here put the rest of the code where progressData is constructed.
```

lib/services/profiles-api-service.js
cgx027: I would think put nodeId into context so that it can be leveraged everywhere

changev:
test this please

changev:
test this please

data/templates/centos-ks
```diff
@@ -17,8 +17,10 @@ selinux --permissive
 services --enabled=NetworkManager,sshd
 network --device=<%=macaddress%> --noipv6 --activate
 
-#enable syslog
-logging --host=<%=server%> --level=info
+# enable syslog
+<% if (remoteLogging) { %>
```
yyscamper: Since remoteLogging is optional, if remoteLogging is not passed in parameter, your code will report unreference error. so Please copy the parameter `kvm` and change to `<% if (typeof remoteLogging !== 'undefined' && remoteLogging) { %>1.

changev:
test this please

changev:
test this please

pengz1:
test this please

benbp:
If I'm understanding this code correctly, we already support rendering a taskId via `<%=task.instanceId%>` (see https://github.com/RackHD/on-tasks/blob/master/lib/task.js#L132), though I agree from a usability standpoint that's not really clear or consistent with the vocabulary we use in our public facing APIs and docs. I think it's much better to keep all task related rendering contained to the Task class itself rather than scattering it through different projects. Would doing something like this fit your needs, or is there a piece of this I'm mis-understanding?

From https://github.com/RackHD/on-tasks/blob/master/lib/task.js#L132
```
        self.renderContext = {
            // other stuff
            taskId: self.instanceId  // <- added
            // other stuff
        };
```

pengz1:
@benbp 
I tried add taskId as you suggested, it looks it is difficult to deliver this parameters. As a compromise, I added this parameters in swagger-api-service to combine two changes in profile/template-api-services.

PengTian0:
test this please

PR #524, opened 2016-11-04
Add obm setting while posting a new node

yyscamper:
This is a part of fix for https://github.com/RackHD/RackHD/issues/483

Root cause: After the obm is moved to a standalone model, we forget to add the obm setting while posting a new node, so some node catalog job will fail.

BTW: the unit-test error due to the node-cache is fixed as well in my PR.

@RackHD/corecommitters @iceiilin @cgx027 @pengz1 @manfrednde

### DIFF
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -249,18 +249,33 @@ function nodeApiServiceFactory(
         return Promise.try(function() {
             query = waterline.nodes.find(query);
 
-            if (options.skip) query.skip(options.skip);
-            if (options.limit) query.limit(options.limit);
+            if (options.skip) {
+                query.skip(options.skip);
+            }
+            if (options.limit) {
+                query.limit(options.limit);
+            }
 
             return query.populate('obms');
         });
     };
 
     NodeApiService.prototype.postNode = function(body) {
-        return Promise.resolve().then(function() {
-            return waterline.nodes.create(body);
+        //for api-v1.1 back-compatibility, the input obm can be either via obms or obmSettings
+        var nodeBody = _.omit(body, ['obms', 'obmSettings']);
+        var obmBody = body.obms || body.obmSettings || null;
+
+        return Promise.resolve()
+        .then(function() {
+            return waterline.nodes.create(nodeBody);
         }).tap(function(node) {
             return eventsProtocol.publishNodeEvent(node, 'added');
+        }).tap(function(node) {
+            if (obmBody) {
+                return Promise.map(obmBody, function(obm) {
+                    return waterline.obms.upsertByNode(node.id, obm);
+                });
+            }
         }).then(function(node) {
             if(node.type === Constants.NodeTypes.Switch &&
                node.snmpSettings && node.autoDiscover) {
@@ -287,7 +302,7 @@ function nodeApiServiceFactory(
                 );
             }
             else if(node.type === Constants.NodeTypes.Mgmt &&
-                    node.obmSettings && node.autoDiscover) {
+                    obmBody && node.autoDiscover) {
                 var configuration = {
                     name: 'Graph.MgmtSKU.Discovery',
                     options: {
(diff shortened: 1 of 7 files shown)


spec/lib/services/upnp-service-spec.js
```diff
@@ -40,13 +40,13 @@ describe("UPnP Service", function() {
...
     };
-    
+
     before(function() {
         helper.setupInjector([
             helper.require("/lib/services/upnp-service"),
             helper.di.simpleWrapper(messenger, 'Services.Messenger'),
             helper.di.simpleWrapper(rx,'Rx'),
-            helper.di.requireWrapper('node-cache', 'node-cache')
+            helper.di.requireWrapper('node-cache', 'node-cache', undefined, __dirname)
```
yyscamper: without this change, the unit-test will always fail due to following error:
```
1) UPnP Service "before all" hook:
     Error: dihelper incapable of finding specified file for require filename:node-cache, directories:(/home/rackhd/src/on-core/spec, /home/rackhd/src/on-core/lib)
      at _require (/home/rackhd/src/on-core/lib/di.js:492:13)
      at Object.requireWrapper (/home/rackhd/src/on-core/lib/di.js:518:13)
      at Context.<anonymous> (spec/lib/services/upnp-service-spec.js:49:23)
```

iceiilin:
:+1:

srinia6:
@yyscamper What is the issue here? When you post a node, you don't see obmSettings, they come-up as obms. Can you plese explain?

lib/services/nodes-api-service.js
```diff
@@ -249,18 +249,36 @@ function nodeApiServiceFactory(
...
+                throw new Error('Unknown node type "' + body.type + '"');
+            }
+            assert.optionalArrayOfObject(
+                body.obmSettings, 'obmSettings should be an array of object');
+        })
+        .then(function() {
             return waterline.nodes.create(body);
         }).tap(function(node) {
             return eventsProtocol.publishNodeEvent(node, 'added');
+        }).tap(function(node) {
```
yyscamper: @srinia6: I examined the schema for `POST /nodes` (see https://github.com/RackHD/on-http/blob/master/static/monorail-2.0.yaml#L83), it doesn't tell how to specify the obm when manually adding a node.
And by looking at the rest code of `postNode` function, you will it still use the `obmSettings` (see https://github.com/yyscamper/on-http/blob/b31f6b199676172de8ddb37c0504de6df1cac197/lib/services/nodes-api-service.js#L308).

Anyway, we always need a way to let user specify the obm while manually adding a node, previously the obm document is bind with node, so it's pretty straightforward, now the obm is moved out, the situation is a little hard. If let user manually add the obm before creating node, but you don't know the node id; If bind obm document with node, then there is some what mismatch with our internal structure.

So @srinia6 @brianparry : What's your preference for how to specify the obm while adding a node?

lib/services/nodes-api-service.js
```diff
@@ -249,18 +249,36 @@ function nodeApiServiceFactory(
...
+            }
 
             return query.populate('obms');
         });
     };
 
     NodeApiService.prototype.postNode = function(body) {
-        return Promise.resolve().then(function() {
+        return Promise.try (function() {
+            assert.string(body.type, 'node type is required');
```
brianparry: @yyscamper This should be handled in the schema; we should be adding `type` to the list of required properties: https://github.com/RackHD/on-http/blob/master/static/schemas/2.0/node.2.0.json#L54.  You shouldn't need to validate the type since we have a enum in the schema with all node types.  The new `obms` request option should be added to the nodes schema as an array of objects.  The tricky part here is that we should be referencing the obm schema (https://github.com/RackHD/on-http/tree/master/static/schemas/obms) here to validate the obms themselves.

yyscamper:
@brianparry : Your comment has been resolved. Actually this is more difficult that I thought to move the validation code into schema. The challenge is:
(1) Those obm schemas are defined in different namespace.
(2) The schema for noop-obm-service is so free so that any arbitrary attributes can be put into `config`, so if I combines all obm schema with `allOf`, I tested arbitrary data can pass the validation.

So my solution is:
(1) Use the full schema name (namespace + filename) to reference the existing obm schemas. 
(2) Set `additionalProperties" to `false` to obm schemas, so that the `anyOf` can take effect to screen out some invalid input.

yyscamper:
@brianparry : As we talked last week. I have removed the newly created obm schema, as a follow-up, the schema validation need to be moved into the swagger fitting. Please review this PR, whether this is OK or not.

PR #526, opened 2016-11-07
Add powershell command to disable firewall 

uppalk1:
* to be able to test the link after the static IP is set as part of max payload, need to disable firewall

### DIFF
--- a/data/templates/unattend_server2012.xml
+++ b/data/templates/unattend_server2012.xml
@@ -109,8 +109,10 @@
                 </SynchronousCommand>
                 <SynchronousCommand wcm:action="add">
                     <Order>2</Order>
+                    <% if( firewallDisable ) { %>
                     <CommandLine>powershell -Command "Set-NetFirewallProfile -Profile Domain,Public,Private -Enabled False" </CommandLine>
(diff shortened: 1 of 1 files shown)


data/templates/unattend_server2012.xml
```diff
@@ -109,12 +109,18 @@
                 </SynchronousCommand>
                 <SynchronousCommand wcm:action="add">
                     <Order>2</Order>
+                    <CommandLine>powershell -Command "Set-NetFirewallProfile -Profile Domain,Public,Private -Enabled False" </CommandLine>
+                    <Description>disable firewall on node to be able to ping the new IP when max payload is used</Description>
+                    <RequiresUserInput>false</RequiresUserInput>
```
yyscamper: I have some security concern for disable the firewall forever. I think the disable of firewall is to bypass some problem so that the install windows can work well. For a short while, I think it is OK, but I hope it can be enabled finally.

geoff-reid:
What if you were to make disabling the firewall an option for the install job?  This way we could specify an option to disable the firewall for the Jenkins tests, however the default setting for other users would be to leave the firewall enabled.

uppalk1:
Creating a new PR....

PR #527, opened 2016-11-08
create ssh keys at startup for Esxi

lanchongyizu:
It's related with ODR-259.
create ssh keys for Esxi at startup since it can't be persistent.

@RackHD/corecommitters @panpan0000 @iceiilin


iceiilin:
:+1:

lanchongyizu:
test this please.

PR #536, opened 2016-11-17
Fix exit code handling in bootstrap.js

jeamland:
When we ask bootstrap.js to exit we pass the value as the `exit` member of the tasks structure. Problem is that bootstrap.js seems to expect it to be `exit.code`. I've fixed this defensively as I'm not sure if there are prior installations that expect the `exit.code` value.


benbp:
:+1:

tldavies:
👍

PR #537, opened 2016-11-18
Windows os regression script enhancement

uppalk1:
* Add firewallDisable to let ping function test the new static IP

### DIFF
--- a/data/templates/unattend_server2012.xml
+++ b/data/templates/unattend_server2012.xml
@@ -109,8 +109,10 @@
                 </SynchronousCommand>
                 <SynchronousCommand wcm:action="add">
                     <Order>2</Order>
+                    <% if( firewallDisable ) { %>
                     <CommandLine>powershell -Command "Set-NetFirewallProfile -Profile Domain,Public,Private -Enabled False" </CommandLine>
                     <Description>disable firewall on node to be able to ping the new IP when max payload is used</Description>
+                    <% } %>
                     <RequiresUserInput>false</RequiresUserInput>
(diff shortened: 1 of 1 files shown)


data/templates/unattend_server2012.xml
```diff
@@ -109,12 +109,20 @@
                 </SynchronousCommand>
                 <SynchronousCommand wcm:action="add">
                     <Order>2</Order>
+                    <% if( firewallDisable ) { %>
+                    <CommandLine>powershell -Command "Set-NetFirewallProfile -Profile Domain,Public,Private -Enabled False" </CommandLine>
```
geoff-reid: I think you need this line and the next line to be enclosed with <% %>, otherwise the firewall is going to be disabled for all installs.

data/templates/unattend_server2012.xml
```diff
@@ -109,12 +109,20 @@
                 </SynchronousCommand>
                 <SynchronousCommand wcm:action="add">
                     <Order>2</Order>
+                    <% if( firewallDisable ) { %>
```
geoff-reid: should add check for undefined value

uppalk1:
@geoff-reid firewallDisable defaults to false. If I check for undefined I need to remove the default value or else xml script complains.

uppalk1:
test this please...

data/templates/unattend_server2012.xml
geoff-reid: sounds good.  I retract my comment.

PR #230, opened 2016-04-15
Add file latest and file md5 latest to south API router

WangWinson:
To solve issue https://github.com/RackHD/on-http/issues/215
- Add file latest and file md5 latest router to south file router
- Add unit test

@RackHD/corecommitters @cgx027

PR #538, opened 2016-11-28
Fix error where InternalServerError is called with a Error object

cgx027:
Jenkins depends on: https://github.com/RackHD/on-core/pull/230 

## Background
on-core PR #230 fixed a typo on errors.js where InternalServerError should be a subclass of BaseError, thus InternalServerError should be called with a string parameter. 

## The problem
In lib/services/profiles-api-service.js Lin 384, InternalServerError is mistakenly with an Error object. On-http unit test fails because of this. Here is the Jenkins log: http://rackhdci.lss.emc.com/job/on-http/2881/

## The Fix: 
Change parameter from Object to string.

### DIFF
--- a/lib/services/profiles-api-service.js
+++ b/lib/services/profiles-api-service.js
@@ -381,7 +381,7 @@ function profileApiServiceFactory(
             })
             .catch(function (err) {
                 if (!err.status) {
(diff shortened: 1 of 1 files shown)


changev:
:+1:  Fixed an ancient issue!

PR #540, opened 2016-12-09
Mount host dev in the micro container

keedya:
- Required for task driveId
- Adding  -v=dev:/dev to the docker run command


anhou:
It happened that previously I used the same method to access /dev/* directory :+1:

PR #541, opened 2016-12-16
Inject GRUB customizations to support CoreOS Ignition

tgelter:
Add support to CoreOS GRUB configuration to support bare-metal Ignition templating support.

### DIFF
--- a/data/templates/ignition.json
+++ b/data/templates/ignition.json
@@ -1,0 +1,11 @@
+{
+  "ignition": { "version": "2.0.0" },
+  "storage": {
+    "files": [{
+      "filesystem": "root",
+      "path": "/root/ignition-RackHD.txt",
+      "mode":400,
+      "contents": { "source": "data:text/plain;charset=utf-8;base64,WW91IG1heSBjdXN0b21pemUgeW91ciBDb3JlT1MgaW5zdGFsbGF0aW9uIHVzaW5nIElnbml0aW9uIGJ5IGNvbmZpZ3VyaW5nIHRoZSBpZ25pdGlvblNjcmlwdFVyaSB2YXJpYWJsZSB0byBwb2ludCB0byBhIGN1c3RvbSBJZ25pdGlvbiBmaWxlIHByaW9yIHRvIGV4ZWN1dGluZyB0aGUgUmFja0hEIFRhc2suT3MuSW5zdGFsbC5Db3JlT1MgdGFzay4=" }
+    }]
+  }
+}
--- a/data/templates/ignition.json
+++ b/data/templates/ignition.json
@@ -5,7 +5,10 @@
       "filesystem": "root",
       "path": "/root/ignition-RackHD.txt",
       "mode":400,
-      "contents": { "source": "data:text/plain;charset=utf-8;base64,WW91IG1heSBjdXN0b21pemUgeW91ciBDb3JlT1MgaW5zdGFsbGF0aW9uIHVzaW5nIElnbml0aW9uIGJ5IGNvbmZpZ3VyaW5nIHRoZSBpZ25pdGlvblNjcmlwdFVyaSB2YXJpYWJsZSB0byBwb2ludCB0byBhIGN1c3RvbSBJZ25pdGlvbiBmaWxlIHByaW9yIHRvIGV4ZWN1dGluZyB0aGUgUmFja0hEIFRhc2suT3MuSW5zdGFsbC5Db3JlT1MgdGFzay4=" }
+      "contents": {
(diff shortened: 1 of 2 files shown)


brianparry:
ok to test

data/templates/install-coreos.sh
```diff
@@ -3,5 +3,11 @@ set -e
 CLOUD_CONFIG_FILE=pxe-cloud-config.yml
 curl -o $CLOUD_CONFIG_FILE http://<%=server%>:<%=port%>/api/current/templates/$CLOUD_CONFIG_FILE?nodeId=<%=nodeId%>
 sudo coreos-install -d <%=installDisk%> -c $CLOUD_CONFIG_FILE -b <%=repo%>
+
+# Customizations for supporting CoreOS Ignition:
+mkdir /mnt/coreos
+mount <%=installDisk%>6 /mnt/coreos/
```
yyscamper: I think the additional number `6` should be removed

tgelter: Hello @yyscamper,
The number `6` is required, although I can see how it may look like a typo.
Please see [the CoreOS partition table](https://coreos.com/os/docs/latest/sdk-disk-partitions.html#partition-table). The change to `grub.cfg` that we make then gets picked up by CoreOS at first boot. Let me know if I can offer any more clarification.
Thanks!
Tim

anhou: @tgelter so it looks like either an annotation or a variable whose name could hint what '6' means is needed. could be OEM_PARTITION or any meaningful name you think.

yyscamper: @tgelter : Thanks for you explanation, surely a strange writing here. Please at least add some comment here to make sure other people will not have my same feeling of typo.

Meanwhile, whether the number 6 partition table is specifically for your own system? Because as a general solution, I'd suggest adding one more parameter (such as `"partitionTableNumber"`) so it fits to more user cases.

changev:
test this please

tgelter:
I've pushed some changes to this branch per feedback from @yyscamper in the paired PR: https://github.com/RackHD/on-tasks/pull/377
I've tested them and found them to be functioning as intended. Please review & test as necessary prior to merge.
Thanks!

data/templates/ignition.json
```diff
@@ -0,0 +1,11 @@
+{
+  "ignition": { "version": "2.0.0" },
+  "storage": {
+    "files": [{
+      "filesystem": "root",
+      "path": "/root/ignition-RackHD.txt",
+      "mode":400,
+      "contents": { "source": "data:text/plain;charset=utf-8;base64,WW91IG1heSBjdXN0b21pemUgeW91ciBDb3JlT1MgaW5zdGFsbGF0aW9uIHVzaW5nIElnbml0aW9uIGJ5IGNvbmZpZ3VyaW5nIHRoZSBpZ25pdGlvblNjcmlwdFVyaSB2YXJpYWJsZSB0byBwb2ludCB0byBhIGN1c3RvbSBJZ25pdGlvbiBmaWxlIHByaW9yIHRvIGV4ZWN1dGluZyB0aGUgUmFja0hEIFRhc2suT3MuSW5zdGFsbC5Db3JlT1MgdGFzay4=" }
```
yyscamper: I am wondering what's this long (seems random) string means?

tgelter: Hello @yyscamper, that string is base64 encoded (per guidance in the Ignition documentation). When decoded (feel free to verify if you're concerned), it says the following:

> You may customize your CoreOS installation using Ignition by configuring the ignitionScriptUri variable to point to a custom Ignition file prior to executing the RackHD Task.Os.Install.CoreOS task.

Please let me know if I can help clarify anything else.

yyscamper: @tgelter : 

Thanks for you explaintation.
my suggestion is adding some comment to describe what this base64 encoded string means. It's a pity that JSON doesn't support comment, but I see the ejs supports, please check this help document: https://www.npmjs.com/package/ejs

The example is:
```javascript
{
    ...
   "mode": 400,
   "contents": {
     <%#  The source is encoded with base64, after decoded, it says: %>
     <%#  You may customize your CoreOS installation using Ignition by configuring the ignitionScriptUri variable to point to a custom Ignition file prior to executing the RackHD Task.Os.Install.CoreOS task. %>
     <%#  If you want to change the source, you can use this command tool: xxxx %>
      "source": "data:test/plain;charset=utf-8;base64,xxxxxxxxxxxxxxxxxxxxxxxx"
   }
}
```

tgelter: Again, much appreciated @yyscamper. I've updated this file with the additional information you've requested. Please let me know if anything else is missing.

yyscamper:
Thanks @tgelter , this is perfect.

PR #542, opened 2016-12-16
Compare RoleId value to existing roles and only accepts known roles

lacarb:
Presently, redfish users can be created with any string given as their RoleId. This change ensures that users can only be made with known role values taken from getRoleByName()

### DIFF
--- a/lib/api/redfish-1.0/account-service.js
+++ b/lib/api/redfish-1.0/account-service.js
@@ -64,7 +64,13 @@ var createAccount = controller({success: 201}, function(req, res) {
         assert.string(payload.UserName);
         assert.string(payload.Password);
         assert.string(payload.RoleId);
-        return accountService.listUsers();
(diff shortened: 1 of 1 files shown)


lacarb:
test this please

lib/api/redfish-1.0/account-service.js
```diff
@@ -64,10 +64,17 @@ var createAccount = controller({success: 201}, function(req, res) {
         assert.string(payload.UserName);
         assert.string(payload.Password);
         assert.string(payload.RoleId);
+        return accountService.listRoles().filter(function(roleObj) {
```
lacarb: That's much better, thank you for pointing it out.

DavidjohnBlodgett:
👍

lacarb:
@RackHD/corecommitters 
Small commit here. Looking for a single +1 or feedback.

PR #547, opened 2017-01-03
RAC-3801 ESXi 6.0 OS installation failure when network interface is s…

manfrednde:
…et to dhcp

This code change is to solve the issue where the network interface is specified in the payload and it is set to dhcp. The payload may look like the following in this case:

{
    "options": {
        "defaults": {
            "version": "6.0",
            "repo": "http://172.31.128.1:9080/esxi/esxi6",
            "networkDevices":[{"device": "vmnic2"}]
        }
    }
}
 It looks like when the host try to download the call back script the link is not ready. So I included a retry option in the kickstart script. 

@RackHD/corecommitters 

@stuart-stanley 

@derrickostertag 

@johren

### DIFF
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -216,10 +216,10 @@ for retry in $(seq 1 60);
 do
     wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%>?nodeId=<%=nodeId%> -O /etc/rc.local.d/local.sh
     if [ $? -eq 0 ]; then
-        echo "The call back script was downloaded successfully."
+        logger "The call back script was downloaded successfully after $retry attempt(s)."
         break
     else
-        echo "Failed to download call back script, retrying..."
+        logger "Failed to download call back script after $retry attempt(s)."
         sleep 1
     fi
 done;
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -216,10 +216,10 @@ for retry in $(seq 1 60);
 do
     wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%>?nodeId=<%=nodeId%> -O /etc/rc.local.d/local.sh
     if [ $? -eq 0 ]; then
-        logger "The call back script was downloaded successfully after $retry attempt(s)."
+        logger -p user.info "RackHD's call back script was downloaded successfully after $retry attempt(s)."
         break
     else
-        logger "Failed to download call back script after $retry attempt(s)."
+        logger -p user.info "Failed to download RackHD's call back script after $retry attempt(s)."
         sleep 1
     fi
 done;
--- a/data/templates/esx-ks
+++ b/data/templates/esx-ks
@@ -209,7 +209,20 @@ cp /var/log/esxi_install.log "/vmfs/volumes/datastore1/firstboot-esxi_install.lo
 #
 # NOTE: this script will execute right away as a result of writing it to local.sh
 # along with executing on every subsequent boot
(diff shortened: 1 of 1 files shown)


manfrednde:
@jimturnquist

data/templates/esx-ks
```diff
@@ -209,7 +209,20 @@ cp /var/log/esxi_install.log "/vmfs/volumes/datastore1/firstboot-esxi_install.lo
 #
 # NOTE: this script will execute right away as a result of writing it to local.sh
 # along with executing on every subsequent boot
-wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%>?nodeId=<%=nodeId%> -O /etc/rc.local.d/local.sh
+#
+# Try to download call back script 60 times 1 second
+# sleep in between to allow link to be up after DHCP
+for retry in $(awk 'BEGIN { for ( i=0; i<60; i++ ) { print i; } }');
```
jimturnquist: This iteration technique seems overly complex.  How about:

for retry in $(seq 1 60); do
    ...
    ...
done

jimturnquist: So, the question is if busy box supports the above iteration pattern.  I will check...

jimturnquist: I just ran under busy box and the iteration form above worked fine.

jimturnquist: Busy box v1.21.1

yyscamper:
@manfrednde change logger to echo is better for me as well.  Do you have ever tried if on-syslog is down, whether the `logger` command will fail and continue to fail the whole installation? 

Meanwhile, I'd suggest you use the `-p` to specify the logger level, so that user not feel it is a bug while seeing "Failed to download call back script after 1 attempt"

```
-p, --priority priority
              Enter the message into the log with the specified priority.  The priority may be specified numerically or as a facility.level pair.  For example, -p local3.info logs the message
              as informational in the local3 facility.  The default is user.notice.
```

manfrednde:
@yyscamper 
@jimturnquist 

Please review above change

jimturnquist:
Could we catch the case where the retries were exhausted and report via user.err?  Something like:

if [ $retry -eq 60 ]; then
    logger -p user.err "Failed to download RackHD's call back script, kickstart failed"
fi

manfrednde:
Thanks Jim I will add that case but executing the kickstart file will not fail

manfrednde:
@jimturnquist 
@yyscamper 
Please review new change

data/templates/esx-ks
```diff
@@ -209,7 +209,23 @@ cp /var/log/esxi_install.log "/vmfs/volumes/datastore1/firstboot-esxi_install.lo
...
+for retry in $(seq 1 60);
+do
+    wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%>?nodeId=<%=nodeId%> -O /etc/rc.local.d/local.sh
+    if [ $? -eq 0 ]; then
+        logger -p user.info "RackHD's call back script was downloaded successfully after $retry attempt(s)."
+        break
+    else
+        logger -p user.info "Failed to download RackHD's call back script after $retry attempt(s)."
+        sleep 1
+        if [ $retry -eq 60 ]; then
```
jimturnquist: You should be able to just check it once after the loop is finished instead of every iteration.

manfrednde:
@jimturnquist 
@yyscamper

PengTian0:
test this please

manfrednde:
@johren 

Can you please check the above changes and merge if all good ?

Thank you 
Manfred

PR #548, opened 2017-01-04
Fixed returned value of cli command in catalog config

nucklehead:
this command `cli('show startup-config')` returns the output as a string  so using `[0]` will only get the first character.

### DIFF
--- a/data/templates/cisco-catalog-config.py
+++ b/data/templates/cisco-catalog-config.py
@@ -8,8 +8,8 @@ def main():
     data = {}
 
     try:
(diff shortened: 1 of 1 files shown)


yyscamper:
jenkins: ok to test

yyscamper:
@nucklehead : I am just curious why such kind of simple error isn't detected at the very beginning.

nucklehead:
@yyscamper not sure. But the task will run successfully but it will show that the startup-config and running-conig in the catalogs to be `"\n"`.

PR #551, opened 2017-01-05
Rac 1984 node/id/workflow returns appropriate response codes

lacarb:
The 2.0 api for node/{identifier}/workflows presently returns an empty array and 200 no matter what it is passed.
The proposed solution is taken from 1.1 which rejects bad identifiers
```
{
  "message": "Could not find node with identifier BAD_ID",
  "status": "404",
  "UUID": "03effd02-4f75-480e-aa5b-0759a59de585"
}
```
and responds with the workflows run against a node if the ID matches.

The method `WorkflowApiService.prototype.getWorkflowsByNodeId` was deleted because it was only used once in the 2.0 method which never worked as intended.


PengTian0:
test this please

lib/api/2.0/nodes.js
```diff
@@ -7,7 +7,7 @@ var controller = injector.get('Http.Services.Swagger').controller;
 var addLinks = injector.get('Http.Services.Swagger').addLinksHeader;
 var nodes = injector.get('Http.Services.Api.Nodes');
 var waterline = injector.get('Services.Waterline');
-var workflowApiService = injector.get('Http.Services.Api.Workflows');
+//var workflowApiService = injector.get('Http.Services.Api.Workflows');
```
yyscamper: Please directly remove this.

lacarb:
test this please

tldavies:
test this please

lacarb:
test this please

lib/api/2.0/nodes.js
```diff
@@ -61,26 +60,7 @@ var nodesGetPollersById = controller(function(req) {
 });
 
 var nodesGetWorkflowById = controller(function(req) {
-    var newQuery;
```
brianparry: Removing this code breaks the `?active=true` and `?active=false` query parameters.  These were explicitly changed from the /active endpoint in the 1.1 API because, at the time, there was some confusion about the semantics.  Adding the query makes things more explicit.

lib/api/2.0/nodes.js
```diff
@@ -61,26 +60,7 @@ var nodesGetPollersById = controller(function(req) {
...
-        }
-        newQuery = _.merge({}, newQuery, req.query);
-    } else {
-        newQuery = req.query;
-    }
-
-    return workflowApiService.getWorkflowsByNodeId(req.swagger.params.identifier.value,
-                                                   newQuery);
-
+    return nodes.getNodeWorkflowById(req.swagger.params.identifier.raw);
```
brianparry: The nodes service can be used here in place of the workflow service, but it needs to be modified to accept a query.  Otherwise, this is an API breaking change.  As far as I can tell, the only thing that's actually needed to fix this bug is a call to `waterline.nodes.needByIdentifier` prior to the graphobjects query.

lacarb:
test this please

lib/api/2.0/nodes.js
```diff
@@ -78,9 +77,7 @@ var nodesGetWorkflowById = controller(function(req) {
         newQuery = req.query;
     }
 
-    return workflowApiService.getWorkflowsByNodeId(req.swagger.params.identifier.value,
-                                                   newQuery);
-
+    return nodes.getNodeWorkflowById(req.swagger.params.identifier.value.newQuery);
```
brianparry: @lacarb I think this is a typo.  Shouldn't this be `req.swagger.params.identifier.value, newQuery` rather than `req.swagger.params.identifier.value.newQuery`?

lacarb:
There is a single CIT test that is failing. It fails because it was made based on the old implementation of this command.
The code here should be sound. I'll make corrections to this CIT test and open a relating PR.

lacarb:
test this please

lacarb:
test this please

lacarb:
test this please

PengTian0:
test this please

lacarb:
@brianparry, we're green on Jenkins due to another PR I got through. Can this be merged or do you still have feedback? Thanks!

pscharla:
@brianparry : Can you merge this one for us?

anhou:
@lacarb  this PRs break CI test, and it seems not only one place need to be updated in test, I'll revert it temperally to let CI pass.

PR #554, opened 2017-01-09
Hook API design

pengz1:
Hook is required in RI-46: https://rackhd.atlassian.net/browse/RI-46
This PR is to implement RackHD hook API design for user. User can use those APIs to review, update or delete hooks managed by RackHD.
POST/GET/DELETE/PATCH API methods are implemented for hooks.

### DIFF
--- a/lib/api/2.0/hooks.js
+++ b/lib/api/2.0/hooks.js
@@ -1,0 +1,55 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var injector = require('../../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var hooks = injector.get('Http.Services.Api.Hooks');
+
+/**
+ * @api {get} /api/2.0/hooks GET /
+ * @apiDescription get list of hooks
+ * @apiName hooks-get
+ * @apiGroup hooks
+ */
+var hooksGetAll = controller(function(req) {
+    return hooks.getHooks(req.query);
+});
+
+/**
+ * @api {POST} /api/2.0/hooks POST /
+ * @apiDescription post a new hook
+ * @apiName hooks-post
+ * @apiGroup hooks
+ */
+var hooksPost = controller({success: 201}, function(req) {
+    return hooks.createHook(req.body);
+});
+
+/**
+ * @api {PATCH} /api/2.0/hooks/:identifier POST /:identifier
+ * @apiDescription patch an existing hook
+ * @apiName hooks-patch
+ * @apiGroup hooks
+ */
+var hooksPatchById = controller(function(req) {
+    return hooks.updateHookById(req.swagger.params.identifier.value, req.body);
+});
+
+/**
+ * @api {DELETE} /api/2.0/hooks/:identifier DELETE /:identifier
+ * @apiDescription delete an existing hook
+ * @apiName hooks-delete
+ * @apiGroup hooks
+ */
+var hooksDelById = controller({success: 204}, function(req) {
+    return hooks.deleteHookById(req.swagger.params.identifier.value);
+});
+
+module.exports = {
+    hooksGetAll: hooksGetAll,
+    hooksPost: hooksPost,
+    hooksPatchById: hooksPatchById,
+    hooksDelById: hooksDelById
+};
+
--- a/lib/api/2.0/hooks.js
+++ b/lib/api/2.0/hooks.js
@@ -1,4 +1,4 @@
-// Copyright 2016, EMC, Inc.
+// Copyright 2017, Dell EMC, Inc.
 
 'use strict';
 
@@ -14,6 +14,16 @@ var hooks = injector.get('Http.Services.Api.Hooks');
  */
 var hooksGetAll = controller(function(req) {
     return hooks.getHooks(req.query);
+});
+
+/**
+ * @api {get} /api/2.0/hooks/:identifier GET /:identifier
+ * @apiDescription get list of hooks
+ * @apiName hooks-get
+ * @apiGroup hooks
+ */
+var hooksGetById = controller(function(req) {
+    return hooks.getHookById(req.swagger.params.identifier.value);
 });
 
 /**
@@ -50,6 +60,7 @@ module.exports = {
     hooksGetAll: hooksGetAll,
     hooksPost: hooksPost,
     hooksPatchById: hooksPatchById,
-    hooksDelById: hooksDelById
+    hooksDelById: hooksDelById,
+    hooksGetById: hooksGetById
 };
 
--- a/lib/services/hooks-api-service.js
+++ b/lib/services/hooks-api-service.js
@@ -1,0 +1,91 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = hookApiServiceFactory;
+di.annotate(hookApiServiceFactory, new di.Provide('Http.Services.Api.Hooks'));
+di.annotate(hookApiServiceFactory,
+    new di.Inject(
+        'Services.Waterline',
+        'Logger',
+        '_',
+        'Errors'
+    )
+);
+function hookApiServiceFactory(
+    waterline,
+    Logger,
+    _,
+    Errors
+) {
+    var logger = Logger.initialize(hookApiServiceFactory);
+    function HooksApiService() {
+    }
+
+    /**
+     * Get hooks
+     * @return {Promise}
+     */
+    HooksApiService.prototype.getHooks = function (query) {
+        return waterline.hooks.find(query);
+    };
+
+    /**
+     * Create new hooks
+     * @param {Object} [req] HTTP request
+     * @param {Object} [res] HTTP response
+     * @return
+     */
+    HooksApiService.prototype.createHook = function(body) {
+        var self = this;
+        return waterline.hooks.findOne({url: body.url})
+        .then(function(entry){
+            if(_.isEmpty(entry)) {
+                return waterline.hooks.create(body)
+                .then(function(hook){
+                    return hook;
+                });
+            } else if (body.filters && !_.isEqual(body.filters, entry.filters)) {
+                var filters = entry.filters || [];
+                filters = _.uniq(filters.concat(body.filters));
+                entry.filters = filters;
+                return waterline.hooks.update(entry)
+                .then(function(){
+                    return entry;
+                });
+            } else {
+                var err = new Errors.BaseError('duplicate hook found');
+                err.status = 409;
+                throw err;
+            }
+        });
+    };
+
+    /**
+     * Delete hook by identifier
+     * @param {String} id hook idenfier
+     * @return
+     */
+    HooksApiService.prototype.deleteHookById = function(id) {
+        var self = this;
+        return waterline.hooks.destroyByIdentifier(id);
+    };
+
+    /**
+     * Update hook by identifier
+     * @param {String} id hook idenfier
+     * @param {Object} body hook body to be updated
+     * @return
+     */
+    HooksApiService.prototype.updateHookById = function(id, body) {
+        var self = this;
+        return waterline.hooks.updateByIdentifier(id, body)
+        .then(function(hook){
+            return hook;
+        });
+    };
+
+    return new HooksApiService();
+}
--- a/lib/services/hooks-api-service.js
+++ b/lib/services/hooks-api-service.js
@@ -1,4 +1,4 @@
-// Copyright 2016, EMC, Inc.
+// Copyright 2017, Dell EMC, Inc.
 
 'use strict';
 
@@ -33,28 +33,22 @@ function hookApiServiceFactory(
     };
 
     /**
+     * Get hook by identifier
+     * @return {Promise}
+     */
+    HooksApiService.prototype.getHookById = function (id) {
+        return waterline.hooks.findOne({id: id});
+    };
+    /**
      * Create new hooks
-     * @param {Object} [req] HTTP request
-     * @param {Object} [res] HTTP response
+     * @param {Object} body HTTP request body
      * @return
      */
     HooksApiService.prototype.createHook = function(body) {
-        var self = this;
         return waterline.hooks.findOne({url: body.url})
         .then(function(entry){
             if(_.isEmpty(entry)) {
-                return waterline.hooks.create(body)
-                .then(function(hook){
-                    return hook;
-                });
-            } else if (body.filters && !_.isEqual(body.filters, entry.filters)) {
-                var filters = entry.filters || [];
-                filters = _.uniq(filters.concat(body.filters));
-                entry.filters = filters;
-                return waterline.hooks.update(entry)
-                .then(function(){
-                    return entry;
-                });
+                return waterline.hooks.create(body);
             } else {
                 var err = new Errors.BaseError('duplicate hook found');
                 err.status = 409;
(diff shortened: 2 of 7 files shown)


pengz1:
@yyscamper @iceiilin @lanchongyizu @anhou

lib/api/2.0/hooks.js
```diff
@@ -0,0 +1,55 @@
+// Copyright 2016, EMC, Inc.
```
yyscamper: Now it is 2017, and also change EMC to "Dell EMC"

lib/services/hooks-api-service.js
```diff
@@ -0,0 +1,91 @@
...
+     * @return
+     */
+    HooksApiService.prototype.createHook = function(body) {
+        var self = this;
+        return waterline.hooks.findOne({url: body.url})
+        .then(function(entry){
+            if(_.isEmpty(entry)) {
+                return waterline.hooks.create(body)
+                .then(function(hook){
+                    return hook;
```
yyscamper: Since the `hook` is returned directly, so line 47-49 is not necessary.

lib/services/hooks-api-service.js
```diff
@@ -0,0 +1,91 @@
...
+                });
+            } else if (body.filters && !_.isEqual(body.filters, entry.filters)) {
+                var filters = entry.filters || [];
+                filters = _.uniq(filters.concat(body.filters));
+                entry.filters = filters;
+                return waterline.hooks.update(entry)
```
yyscamper: `POST` is always to create a new resource, not update/modify the existing resource, so the logic about `hooks.update` should be removed.

lib/services/hooks-api-service.js
```diff
@@ -0,0 +1,91 @@
...
+                var err = new Errors.BaseError('duplicate hook found');
+                err.status = 409;
+                throw err;
+            }
+        });
+    };
+
+    /**
+     * Delete hook by identifier
+     * @param {String} id hook idenfier
```
yyscamper: typo: idenfier ==> identifier

lib/services/hooks-api-service.js
```diff
@@ -0,0 +1,91 @@
...
+     * Update hook by identifier
+     * @param {String} id hook idenfier
+     * @param {Object} body hook body to be updated
+     * @return
+     */
+    HooksApiService.prototype.updateHookById = function(id, body) {
+        var self = this;
+        return waterline.hooks.updateByIdentifier(id, body)
+        .then(function(hook){
+            return hook;
```
yyscamper: line 85-87 is not necessary here.

lib/services/hooks-api-service.js
```diff
@@ -0,0 +1,91 @@
...
+     * @return {Promise}
+     */
+    HooksApiService.prototype.getHooks = function (query) {
+        return waterline.hooks.find(query);
+    };
+
+    /**
+     * Create new hooks
+     * @param {Object} [req] HTTP request
+     * @param {Object} [res] HTTP response
```
yyscamper: This JSDoc doesn't match with the implementation.

static/monorail-2.0.yaml
```diff
@@ -60,6 +60,23 @@ definitions:
...
       username:
         type: string
+  hooks.2.0_HooksBase:
+    description: Hook template
+    type: object
+    properties:
+      url:
+        type: string
+        format: uri
+      contentType:
```
yyscamper: What's the `contentType`? I think it should be removed.

static/monorail-2.0.yaml
```diff
@@ -650,6 +667,130 @@ paths:
           schema:
             $ref: '#/definitions/Error'
 
+  /hooks:
+    get:
+      description: Get the RackHD hooks configuration properties.
```
yyscamper: Suggest refactor the description "Get all hooks", the word "RackHD" and "configuration" is not necessary. The same for other description.

static/monorail-2.0.yaml
```diff
@@ -650,6 +667,130 @@ paths:
...
+        name: body
+        required: true
+        schema:
+          $ref: '#/definitions/hooks.2.0_HooksBase'
+      responses:
+        201:
+          description: Successfully created new hooks
+          schema:
+            type: object
+        default:
```
yyscamper: There is another error code is 409, it will be returned if the input url is duplicated.

static/monorail-2.0.yaml
```diff
@@ -650,6 +667,130 @@ paths:
...
+      - jwt
+      x-privileges:
+      - Write
+      - hooksWrite
+      x-swagger-schema: hooks.2.0.json#/definitions/HookBase
+    x-view: hooks.2.0.json
+    x-swagger-router-controller: hooks 
+  /hooks/{identifier}:
+    delete:
+      description: Delete a hooks based on uuid.
```
yyscamper: At least at present, the hook id is not uuid, it is mongod id.

static/monorail-2.0.yaml
```diff
@@ -650,6 +667,130 @@ paths:
...
+      summary: Delete a hook
+      tags:
+      - /api/2.0
+      x-authentication-type:
+      - jwt
+      x-privileges:
+      - Remove
+      - hooksRemove
+    patch:
+      description: Modify the properties of a hooks.
```
yyscamper: typo: `hooks` ==> `hook`

static/schemas/2.0/hooks.2.0.json
```diff
@@ -0,0 +1,42 @@
...
+        },
+        "HookArray": {
+            "description": "A hook list",
+            "type": "array",
+            "items": {
+                "$ref": "#/definitions/HookBase"
+            },
+            "uniqueItems": true
+        },
+        "Hooks": {
```
yyscamper: What this used for? I don't see any user case, why not remove it?

static/monorail-2.0.yaml
```diff
@@ -60,6 +60,23 @@ definitions:
         type: string
       username:
         type: string
+  hooks.2.0_HooksBase:
```
yyscamper: @brianparry : I need your confirm here. Since we have created a schema file (schemas/2.0/hooks.2.0.json), so whether this inline schema is still needed?

spec/lib/api/2.0/hooks-spec.js
```diff
@@ -0,0 +1,99 @@
...
+            .post('/api/2.0/hooks')
+            .send({})
+            .expect(400)
+            .expect(function () {
+                expect(hookService.createHook).to.have.not.been.called;
+            });
+    });
+
+    it('should patch hook', function () {
+        sandbox.stub(hookService, 'updateHookById').resolves(hookSample[0]);
```
yyscamper: It can be better if the stub only takes effect when the input identifier is equal to `test`. so that if the hook identifier is not correct, the code can detect such error. The example is:
```
sandbox.stub(hookService, 'updateHookById').withArgs('test').resolves(hookSample[0]);
```

The same for delete.

spec/lib/services/hooks-api-service-spec.js
```diff
@@ -0,0 +1,113 @@
...
+        .then(function(result) {
+            expect(waterline.hooks.findOne).to.be.calledOnce;
+            expect(waterline.hooks.findOne).to.be.calledWith({url: body.url});
+            expect(waterline.hooks.create).to.be.calledOnce;
+            expect(waterline.hooks.create).to.be.calledWith(body);
+            expect(result).to.be.deep.equal(hook);
+        });
+    });
+
+    it('should update hooks filters if hook url exist', function () {
```
yyscamper: In this case, directly throw error.

spec/lib/services/hooks-api-service-spec.js
```diff
@@ -0,0 +1,113 @@
...
+            expect(result).to.be.deep.equal(newHook);
+        });
+    });
+
+    it('should throw errors if new hook is identical with existing one', function (done) {
+        var existingHook = _.defaults(body, {id: 'test'});
+        this.sandbox.stub(waterline.hooks, 'findOne').resolves(existingHook);
+        return hooksApiService.createHook(body)
+        .then(function(){
+            throw new Error("test should fail");
```
yyscamper: Don't use throw, use `done(new Error("test should fail"))`

static/monorail-2.0.yaml
```diff
@@ -650,6 +667,130 @@ paths:
...
+      summary: Post a hook
+      tags:
+      - /api/2.0
+      x-authentication-type:
+      - jwt
+      x-privileges:
+      - Write
+      - hooksWrite
+      x-swagger-schema: hooks.2.0.json#/definitions/HookBase
+    x-view: hooks.2.0.json
```
yyscamper: typo: The real view file is `hook.2.0.json` not `hooks.2.0.json`! Remove the extra `s`

lib/api/2.0/hooks.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @api {get} /api/2.0/hooks GET /
+ * @apiDescription get list of hooks
+ * @apiName hooks-get
+ * @apiGroup hooks
+ */
+var hooksGetAll = controller(function(req) {
+    return hooks.getHooks(req.query);
+});
+
```
yyscamper: You forgot another API : ` GET /hooks/:identifier`, GET the specific hook by identifier.

spec/lib/api/2.0/hooks-spec.js
```diff
@@ -0,0 +1,99 @@
...
+            .expect('Content-Type', /^application\/json/)
+            .expect(200, hookSample)
+            .expect(function (res) {
+                expect(hookService.getHooks).to.have.been.calledOnce;
+                expect(hookService.getHooks).to.have.been.calledWith({});
+            });
+    });
+
+    it('should post new hooks', function () {
+        sandbox.stub(hookService, 'createHook').resolves(hookSample[0]);
```
yyscamper: In the viewer ( hook.2.0.json ), the `id` is required, but the stubbed return value doesn't contain a `id`, I think the unit-test will fail.

lib/services/hooks-api-service.js
```diff
@@ -0,0 +1,91 @@
...
+     * @return
+     */
+    HooksApiService.prototype.createHook = function(body) {
+        var self = this;
+        return waterline.hooks.findOne({url: body.url})
+        .then(function(entry){
```
brianparry: What happens if there are multiple hooks with the same name?  Looking at this PR: https://github.com/RackHD/on-core/pull/238/files it doesn't appear that the `name` property is uniquely indexed.

lib/services/hooks-api-service.js
```diff
@@ -0,0 +1,91 @@
...
+        return waterline.hooks.find(query);
+    };
+
+    /**
+     * Create new hooks
+     * @param {Object} [req] HTTP request
+     * @param {Object} [res] HTTP response
+     * @return
+     */
+    HooksApiService.prototype.createHook = function(body) {
```
brianparry: I'd suggest changing the name of this method to something more indicative of its function.  Perhaps `upsertHook` or 'createOrModifyHook` would be a better choice.

static/monorail-2.0.yaml
brianparry: This schema is still needed for the swagger UI and documentation.  It is not needed for validation although the validator will still validate against it.

static/monorail-2.0.yaml
```diff
@@ -650,6 +665,162 @@ paths:
...
+            $ref: '#/definitions/Error'
+      summary: Patch a hook
+      tags:
+        - /api/2.0
+      x-authentication-type:
+        - jwt
+      x-privileges:
+        - Write
+        - hooksModify
+      x-swagger-schema: hooks.2.0.json#/definitions/HookBase
```
brianparry: The way this is structured, patch will always need to have an `url` property.  I'd suggest modifying `HookBase` to have no required properties and then use an `allOf` to add required properties in another definition.  The definition containing `allOf` would be used for post and `HookBase` would be used for patch.  See the `Node` and `PartialNode` definitions here: https://github.com/RackHD/on-http/blob/master/static/schemas/2.0/node.2.0.json for an example.

pengz1: @brianparry 
Schema is updated as suggested, I created a new schema for post only named as HookPost

static/monorail-2.0.yaml
```diff
@@ -650,6 +670,162 @@ paths:
...
+            $ref: '#/definitions/Error'
+      summary: Post a hook
+      tags:
+      - /api/2.0
+      x-authentication-type:
+      - jwt
+      x-privileges:
+      - Write
+      - hooksWrite
+      x-swagger-schema: hooks.2.0.json#/definitions/HookBase
```
brianparry: @pengz1 This also needs to be change to `hooks.2.0.json#/definitions/HookPost`

static/monorail-2.0.yaml
brianparry: @pengz1 Thanks!  I think you missed one reference to `HookPost` (see my other comment), but looks good otherwise.

static/monorail-2.0.yaml
```diff
@@ -650,6 +670,162 @@ paths:
...
+          schema:
+            $ref: '#/definitions/Error'
+      summary: Get all hooks
+      tags:
+      - /api/2.0
+      x-authentication-type:
+      - jwt
+      x-privileges:
+      - Read
+      - hooksRead
```
pengz1: I tried to filter timestamp here.

brianparry: @pengz1 @iceiilin `x-view` is needed for any route that returns a hook object in the response body.

static/monorail-2.0.yaml
```diff
@@ -650,6 +670,162 @@ paths:
...
 
+  /hooks:
+    get:
+      description: Get all hooks
+      operationId: hooksGetAll
+      responses:
+        200:
+          description: Successfully retrieved the hook configuration
+          schema:
+            type: object
```
brianparry: @pengz1 Shouldn't this be an array with items of type `HookBase`?

static/monorail-2.0.yaml
```diff
@@ -650,6 +670,162 @@ paths:
...
+        in: body
+        name: body
+        required: true
+        schema:
+            $ref: '#/definitions/hooks.2.0_HookPost'
+      responses:
+        201:
+          description: Successfully created new hook
+          schema:
+            type: object
```
brianparry: @pengz1 Shouldn't this be an array with items of type HookBase?

static/monorail-2.0.yaml
```diff
@@ -650,6 +670,162 @@ paths:
...
+        in: body
+        name: body
+        required: true
+        schema:
+          $ref: '#/definitions/hooks.2.0_HookBase'
+      responses:
+        200:
+          description: Successfully modified the hook
+          schema:
+            type: object
```
brianparry: @pengz1 Shouldn't this be an array with items of type HookBase?

static/monorail-2.0.yaml
```diff
@@ -650,6 +670,162 @@ paths:
...
+      - description: id of the hook you wish to delete
+        in: path
+        name: identifier
+        required: true
+        type: string
+      responses:
+        200:
+          description: Successfully retrieved the specified hook
+          schema:
+            type: object
```
brianparry: @pengz1 Shouldn't this be an array with items of type HookBase?

static/monorail-2.0.yaml
brianparry: `x-view` property is needed here.

static/monorail-2.0.yaml
brianparry: @pengz1 Looks good.  Thank you!

changev:
test this please

pengz1:
@brianparry , could you help to approve this PR?

PR #555, opened 2017-01-10
Split existing events to internal events and external events

sunnyqianzhang:
This code review depends on the code changes in on-core:
https://github.com/RackHD/on-core/pull/239
@iceiilin @pengz1 @anhou @yyscamper


lib/services/upnp-service.js
```diff
@@ -158,7 +158,7 @@ function UPnPServiceFactory(
     };
     
     UPnPService.prototype.sendSSDPAlert = function(message) {
-        return messenger.publish(
+        return messenger.publishInternalEvents(
```
pengz1: messenger.publish should be move to lib/protocol/events.js

PR #557, opened 2017-01-10
 Refactor progress code

pengz1:
* This PR is to address Felix's comments on ipxe progress
   API,_getActiveTask function and progress data type.
 * Code is udpated based on comments.
 * No feature impacted.

### DIFF
--- a/lib/services/swagger-api-service.js
+++ b/lib/services/swagger-api-service.js
@@ -70,7 +70,7 @@ function swaggerFactory(
     }
 
     function _sortQuery(query){
-        var sortParams = new Object();
+        var sortParams = {};
         var match = query.raw.match(/([-+]{0,1})(.+)/);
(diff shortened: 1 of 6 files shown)


pengz1:
@yyscamper @lanchongyizu @linlynn

lib/services/swagger-api-service.js
```diff
@@ -356,6 +355,8 @@ function swaggerFactory(
                 .then(function (activeTask) {
                     return activeTask.taskId;
                 });
+        } else {
```
brianparry: Just a minor suggestion, but I think this would be a little easier to read if written like:
```
return Promise.try(function() {
    if (nodeId) {
        return taskProtocol.activeTaskExists(nodeId)
        .then(function (activeTask) {
             return activeTask.taskId;
        });
});
```

PR #559, opened 2017-01-11
Add skipPollerOption to determine creating or skipping 

keedya:
Add a new option in the config file to allow user to skip the generation of pollers on a compute node


geoff-reid:
:+1:

PR #566, opened 2017-01-18
only push notification for rack nodes

VulpesArtificem:
related to/requires https://github.com/RackHD/docs/pull/367

cleaned up the code so that we only publish this event for rack nodes added via api

@rolandpoulter @pscharla


anhou:
@VulpesArtificem after merging the docs, just found it's only for rack node, could it extend to other types of nodes? is there some concerns about this?
 please rebase the codes, otherwise, the houndci would fail, recently, @yyscamper has a fix about the houndci.

lib/services/nodes-api-service.js
```diff
@@ -316,6 +314,9 @@ function nodeApiServiceFactory(
                 };
                 return workflowApiService.createAndRunGraph(configuration);
             }
+            else if(node.type === Constants.NodeTypes.Rack) {
+                return eventsProtocol.publishNodeEvent(node, 'added');
```
yyscamper: I still don't understand why only publish notification for Rack node?

VulpesArtificem: We can do it for all nodes, that's what it did do, I was just changing it to actually meet the requirements of the original work item but keeping it for all nodes is fine. If we want that I can just close the PR

anhou: @VulpesArtificem I think keep it for all types of nodes are good than just meet the 'rack' node requirement, because current codes has cover the 'rack' related event requirement， so I suggest close this PR, for docs, I'm submitting a PR for event refactoring, I can modify the docs there. you don't need to submit another PR for docs

PR #567, opened 2017-01-18
add the identifier property to the node schema

tannoa2:


static/schemas/2.0/node.2.0.json
```diff
@@ -47,6 +47,10 @@
                         "type": "string",
                         "uniqueItems": true
                     }
+                },
+                "identifiers": {
+                    "description": "Node Macs",
+                    "type": "array"
```
yyscamper: The identifier is a list of string, so copy the property `tags`, and the sub schema for each array item:
```
items: {
  type: "string"
}
```

tannoa2:

keedya:
test this please

PR #573, opened 2017-01-26
Cisco Version Header string contains extra escape characters

nucklehead:
that json encoding escapes on and renders incorrectly.
This removes those characters and allows correct encoding.

Signed-off-by: Ethan Kaley <ethan.kaley@emc.com>

### DIFF
--- a/data/templates/cisco-catalog-version.py
+++ b/data/templates/cisco-catalog-version.py
@@ -8,7 +8,10 @@ def main():
     data = {}
 
     try:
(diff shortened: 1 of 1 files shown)


nucklehead:
@keedya this is a fix from the issue you and @king-jam identified weeks back.

king-jam:
:+1: was tested against a Cisco switch

keedya:
looks good. Thanks @nucklehead

PR #575, opened 2017-01-29
Update monorail-2.0.yaml

BillyAbildgaard:
updating yaml so new documentation is generated in bintray showing new ibms routes


BillyAbildgaard:
@RackHD/veyron

PR #580, opened 2017-02-03
updating postNode w/ new snmpSettings

BillyAbildgaard:
changing post node so it doesn't use old snmpSettings from node 

@RackHD/veyron


lib/services/nodes-api-service.js
```diff
@@ -276,26 +277,36 @@ function nodeApiServiceFactory(
                     return waterline.obms.upsertByNode(node.id, obm);
                 });
             }
+        }).tap(function(node) {
+            if (snmpBody) {
+                var snmpIbm = {
+                    "service": "snmp-ibm-service",
+                    "config": snmpBody
+                };
+                    return waterline.ibms.upsertByNode(node.id, snmpIbm);
```
anhou: indent

PR #582, opened 2017-02-06
prevent render error if networkDevices is not defined

geoff-reid:
This is to fix the failure with windows OS install, https://rackhd.atlassian.net/browse/RAC-4035

@RackHD/corecommitters @tannoa2 @keedya @uppalk1


keedya:
test this please

PR #589, opened 2017-02-15
Fix redfish/manager/id to return data, when no nodes are discovered

srinia6:
Fix for RAC-72. Now return data, even when no nodes are discovered instead of throwing an error.


srinia6:
test this please

amymullins:
test this please

PR #590, opened 2017-02-16
workflow level progress

lanchongyizu:
Implement workflow level progress indicator after task level progress.
https://rackhd.atlassian.net/browse/RAC-3923

@RackHD/corecommitters @pengz1 @iceiilin

### DIFF
--- a/lib/services/notification-api-service.js
+++ b/lib/services/notification-api-service.js
@@ -1,4 +1,4 @@
-// Copyright 2016-2017, Dell EMC, Inc.
+// Copyright © 2016-2017 Dell Inc. or its subsidiaries.  All Rights Reserved.
 
 'use strict';
 
@@ -16,7 +16,7 @@ di.annotate(NotificationApiServiceFactory,
         'Services.Waterline',
         'Errors',
         'Promise',
-        'Progress',
+        'GraphProgress',
         '_'
     )
 );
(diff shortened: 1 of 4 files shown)


lib/services/notification-api-service.js
```diff
@@ -80,50 +82,40 @@ function NotificationApiServiceFactory(
...
-                        taskName: graphObject.tasks[taskId].friendlyName,
-                        progress: progress
-                    }
-                };
-                return taskMessenger.publishProgressEvent(graphObject.instanceId, progressData);
+                var progress = new Progress(graph);
+                progress.calculateGraphProgress();
+                progress.setGraphDescription(taskProgress.description);
+                progress.setTaskProgress(taskId, taskProgress);
+                return eventsProtocol.publishProgressEvent(graph.instanceId, progress.getData());
```
pengz1: Suggest moving progress action to Progress instead of each events.

lib/services/notification-api-service.js
```diff
@@ -80,50 +82,40 @@ function NotificationApiServiceFactory(
...
-                        taskId: taskId,
-                        taskName: graphObject.tasks[taskId].friendlyName,
-                        progress: progress
-                    }
-                };
-                return taskMessenger.publishProgressEvent(graphObject.instanceId, progressData);
+                var progress = new Progress(graph);
+                progress.calculateGraphProgress();
+                progress.setGraphDescription(taskProgress.description);
+                progress.setTaskProgress(taskId, taskProgress);
```
lanchongyizu: I think you should comment on old code.

lib/services/notification-api-service.js
lanchongyizu: Same to the above.

pengz1:
+1

PR #591, opened 2017-02-17
Refactor OS installation progress

yyscamper:
1. Directly reference the milestone definition rather than the raw value.
2. Change all progress notification to be sent via URI /notification/progress
3. Explain the strange implemention for GET /notification/progress
4. Remove totalSteps/currentStep, always use the typic variable maximum/value

jenkins: depends on https://github.com/RackHD/on-tasks/pull/407

@pengz1 @lanchongyizu

### DIFF
--- a/data/profiles/install-centos.ipxe
+++ b/data/profiles/install-centos.ipxe
@@ -3,8 +3,8 @@ echo Starting CentOS/RHEL <%=version%> installer for ${hostidentifier}
 # The progress notification is just something nice-to-have, so progress notification failure should
 # never impact the normal installation process
 <% if( typeof progressMilestones !== 'undefined' && progressMilestones.enterProfileUri ) { %>
-# since there is no curl like http client in ipxe, so use imgfetch instead
-# note: the progress milestones uri must be wrapped in unescaped format, otherwise imgfetch will fail
+    # since there is no curl like http client in ipxe, so use imgfetch instead
+    # note: the progress milestones uri must be wrapped in unescaped format, otherwise imgfetch will fail
     imgfetch --name fakedimage http://<%=server%>:<%=port%><%-progressMilestones.enterProfileUri%> ||
     imgfree fakedimage ||
 <% } %>
--- a/data/profiles/install-centos.ipxe
+++ b/data/profiles/install-centos.ipxe
@@ -1,11 +1,23 @@
 echo Starting CentOS/RHEL <%=version%> installer for ${hostidentifier}
+
+# The progress notification is just something nice-to-have, so progress notification failure should
+# never impact the normal installation process
+<% if( typeof progressMilestones !== 'undefined' && progressMilestones.enterProfileUri ) { %>
+# since there is no curl like http client in ipxe, so use imgfetch instead
+# note: the progress milestones uri must be wrapped in unescaped format, otherwise imgfetch will fail
+    imgfetch --name fakedimage http://<%=server%>:<%=port%><%-progressMilestones.enterProfileUri%> ||
+    imgfree fakedimage ||
+<% } %>
+
 set base-url <%=repo%>/images/pxeboot
 set params initrd=initrd.img ks=<%=installScriptUri%> hostname=<%=hostname%> ksdevice=bootif BOOTIF=01-${netX/mac} console=<%=comport%>,115200n8 console=tty0
 kernel ${base-url}/vmlinuz repo=<%=repo%> ${params}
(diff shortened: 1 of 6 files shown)


changev:
test this please

lanchongyizu:
:+1:

data/profiles/install-centos.ipxe
```diff
@@ -1,11 +1,23 @@
 echo Starting CentOS/RHEL <%=version%> installer for ${hostidentifier}
+
+# The progress notification is just something nice-to-have, so progress notification failure should
+# never impact the normal installation process
+<% if( typeof progressMilestones !== 'undefined' && progressMilestones.enterProfileUri ) { %>
```
anhou: this code line is mixed in comments, the comments below could be indented

lib/services/common-api-presenter.js
```diff
@@ -208,9 +213,10 @@ function CommonApiPresenterFactory(
                     } catch (error) {
                         // If we failed to render the error then we've got larger problems.
 
-                        logger.error('Unable to render error template.', {
-                            macaddress: options.macaddress,
-                            error: error
+                        logger.error('Unable to render error profile.', {
```
anhou: the logs should be "Unable to render boilerplate.ipxe" ?

lib/api/2.0/notification.js
```diff
@@ -39,5 +47,8 @@ var notificationProgressGet = controller(function(req, res){
 
 module.exports = {
     notificationPost: notificationPost,
-    notificationProgressGet: notificationProgressGet
+    notificationProgressPost: notificationProgressPost,
+    //NOTE: in some environment, the HTTP client doesn't support POST, so we have to define
```
anhou: 'environments'  could you help provide the example 'ipxe' explicitly for understanding

static/monorail-2.0-sb.yaml
```diff
@@ -158,29 +158,73 @@ paths:
             $ref: '#/definitions/Error'
   /notification/progress:
     x-swagger-router-controller: notification
+# In some environment (such as ipxe), the avaiable http client doesn't support POST, so we have
+# to compromise to define below GET api to update progress information, but actually it does the
+# job of verb "POST"
     get:
       operationId: notificationProgressGet
```
anhou: how about  using 'notificationProgressPost' directly,  it could avoid confusion in two places( another is "notificationProgressGet: notificationProgressPost" ) and avoid you have to comments  in two places.

lib/api/2.0/notification.js
```diff
@@ -39,5 +47,8 @@ var notificationProgressGet = controller(function(req, res){
...
+    //a ugly API which use verb GET to update the progress information
```
pengz1: a =>an, use => uses

static/monorail-2.0-sb.yaml
```diff
@@ -158,29 +158,73 @@ paths:
...
-        send taskid, total setps and current step sequence to rackhd via notification get function
+        This API is used to update the progress information for a task instance
       parameters:
         - name: taskId
           in: query
           description: |
-            Task instance identifier
-          required: true 
+            The identifier of task instance which the progress applies to
+          required: false
```
pengz1: Without taskId, progress actually can't work, suppose it is required.

static/monorail-2.0-sb.yaml
```diff
@@ -158,29 +158,73 @@ paths:
...
+      summary: |
+        Send progress notification to task
+      description: |
+        This API is used to update the progress information for a task instance
+      parameters:
+        - name: taskId
+          in: query
+          description: |
+            The identifier of task instance which the progress applies to
+          required: false
```
pengz1: taskId should be required

static/monorail-2.0-sb.yaml
yyscamper: Good idea

static/monorail-2.0-sb.yaml
yyscamper: Actually if the progress data is embedded in http request body, then the taskId is optional in query.

lib/services/common-api-presenter.js
yyscamper: The error occurs when both the input profile & boilerplate.ipxe rendering fail, but the error input profile is really something we care while analyzing log, so I don't think your log message is better than the original one.

yyscamper:
jenkins: test this please

PR #598, opened 2017-02-27
Refine workflow progress

lanchongyizu:
Refactor workflow progress implementation with progress service design.
1) publish graph.started event for AMQP
2) move postProgressNotification into Services.GraphProgress

This PR is related with https://rackhd.atlassian.net/browse/RAC-4390

Depends on https://github.com/RackHD/on-core/pull/261 https://github.com/RackHD/on-taskgraph/pull/220 https://github.com/RackHD/on-tasks/pull/418

@RackHD/corecommitters @pengz1

### DIFF
--- a/lib/services/notification-api-service.js
+++ b/lib/services/notification-api-service.js
@@ -56,7 +56,7 @@ function NotificationApiServiceFactory(
             throw new Errors.BadRequestError('maximum is required for progress notification');
         }
         if(!message.value) {
-            throw new Errors.BadRequestError('maximum is required for progress notification');
+            throw new Errors.BadRequestError('value is required for progress notification');
         }
 
         if (message.value) {
@@ -77,7 +77,7 @@ function NotificationApiServiceFactory(
                 task.graphId,
                 message.taskId,
                 progressData,
-                false
+                {swallowError: false}
             );
         });
     };
--- a/lib/services/notification-api-service.js
+++ b/lib/services/notification-api-service.js
@@ -41,19 +41,44 @@ function NotificationApiServiceFactory(
 
         if (_.has(message, 'nodeId')) {
             return self.postNodeNotification(message);
-        } else if (_.has(message, 'taskId')) {
-            if (_.has(message, 'progress')) {
-                // This will be progress update notification if taskId is specified
-                return graphProgressService.postProgressNotification(
-                    message.taskId,
-                    message.progress
-                );
-            }
         }
         else {
             // This will be a broadcast notification if no id (like nodeId) is specified
             return self.postBroadcastNotification(message);
         }
+    };
+
+    NotificationApiService.prototype.publishTaskProgress = function(message) {
+        if(!_.isString(message.taskId)) {
+            throw new Errors.BadRequestError('taskId is required for progress notification');
+        }
+        if(!message.maximum) {
+            throw new Errors.BadRequestError('maximum is required for progress notification');
+        }
+        if(!message.value) {
+            throw new Errors.BadRequestError('maximum is required for progress notification');
+        }
+
+        if (message.value) {
+            message.value = parseInt(message.value);
+        }
+        if (message.maximum) {
+            message.maximum = parseInt(message.maximum);
+        }
+        var progressData = _.pick(message, ['maximum', 'value', 'description']);
+
+        return waterline.taskdependencies.findOne({taskId: message.taskId})
+        .then(function(task) {
+            if (_.get(task, 'graphId') === undefined) {
+                throw new Errors.BadRequestError('Cannot find the active task for taskId ' + message.taskId); //jshint ignore: line
+            }
+
+            return graphProgressService.publishTaskProgressNotCatch(
+                task.graphId,
+                message.taskId,
+                progressData
+            );
+        });
     };
 
     NotificationApiService.prototype.postNodeNotification = function(message) {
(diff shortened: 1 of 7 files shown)


lib/services/notification-api-service.js
```diff
@@ -43,18 +41,46 @@ function NotificationApiServiceFactory(
...
+        }
+        if (message.maximum) {
+            message.maximum = parseInt(message.maximum);
+        }
+        var progressData = _.pick(message, ['maximum', 'value', 'description']);
+
+        return waterline.taskdependencies.findOne({taskId: message.taskId})
+        .then(function(task) {
+            if (_.get(task, 'graphId') === undefined) {
+                throw new Errors.BadRequestError('Cannot find the active task for taskId ' + message.taskId); //jshint ignore: line
```
yyscamper: The error message doesn't match with the real function, `waterline.taskdependencies.findOne` is only trying to find a task, this doesn't mean the returned task is the active task. The definition of active task is the task that is running (RackHD is executing its job code).

lib/services/workflow-api-service.js
```diff
@@ -96,11 +96,8 @@ function workflowApiServiceFactory(
                         definition, configuration.options, context, configuration.domain, true);
             });
         })
-        .tap(function(graph) {
-            var progress = GraphProgress.create(graph, 'Graph "' + graph.name + '" started');
-            return eventsProtocol.publishProgressEvent(graph.instanceId,
-                                                       progress.getProgressEventData());
-        })
+        .tap(taskMessenger.publishGraphStarted)
+        .tap(graphProgressService.publishGraphStarted)
```
yyscamper: Why not move these two kinds of events into on-taskgraph, on-taskgraph should own how/where to generate the task graph related event.

spec/lib/api/2.0/notification-spec.js
```diff
@@ -88,187 +88,160 @@ describe('Http.Api.Notification', function () {
     });
 
     describe('POST /notification/progress', function () {
-        beforeEach(function() {
-            sinon.stub(notificationApiService, 'postNotification').resolves();
-        });
+        describe('should return 200', function() {
```
yyscamper: It doesn't make sense to create such kind of test suites. Any reason to do this?
And if you are surely want to create a test suite here, the test suite name need be changed to a better one. The text `should xxx` is usually used only for test case.

lib/services/workflow-api-service.js
yyscamper: I see graphFinished related events are generated inside on-taskgraph, it's odd that graphStarted is not within on-taskgraph but instead inside on-http.

lib/services/notification-api-service.js
```diff
@@ -43,18 +41,46 @@ function NotificationApiServiceFactory(
 
         if (_.has(message, 'nodeId')) {
             return self.postNodeNotification(message);
-        } else if (_.has(message, 'taskId')) {
-            if (_.has(message, 'progress')) {
-                // This will be progress update notification if taskId is specified
-                return self.postProgressNotification(message.taskId, message.progress);
-            }
```
pengz1: If you remove these lines, you need update Secure Erase job which still depends on this API.

lib/services/notification-api-service.js
```diff
@@ -43,18 +41,46 @@ function NotificationApiServiceFactory(
...
+        .then(function(task) {
+            if (_.get(task, 'graphId') === undefined) {
+                throw new Errors.BadRequestError('Cannot find the active task for taskId ' + message.taskId); //jshint ignore: line
+            }
+
+            return graphProgressService.publishTaskProgressNotCatch(
+                task.graphId,
+                message.taskId,
+                progressData
+            );
```
pengz1: Suggest you to log those errors in logs instead of letting them unhandled. After all API errors are not as well organized as our logs.

lib/services/notification-api-service.js
```diff
@@ -43,18 +41,46 @@ function NotificationApiServiceFactory(
...
+            message.value = parseInt(message.value);
+        }
+        if (message.maximum) {
+            message.maximum = parseInt(message.maximum);
+        }
+        var progressData = _.pick(message, ['maximum', 'value', 'description']);
+
+        return waterline.taskdependencies.findOne({taskId: message.taskId})
+        .then(function(task) {
+            if (_.get(task, 'graphId') === undefined) {
```
pengz1: Not sure if there is case grahpId is null, it looks to me _.isEmpty is more powerful here.

lib/services/workflow-api-service.js
lanchongyizu: Good catch!
I just refactored the code here but didn't check whether it's the best place to publish the events. We can raise another PR to do that.

lib/services/notification-api-service.js
lanchongyizu: Thanks for the reminder.

lib/services/notification-api-service.js
lanchongyizu: Just move the code here. But after discussion, I will remove the `active.`

pengz1:
+1

lib/services/notification-api-service.js
```diff
@@ -43,18 +41,47 @@ function NotificationApiServiceFactory(
...
 
+    NotificationApiService.prototype.publishTaskProgress = function(message) {
+        if(!_.isString(message.taskId)) {
+            throw new Errors.BadRequestError('taskId is required for progress notification');
+        }
+        if(!message.maximum) {
+            throw new Errors.BadRequestError('maximum is required for progress notification');
+        }
+        if(!message.value) {
+            throw new Errors.BadRequestError('maximum is required for progress notification');
```
anhou: copy paste line, I think your intended error is not 'maxisum'

yyscamper:
travis ci error is expected as there are dependency PRs.

PR #600, opened 2017-03-01
Fix bug  introduced by hook-schema that breaks swagger-ui

srinia6:
The $ref under HooksPost was causing the swagger-ui crash.

### DIFF
--- a/static/monorail-2.0.yaml
+++ b/static/monorail-2.0.yaml
@@ -80,7 +80,15 @@ definitions:
     description: Hook post schema 
     type: object
     properties:
-      $ref: "#/definitions/hooks.2.0_HookBase/properties"
+      url:
+        type: string
+        format: uri
+      name:
+        type: string
+      filters:
+        type: array
+        items:
+          type: object
     required:
       - url
   ipmi-obm-service_Obm:


panpan0000:
:+0
thanks to fixing it

pengz1:
+1

yyscamper:
I agree this is a bug, but I don't agree your judgement about the root cause, as I don't see any circular reference that introduced by the `$ref`. 

My understanding is that this is an invalid usage of JSON schema $ref, the $ref needs be pointed to a schema, but obviously `#/definitions/hooks.2.0_HookBase/properties` is not a valid schema.

I have tested below fix, it works well:
```yaml
  hooks.2.0_HookPost:
    description: Hook post schema
    $ref: "#/definitions/hooks.2.0_HookBase"
    required:
      - url
```

If you agree my explanation, please fix the commit message and PR title.  Though your fix is good as well, But I think my fix will be more simple and avoid potentially divergence. (Of course, you fix is also acceptable to me)


Meanwhile, I am thinking whether we can add a test case to make sure the swagger-ui is not broken. Because at the time when the hooks feature was merged, actually the jenkins light was green. If we can catch this failure earlier, then I believe this bug can be avoided.

@brianparry : do you know is there any swagger functionality we can leverage for the testing?

panpan0000:
Aha, @yyscamper , we may introduce web-frond-end testing technology to auto validate on-web-ui and swagger-ui , Woo! we become a "dot-com company" :-)

anhou:
@srinia6 Please keep 100644 file mode, don't change it if no other concerns

yyscamper:
@panpan0000 : maybe swagger has some API call that can be used to check its internal error, so we don't need web-front-end testing. @brianparry is the expert, so hopefully he will find a way to easily test such kind of function.

srinia6:
@yyscamper: I did try that ["#/definitions/hooks.2.0_HookBase"], though the unit-test and the API does pass with that , **but still the validation does not have any affect,** and goes inside the service code. Which i think is an issue. Not sure if that is a swagger issue or the validation issue.

srinia6:
@anhou Thank you, changed it.

yyscamper:
@srinia6 : I just checked the JSON schema spec, "All other properties in a "$ref" object MUST be ignored." (see http://json-schema.org/latest/json-schema-core.html#rfc.section.7 ) , This is the reason why `required` doesn't work.

If so, you could use `allOf`, the example is as below:
```json
allOf: [
    { $ref: "#/definitions/hooks.2.0_HookBase" }, 
    { required: "url" }
]
```

Anyway, this is nice-to-have.

yyscamper:
@srinia6 : Please explain the judgement on `circular reference`, if this is not "circular reference", then please change the commit message.

srinia6:
@yyscamper BTW, tried that allOf, and i still see validation error. I see similar examples under schema/.json's but not any under .yaml.

PR #602, opened 2017-03-03
Validate url for webhook

pengz1:
Schema URL validation doesn't work, add url validation code in hook-api

### DIFF
--- a/lib/services/hooks-api-service.js
+++ b/lib/services/hooks-api-service.js
@@ -11,14 +11,18 @@ di.annotate(hookApiServiceFactory,
         'Services.Waterline',
         'Logger',
         '_',
-        'Errors'
+        'Errors',
+        'Assert',
+        'Promise'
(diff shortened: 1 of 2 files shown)


lib/services/hooks-api-service.js
```diff
@@ -45,7 +49,12 @@ function hookApiServiceFactory(
      * @return
      */
     HooksApiService.prototype.createHook = function(body) {
-        return waterline.hooks.findOne({url: body.url})
+        return Promise.try(function(){
```
anhou: Just confirm that, no 'catch' is intended?

pengz1:
@iceiilin @nortonluo

lib/services/hooks-api-service.js
yyscamper: The error should be escalated to upper level, so no catch here.

PR #603, opened 2017-03-03
modify photon ks file to put swap partition as the last item

iceiilin:
fix https://rackhd.atlassian.net/browse/RAC-4324, which complains inproper UI display after installing Photon OS using full payload in https://github.com/RackHD/RackHD/blob/master/example/samples/install_photon_os_payload_full.json

From KS file sample provided by Photon, the swap partition is the last item in the "partition" field. This PR modify the rendering rule in ks template to align with it and pass the test.

### DIFF
--- a/data/templates/install-photon/photon-os-ks
+++ b/data/templates/install-photon/photon-os-ks
@@ -29,14 +29,19 @@
                         <%_ numSize = 0; _%>
                     <%_ } _%>
                     <%_ if (par.fsType === "swap") { _%>
-                        <%_ if (i === length ) { _%>
-        {"size": <%=numSize%>, "filesystem": "<%=par.fsType%>"}
-                        <%_ } else { _%>
-        {"size": <%=numSize%>, "filesystem": "<%=par.fsType%>"},
+                        <%# Through experiments, swap partition should always be the last one in %>
+                        <%# "partition" field of ks file, otherwise GUI will display inproperly %>
+                        <%_ swapPar = par; _%>
+                        <%_ swapParSize = numSize; _%>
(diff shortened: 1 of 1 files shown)


data/templates/install-photon/photon-os-ks
```diff
@@ -29,14 +29,14 @@
...
-        {"size": <%=numSize%>, "filesystem": "<%=par.fsType%>"},
-                        <%_ } _%>
+                        <%_ swapPar = par; _%>
+                        <%_ swapParSize = numSize; _%>
                     <%_ } else { _%>
-                        <%_ if (i === length ) { _%>
+                        <%_ if (i === length && typeof swapPar === 'undefined') { _%>
         {"mountpoint": "<%=par.mountPoint%>", "size": <%=numSize%>, "filesystem": "<%=par.fsType%>"}
+                        <%_ } else if (i === length && typeof swapPar !== 'undefined') { _%>
+        {"mountpoint": "<%=par.mountPoint%>", "size": <%=numSize%>, "filesystem": "<%=par.fsType%>"},
```
anhou: could these break into multiple lines to be more readable

iceiilin: it's just a rendered output in ks file for os installer, not the logic itself when rendering. i think it more readable when putting it in one line, so the reader would know that in a certain condition, something will be rendered, and they just need to skip one line to check other conditions.

yyscamper: Please add some comment so that other developer will be much more easier to understand such kind of special design. 
Ejs has a special tag to add comment.

iceiilin:
comments added in code

PR #604, opened 2017-03-07
Add "associatedServer" property to views for Cisco UCS ServiceProfile support

uppalk1:


uppalk1:
@geoff-reid @tannoa2 @keedya

data/views/node.2.0.json
```diff
@@ -1,5 +1,6 @@
 {
     "autoDiscover": "<%=autoDiscover%>",
+    "associatedServer":"<%=associatedServer%>",
```
yyscamper: From the property name, I think it can be moved into a new kind of relation, rather than create a new property in node document.

anhou: +1 for having a try to include it to the 'relationship' field.

geoff-reid:
test this please

PR #607, opened 2017-03-11
disable 1.1 apis

johren:
Also removed the 1.1 unit tests.  Let me know if there is a better way to disable these.


amymullins:
@brianparry

king-jam:
:+1: delete all the things!

PR #610, opened 2017-03-22
Redfish UpdateService implementation for Dell

dalebremner:
Perform firmware update on Dell server using Racadm through Redfish API.

### DIFF
--- a/lib/api/redfish-1.0/update-service.js
+++ b/lib/api/redfish-1.0/update-service.js
@@ -15,80 +15,57 @@ var updateFirmware = controller({success: 201}, function (req, res) {
     var payload = req.swagger.params.payload.value;
     var targetNodeIds = payload.Targets;
     var errorList = [];
-    var transactionError = false;
     var options = redfish.makeOptions(req, res, payload);
 
-    return Promise.try(function () {
-        // Run update firmware graph for each specified node id
-        // If there are any errors, the entire request fails
-        return Promise.map(
-            targetNodeIds,
-            function (targetNodeId) {
-                return waterline.nodes.getNodeById(targetNodeId)
-                    .then(function (targetNode) {
-                        if (!targetNode) {
-                            transactionError = true;
-                            throw new Errors.NotFoundError(
-                                'Node not Found ' + targetNodeId);
-                        }
-                    })
-                    .then(function () {
-                        return waterline.obms.findAllByNode(targetNodeId, true);
-                    })
-                    .then(function (obmSettings) {
-                        if (!obmSettings) {
-                            transactionError = true;
-                            throw new Errors.NotFoundError(
-                                'OBM settings not defined for node: ' + targetNodeId);
-                        }
-
-                        // Find IPMI settings for the node
-                        var obm = _.find(obmSettings, {'service': 'ipmi-obm-service'});
-                        if (!obm) {
-                            transactionError = true;
-                            throw new Errors.NotFoundError(
-                                'IPMI settings not defined for node: ' + targetNodeId);
-                        }
-
-                        if (!transactionError) {
-                            return workflow.createAndRunGraph(
-                                {
-                                    name: 'Graph.Dell.Racadm.Update.Firmware',
-                                    options: {
-                                        defaults: {
-                                            serverUsername: obm.config.user,
-                                            serverPassword: obm.config.password,
-                                            serverFilePath: payload.ImageURI
-                                        }
-                                    }
-                                },
-                                targetNodeId
-                            );
-                        }
-                    })
-                    .then(function (graph) {
-                        if (graph) {
-                            var output = {
-                                "@odata.id": options.basepath +
-                                '/TaskService/Tasks/' +
-                                graph.instanceId
-                            };
-                            res.status(201).json(output);
-                        }
-                    })
-                    .catch(function (error) {
-                        errorList.push(error);
-                    });
+    // Run update firmware graph for each specified node id
+    // If there are any errors, the entire request fails
+    return Promise.map(
+        targetNodeIds,
+        function (targetNodeId) {
+            return waterline.nodes.getNodeById(targetNodeId)
+                .then(function (targetNode) {
+                    if (!targetNode) {
+                        throw new Errors.NotFoundError(
+                            'Node not Found ' + targetNodeId);
+                    }
+                })
+                .then(function () {
+                    return workflow.createAndRunGraph(
+                        {
+                            name: 'Graph.Dell.Racadm.Update.Firmware',
+                            options: {
+                                defaults: {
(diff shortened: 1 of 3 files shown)


lib/api/redfish-1.0/update-service.js
```diff
@@ -0,0 +1,96 @@
...
+var waterline = injector.get('Services.Waterline');
+
+var updateFirmware = controller({success: 201}, function (req, res) {
+    var payload = req.swagger.params.payload.value;
+    var targetNodeIds = payload.Targets;
+    var errorList = [];
+    var transactionError = false;
+    var options = redfish.makeOptions(req, res, payload);
+
+    return Promise.try(function () {
```
yyscamper: Since there is only `Promise.map` within the `Promise.try`, so Promise.try is not necessary.

lib/api/redfish-1.0/update-service.js
```diff
@@ -0,0 +1,96 @@
...
+                    })
+                    .then(function (obmSettings) {
+                        if (!obmSettings) {
+                            transactionError = true;
+                            throw new Errors.NotFoundError(
+                                'OBM settings not defined for node: ' + targetNodeId);
+                        }
+
+                        // Find IPMI settings for the node
+                        var obm = _.find(obmSettings, {'service': 'ipmi-obm-service'});
```
yyscamper: Using `waterline.obms.findByNode(targetNodeId, 'ipmi-obm-service', true) is more straightforward here, so the promise at line 36 can be removed.

lib/api/redfish-1.0/update-service.js
```diff
@@ -0,0 +1,96 @@
...
+                        if (!obm) {
+                            transactionError = true;
+                            throw new Errors.NotFoundError(
+                                'IPMI settings not defined for node: ' + targetNodeId);
+                        }
+
+                        if (!transactionError) {
```
yyscamper: If there is any transaction error happened before, the promise has been rejected and thus code will never run to this line. So the variable `transactionError` is unnecessary.

lib/api/redfish-1.0/update-service.js
```diff
@@ -0,0 +1,96 @@
...
+                            return workflow.createAndRunGraph(
+                                {
+                                    name: 'Graph.Dell.Racadm.Update.Firmware',
+                                    options: {
+                                        defaults: {
+                                            serverUsername: obm.config.user,
+                                            serverPassword: obm.config.password,
```
yyscamper: serverUsername and serverPassword is for the shared samba server, not ipmi user/password!

You could touch @pengz1 for the usage of Racadm firmware update workflow usage

spec/lib/api/redfish-1.0/update-service-spec.js
```diff
@@ -0,0 +1,131 @@
...
+            .expect('Content-Type', /^application\/json/)
+            .expect(400)
+            .expect(function (response) {
+                expect(response.body).to.have.property('error');
+                expect(redfish.handleError).to.have.been.calledOnce;
+            });
+    });
+
+    it('should return an error if OBM settings are not set', function () {
+        waterline.obms.findAllByNode.rejects();
```
yyscamper: The test condition is `if obm settings are not set`, in this case, `waterline.obms.findAllByNode` should be stubbed to resolves([]) rather than being rejected.

spec/lib/api/redfish-1.0/update-service-spec.js
```diff
@@ -0,0 +1,131 @@
...
+            .send(badPayload)
+            .expect('Content-Type', /^application\/json/)
+            .expect(400)
+            .expect(function (response) {
+                expect(response.body).to.have.property('errors');
+            });
+    });
+
+    it('should return an error if a node is not found', function () {
+        waterline.nodes.getNodeById.rejects();
```
yyscamper: The test condition is `if node is not found`, in this case, `waterline.nodes.getNodeById` should be stubbed to resolves(undefined) rather than being rejected.

lib/api/redfish-1.0/update-service.js
```diff
@@ -0,0 +1,96 @@
...
+                        if (!obm) {
+                            transactionError = true;
+                            throw new Errors.NotFoundError(
+                                'IPMI settings not defined for node: ' + targetNodeId);
+                        }
+
+                        if (!transactionError) {
+                            return workflow.createAndRunGraph(
+                                {
+                                    name: 'Graph.Dell.Racadm.Update.Firmware',
```
yyscamper: Since the API entry `/redfish/v1/UpdateService/Actions/UpdateService.SimpleUpdate` is common for all kinds of servers, so this controller function should take all servers into consideration rather than only Dell.  But for your design, it is actually only valid for Dell server.

We certainly have defined different firmware update graphs for different servers, but they are not unlike Dell (one graph for all firmwares), those server defines different graph for different firmware, so there is certainly some challenge to support all servers.

This comment is not push you to support all servers, but as a reminder for you. In the point of feature integrity, it deserves some following action items.

lib/api/redfish-1.0/update-service.js
dalebremner: @yyscamper thanks, good catch... Looking closer at the code, the racadm job appears to get the OBM settings, so I can eliminate that from my service.

lib/api/redfish-1.0/update-service.js
dalebremner: @yyscamper yes, understood... The intent is to handle more servers in the future.

lib/api/redfish-1.0/update-service.js
dalebremner: I was able to test on a Dell server. It works because the job ignores my graph options.

lib/api/redfish-1.0/update-service.js
dalebremner: Done

lib/api/redfish-1.0/update-service.js
dalebremner: done

lib/api/redfish-1.0/update-service.js
dalebremner: done

spec/lib/api/redfish-1.0/update-service-spec.js
dalebremner: done

dalebremner:
@yyscamper I have addressed all of your comments. Please review. Thanks!

PengTian0:
test this please

dalebremner:
jenkins: test this please

dalebremner:
jenkins: test this please

dalebremner:
jenkins: test this please

PR #615, opened 2017-03-24
Allow backdoor process, to create Administrator user

srinia6:
Extend the Localhost Exception, to allow backdoor access to create an Administrator account, when there are no "Adminstrator" in the collection.
@RackHD/corecommitters

### DIFF
--- a/lib/api/2.0/users.js
+++ b/lib/api/2.0/users.js
@@ -23,9 +23,10 @@ var addUser = controller({success: 201}, function(req, res) {
         return accountService.listUsers();
     })
     .then(function(users) {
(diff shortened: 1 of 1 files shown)


lib/api/2.0/users.js
```diff
@@ -23,9 +23,10 @@ var addUser = controller({success: 201}, function(req, res) {
         return accountService.listUsers();
     })
     .then(function(users) {
-        if(!users.length && localUserException && res.locals.ipAddress === '127.0.0.1') {
+        if((!users.length && localUserException && res.locals.ipAddress === '127.0.0.1') || _.isEmpty(_.filter(users, {role: 'Administrator'}))) {
```
srinia6: Ignoring this.

PR #616, opened 2017-03-25
Modify swagger to return task id from Redfish BootImage

dalebremner:


yyscamper:
@dalebremner : How about to create a unit-test for verify this feature?

dalebremner:
@yyscamper The unit test already exists in:
https://github.com/RackHD/on-http/blob/master/spec/lib/api/redfish-1.0/system-spec.js
... since the rest API returns the task. This PR corrects client libraries generated from the swagger definition.

PR #618, opened 2017-03-28
RAC-4625: adding basic username complexity checking on account creation

diontje:
Adding bare minimum account username complexity checking on account creation for both APIs.
As it stands now, accounts can be created with a single space, for example.
Change necessitates that usernames begin with a letter, followed by any number of letters, numbers, hyphens, periods, and underscores.


diontje:
test this please

static/monorail-2.0.yaml
```diff
@@ -308,6 +308,7 @@ definitions:
         type: string
       username:
         type: string
+      pattern: ^([A-Za-z][0-9A-Za-z._-]+)$
```
yyscamper: The `pattern` has to be put under `username`!

yyscamper:
Please create some unit-test cases to verify the pattern surely works as expected. For API 2.0, the unit-test is in https://github.com/RackHD/on-http/blob/master/spec/lib/api/2.0/users-spec.js

yyscamper:
Meanwhile, How do you think to take this chance to add some restriction for the `password` as well?  This is not a must for your PR, but I encourage you to do it.

static/monorail-2.0.yaml
diontje: interesting. i tested the change on box and did not receive any complaints from on-http.
i will make the change and resubmit. thanx!

static/redfish.yaml
```diff
@@ -3731,6 +3731,7 @@ definitions:
       UserName:
         description: This property contains the user name for the account.
         type: string
+        pattern: '^[A-Za-z][A-Za-z0-9._-]+$'
```
yyscamper: This pattern isn't the same with the monorail-2.0.yaml. I don't find a reason to keep them different, please align them.

spec/lib/api/2.0/users-spec.js
```diff
@@ -298,4 +298,53 @@ describe('Http.Api.Users', function () {
...
+        accountService.listUsers.resolves([ userObj ]);
+        return helper.request().post('/login')
+            .send({username: " readonly", password: "read123"})
+            .expect(201)
+            .then(function(res) {
+                return res.body.token;
+            }).then(function(token) {
+                return helper.request().post('/api/2.0/users')
+                    .set("authorization", 'JWT ' + token)
+                    .send(readOnlyObj)
```
yyscamper: the bad user name is set in line 305, so the 400 expectation should be returned earlier in line 306. Actually the `readOnlyObj` has the valid username, so it's returned http status should be success.

From your test purpose, the bad username need be set in line 312, the username (line 305) for `/login` should be valid.

brianparry:
@diontje I think we actually want to put these checks here: https://github.com/RackHD/on-http/blob/master/static/schemas/2.0/user.2.0.json and here: https://github.com/RackHD/on-http/blob/master/static/schemas/2.0/user.2.0.json#L9.  @dalebremner What are your thoughts on adding a complexity regex to the Redfish `Name` resource schema.

dalebremner:
@brianparry I would be hesitant to add the complexity regex to the schema, as I don't think it's specified by the Redfish standard.

diontje:
test this please

PR #619, opened 2017-03-29
Change /schemas/tasks API implementation according to new task schema design

yyscamper:
Changelog:

- The task schema identifier is changed to use task's injectableName rather than filename, since the built-in schema doesn't have a filename

- GET /schemas/tasks will return the list of task injectableNames

- The `resolved` flag is removed in GET /schemas/tasks/:identifier since now the returned schema is always resolved. The unresolved version doesn't conform to the JSON schema standard, so it is removed.

@anhou @iceiilin @pengz1 @lanchongyizu @brianparry

### DIFF
--- a/lib/api/2.0/schemas2.js
+++ b/lib/api/2.0/schemas2.js
@@ -7,7 +7,9 @@ var controller = injector.get('Http.Services.Swagger').controller;
 var schemaApiService = injector.get('Http.Api.Services.Schema');
 var nameSpace = '/api/2.0/schemas/';
 var Errors = injector.get('Errors');
-var taskOptionValidator = injector.get('TaskOption.Validator');
+var workflowApiService = injector.get('Http.Services.Api.Workflows');
+var Task = injector.get('Task.Task');
+var _ = injector.get('_');
 
 // GET /api/2.0/schemas
 var schemasGet = controller(function() {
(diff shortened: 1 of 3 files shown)


yyscamper:
jenkins: test this please

lib/api/2.0/schemas2.js
```diff
@@ -26,26 +28,25 @@ var schemasIdGet = controller(function(req) {
...
-    return taskOptionValidator.getAllSchemaNames({ includeNameSpace: false });
+    return workflowApiService.getTaskDefinitions()
+    .then(function(tasks) {
+        return tasks.map(function(task) {
+            return task.injectableName;
+        });
+    });
 });
 
 // GET /api/2.0/schemas/tasks/:identifier
```
pengz1: :name should be more accurate than :identifier.

lib/api/2.0/schemas2.js
```diff
@@ -26,26 +28,25 @@ var schemasIdGet = controller(function(req) {
 
 // GET /api/2.0/schemas/tasks
 var taskSchemasGet = controller(function () {
-    return taskOptionValidator.getAllSchemaNames({ includeNameSpace: false });
+    return workflowApiService.getTaskDefinitions()
+    .then(function(tasks) {
```
brianparry: @yyscamper FYI - Bluebird promises have a chaiable `map` method: http://bluebirdjs.com/docs/api/promise.map.html

yyscamper: The promise that returned from `getTaskDefinition` is not bluebird style promise, it is waterline's promise, it doesn't have  a `map` function, so that's the reason I ended up using the array's `map`.

Now in RackHD code there is various Promise (at least there are Bluebird promise, node.js native promise, waterline promise), we'd better to combine them to a single one in future.

brianparry: @yyscamper I like to wrap promise returning function calls in `Promise.try` for this reason.  This normalizes promises to Bluebird style.  It also will handle any synchronous errors that might be thrown by a promise returning function.

yyscamper: I like `Promise.try` as well, it changes a lot of things (synchronous code or erros, non-bluebird promises) into bluebird promise, though a little cumbersome sometimes.

PengTian0:
test this please

yyscamper:
jenkins: test this please

PR #621, opened 2017-03-29
Add a post route to receive redfish alerts

tannoa2:
@RackHD/corecommitters @RackHD/veyron @yyscamper

### DIFF
--- a/lib/api/2.0/notification.js
+++ b/lib/api/2.0/notification.js
@@ -6,6 +6,8 @@ var injector = require('../../../index.js').injector;
 var controller = injector.get('Http.Services.Swagger').controller;
 var notificationApiService = injector.get('Http.Services.Api.Notification');
 var _ = injector.get('_');    // jshint ignore:line
+var parser = require('body-parser');
+var Promise = injector.get('Promise');
 
 var notificationPost = controller({success: 201}, function(req) {
     var message = _.defaults(req.swagger.query || {}, req.query || {}, req.body || {});
@@ -30,7 +32,40 @@ var notificationProgressPost = controller(function(req, res){
     });
 });
 
+/**
+ * @api {post} /api/2.0/notification/alerts
+ * @apiDescription Receives Redfish alerts, refactors the message and publishes it into amqp
+ * @apiName notification post
+ * @apiGroup notification
+ */
+var twistedParser = parser.json({ type: 'application/x-www-form-urlencoded' });
+var notificationAlertsPost = controller({success: 201}, function(req,res) {
+    var format= req.headers["content-type"];
+    return Promise.try(function(){
+        return new Promise(function(resolve, reject) {
+            if (format === 'application/x-www-form-urlencoded'){
+                //This is a work around where some vendors are sending json data
+                // with http header content-type of “x-www-form-urlencoded”
+                twistedParser(req, res, function (err) {
+                    if (err) {
+                        return reject(err);
+                    }else{
+                        resolve();
+                    }
+                });
+            }else{
+                resolve();
+            }
+
+        })
+        .then(function() {
+            return notificationApiService.redfishAlertProcessing(req);
+        });
+    });
+});
+
 module.exports = {
     notificationPost: notificationPost,
-    notificationProgressPost: notificationProgressPost
+    notificationProgressPost: notificationProgressPost,
+    notificationAlertsPost:notificationAlertsPost
 };
--- a/lib/services/notification-api-service.js
+++ b/lib/services/notification-api-service.js
@@ -16,7 +16,9 @@ di.annotate(NotificationApiServiceFactory,
         'Errors',
         'Promise',
         'Services.GraphProgress',
-        '_'
+        '_',
+        'Services.Lookup',
+        'Constants'
     )
 );
 
@@ -29,7 +31,9 @@ function NotificationApiServiceFactory(
     Errors,
     Promise,
     graphProgressService,
-    _
+    _,
+    lookup,
+    Constants
 ) {
     var logger = Logger.initialize(NotificationApiServiceFactory);
 
@@ -109,6 +113,170 @@ function NotificationApiServiceFactory(
             return message;
         });
     };
+    NotificationApiService.prototype.redfishAlertProcessing = function(req) {
+        /*
+        The alert would look like the json below when published on the AMQP bus.
+        It is compliant with RackHD notification standard.
+        Notification Type is "node" and action is "alerts"
+
+         {
+         "type": "node",
+         "action": "alerts",
+         "typeId": "58d94cec316779d4126be134",
+         "data": {
+             "Context": "context string",
+             "EventId": "8689",
+             "EventTimestamp": "2017-03-29T15:39:34-0500",
+             "EventType": "Alert",
+             "MemberId": "7e675c8e-127a-11e7-9fc8-64006ac35232",
+             "Message": "The coin cell battery in CMC 1 is not working.",
+             "MessageArgs": ["1"],
+             "MessageArgs@odata.count": 1,
+             "MessageId": "CMC8572",
+             "Severity": "Critical",
+             "sourceIpAddress": "10.240.19.130",
+             "nodeId": "58d94cec316779d4126be134",
+             "ChassisName": "PowerEdge R630",
+             "ServiceTag": "4666482",
+             "SN": "CN747515A80855"
+         },
+         "severity": "critical",
+         "version": "1.0",
+         "createdAt": "2017-03-29T19:44:12.972Z"
+         }
+         */
+        var amqpMessage = {};
+
+        return new Promise.resolve()
+            .then(function(){
+                return getAlert(req);
+            })
+            .then(function(data){
+                amqpMessage = data;
+                return getObm(amqpMessage.data.sourceIpAddress);
+            })
+            .then(function(obm){
+                return  updateBmcInfo(amqpMessage, obm);
+            })
+            .then(function(data){
+                return updateDeviceInfo(data);
+            })
+            .tap(function(data){
+                return eventsProtocol.publishExternalEvent(data);
+            });
+    };
+
+    function getAlert (req) {
+        var amqpMessage = {};
+        var ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress ||
+            req.socket.remoteAddress || req.connection.socket.remoteAddress;
+        logger.info("Alert was received from the node with IP: " + ip);
+        var alert = _.defaults(req.swagger.query || {},req.query || {}, req.body || {});
+        if(typeof(alert) === 'object' &&  Object.keys(alert).length > 0) {
+            //Format the amqp message by adopting the rackHD notification standard
+            amqpMessage.type = "node";
+            amqpMessage.action = "alerts";
+            amqpMessage.data = alert;
+            amqpMessage.data.sourceIpAddress = ip;
+            if (alert.hasOwnProperty("Severity")) {
+                amqpMessage.severity = alert.Severity.toLowerCase();
+            }
+
+            return amqpMessage;
+        }
+        else{
+            //Unexpected decoded "x-www-forum-urlencoded" data
+            throw new Errors.BadRequestError(
+                'Unexpected alert data'
+            );
+        }
+    }
+
+    function getObm(ip){
+        return waterline.obms.findOne({"config.host": ip, "service" : "ipmi-obm-service" })
+            .then(function(obm) {
+                if (obm !== undefined) {
+                    return obm;
+                }else{
+                    return lookup.lookupByIP(ip)
+                        .then(function(lookup){
+                            if(lookup !== undefined){
+                                var macAdd = lookup.macAddress;
+                                return waterline.obms.findOne({"config.host": macAdd, "service" : "ipmi-obm-service" })// jshint ignore:line
+                                    .then(function(obm){
+                                        return obm;
+                                    });
+                            }
+                            return lookup;
+                        });
+                }
+            });
+    }
+
+    function updateBmcInfo(amqpMessage, obm) {
+        var ip = amqpMessage.data.sourceIpAddress;
+
+            if (obm !== undefined && Constants.Regex.IpAddress.test(obm.config.host)) {
+                //obm.config.host is an IP
+                return waterline.catalogs.findMostRecent({"node": obm.node, source: "bmc"})// jshint ignore:line
+                    .then(function (bmcCatalog) {
+                        if (bmcCatalog !== undefined && bmcCatalog.hasOwnProperty("node") && bmcCatalog.node !== "") { // jshint ignore:line
+                            return extractBmcInfo(amqpMessage, bmcCatalog);
+                        }
+                    });
+            }else if (obm !== undefined && Constants.Regex.MacAddress.test(obm.config.host)) {
+                //obm.config.host is a mac address
+                amqpMessage.typeId = obm.node;
+                amqpMessage.data.nodeId = obm.node;
+                amqpMessage.data.macAddress = obm.config.host;
+
+                return amqpMessage;
+            }else{
+                return waterline.catalogs.findMostRecent({ "data.IP Addresss": ip, source: "bmc"})
+                    .then(function(bmcCatalog) {
+                        if (bmcCatalog !== undefined && bmcCatalog.hasOwnProperty("node") && bmcCatalog.node !== "") {// jshint ignore:line
+                            return extractBmcInfo(amqpMessage, bmcCatalog);
+                        }else{
+                            //unrecognized node
+                            throw new Errors.BadRequestError("unrecognized node");
+                        }
+                    });
+            }
+    }
+ 
+    function extractBmcInfo(amqpMessage, bmcCatalog){
+        amqpMessage.typeId = bmcCatalog.node;
+        amqpMessage.data.nodeId = bmcCatalog.node;
+        amqpMessage.data.macAddress = bmcCatalog.data["MAC Address"];
+
+        return amqpMessage;
+    }
+
+    function updateDeviceInfo(amqpMessage) {
+        amqpMessage.data.ChassisName = null;
+        amqpMessage.data.ServiceTag = null;
+        amqpMessage.data.SN = null;
+
+        return waterline.catalogs.findMostRecent({source: "ipmi-fru"})
+            .then(function(catalogIpmiFru){
+                if(catalogIpmiFru !== undefined && catalogIpmiFru.data.hasOwnProperty("Builtin FRU Device (ID 0)")){// jshint ignore:line
+                    var deviceInfoCatalog = catalogIpmiFru.data["Builtin FRU Device (ID 0)"];
+
+                    //Add useful information to the alert message
+                    if (deviceInfoCatalog.hasOwnProperty("Board Product")){
+                        amqpMessage.data.ChassisName = deviceInfoCatalog["Board Product"];
+                    }
+                    if (deviceInfoCatalog.hasOwnProperty("Product Serial")) {
+                        amqpMessage.data.ServiceTag = deviceInfoCatalog["Product Serial"];
+                    }
+                    if (deviceInfoCatalog.hasOwnProperty("Board Serial")) {
+                        amqpMessage.data.SN = deviceInfoCatalog["Board Serial"];
+                    }
+                }
+                return amqpMessage;
+            });
+
+    }
 
     return new NotificationApiService();
 }
--- a/spec/lib/api/2.0/notification-spec.js
+++ b/spec/lib/api/2.0/notification-spec.js
@@ -26,6 +26,7 @@ describe('Http.Api.Notification', function () {
                 .resolves(nodeNotificationMessage);
             sinon.stub(notificationApiService, 'postBroadcastNotification')
                 .resolves(broadcastNotificationMessage);
+            //sinon.stub(notificationApiService, 'redfishAlertProcessing').resolves();
         });
 
     });
@@ -265,4 +266,39 @@ describe('Http.Api.Notification', function () {
             });
         });
     });
+
+    describe('POST /notification/alerts', function () {
+        var alert = {
+            "Severity":"Critical"
+        };
+        beforeEach(function(){
+            sinon.stub(notificationApiService, 'redfishAlertProcessing').resolves();
+        });
+        afterEach(function(){
+            notificationApiService.redfishAlertProcessing.restore();
+        });
+
+        it('should post alert notification successfully(json) ', function () {
+            return helper.request()
+                .post('/api/2.0/notification/alerts')
+                .set('Content-Type', 'application/json')
+                .send(alert)
+                .expect(201)
+                .then(function () {
+                    expect(notificationApiService.redfishAlertProcessing).to.have.been.calledOnce;// jshint ignore:line
+                });
+        });
+
+        it('should post alert notification successfully(x-www-form-urlencoded)', function () {
+            return helper.request()
+                .post('/api/2.0/notification/alerts')
+                .set('Content-Type', 'application/x-www-form-urlencoded')
+                .send({})
+                .expect(201)  
+                .then(function () {
+                    expect(notificationApiService.redfishAlertProcessing).to.have.been.calledOnce;// jshint ignore:line
+                });
+        });
+    });
+
 });
--- a/spec/lib/services/notification-api-service-spec.js
+++ b/spec/lib/services/notification-api-service-spec.js
@@ -11,6 +11,7 @@ describe('Http.Api.Notification', function () {
     var needByIdentifier;
     var postNodeNotification;
     var postBroadcastNotification;
+    var lookup;
 
     var nodeNotificationMessage = {
         nodeId: '57a86b5c36ec578876878294',
@@ -20,6 +21,44 @@ describe('Http.Api.Notification', function () {
         data: 'dummy data'
     };
     var node = {_id: nodeNotificationMessage.nodeId};
+
+    var bmc = {
+        'IP Address': '10.240.19.130',
+        node : "abc",
+        data : {'MAC Address': "jill"}
+    };
+    var deviceInfoCatalog ={
+        data : {
+            'Builtin FRU Device (ID 0)' : {
+                'Board Product' : "bernie",
+                'Product Serial' : "hillary",
+                'Board Serial' : "theDonald"
+            }
+        }
+    };
+    var alert = {
+        "Severity":"Critical"
+    };
+    var amqpMessage ={action: "alerts",
+        data: {
+            ChassisName: "bernie",
+            SN: "theDonald",
+            ServiceTag: "hillary",
+            Severity: "Critical",
+            macAddress: "jill",
+            nodeId: "abc",
+            sourceIpAddress: "127.0.0.1"
+        },
+        severity: "critical",
+        type: "node",
+        typeId: "abc"
+    };
+    var req = {
+        headers : {
+            'x-forwarded-for' :"127.0.0.1"
+        },
+        swagger: { query: alert}
+    };
 
     var graphId;
     var taskId;
@@ -35,12 +74,14 @@ describe('Http.Api.Notification', function () {
         graphProgressService = helper.injector.get('Services.GraphProgress');
         _ = helper.injector.get('_');
         eventsProtocol = helper.injector.get('Protocol.Events');
+        lookup = helper.injector.get('Services.Lookup');
         waterline = helper.injector.get('Services.Waterline');
         waterline.nodes = {
             needByIdentifier: function() {}
         };
         sinon.stub(eventsProtocol, 'publishNodeNotification').resolves();
         sinon.stub(eventsProtocol, 'publishBroadcastNotification').resolves();
+        sinon.stub(eventsProtocol, 'publishExternalEvent').resolves();
         this.sandbox = sinon.sandbox.create();
         needByIdentifier = sinon.stub(waterline.nodes, 'needByIdentifier');
         needByIdentifier.resolves(node);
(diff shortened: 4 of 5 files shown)


tannoa2:
test this please

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
 
+/**
+ * @api {post} /api/2.0/notification/alerts
+ * @apiDescription Receives Redfish alerts, refactors the message and publishs it into amqp
+ * @apiName notification post
+ * @apiGroup notification
+ */
+
+var notificationAlertsPost = controller({success: 201}, function(req,res) {
+    /* The alert would look like the following:
```
BillyAbildgaard: Do we want this big block of commented code to show an example, or was this just for testing purposes?

static/monorail-2.0.yaml
```diff
@@ -4680,6 +4680,38 @@ paths:
...
     x-swagger-router-controller: workflows
+  /notification/alerts:
+      post:
+        operationId: notificationAlertsPost
+        consumes:
+              - application/x-www-form-urlencoded
+        summary: |
+          Publishes alerts(redfish) received to rackHD
+        description: |
+          Publishes alerts(redfish) received to rackHD1
```
BillyAbildgaard: typo in `rackHD1`

tannoa2: Thanks !

lib/api/2.0/notification.js
tannoa2: This is jsut make it easy for folks reading the code

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
 });
 
+/**
+ * @api {post} /api/2.0/notification/alerts
+ * @apiDescription Receives Redfish alerts, refactors the message and publishs it into amqp
+ * @apiName notification post
+ * @apiGroup notification
+ */
+
+var notificationAlertsPost = controller({success: 201}, function(req,res) {
```
yyscamper: Base our existing practice, the API controller function should be simple, the complex logic and implementation should go into the api service module. For notification, the module is https://github.com/tannoa2/on-http/blob/master/lib/services/notification-api-service.js

So please move your complex logic into `notification-api-service`

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+            var parse = parser.urlencoded({ extended: false });
+            parse(req, res, function (err) {
+                if (err) {
+                    return reject(err);
+                }
+                    resolve();
+            });
+        })
+            .then(function() {
+            return waterline.catalogs.findOne({source: "bmc", "data.IP Address": ip})
```
yyscamper: indent error

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                .then(function(data) {
+                    if (data !== undefined && data.hasOwnProperty("node") && data.node !== ""){
+                         bmcCatalog = data;
+                        //Format the amqp message by adopting the rackHD notification standard
+                        amqpMessage.type = "node";
+                        amqpMessage.action = "alerts";
```
yyscamper: This event is not defined in Event table (http://rackhd.readthedocs.io/en/latest/rackhd/event_notification.html?highlight=event#events-payloads), for new event, please update the doc accordingly.

lib/api/2.0/notification.js
yyscamper: @nortonluo had written an obm-autoset script (https://github.com/RackHD/on-tools/blob/master/obm-autoset/applyobmsetting.py#L239), he found the BMC address is not address in `bmc` catalog,  for some servers, it is stored in `rmm` catalog.

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                                        amqpMessage.data.macAddress = bmcCatalog["MAC Address"];
+                                        amqpMessage.severity = alert.Severity.toLowerCase();
+                                        amqpMessage.data.ChassisName =
+                                            deviceInfoCatalog["Board Product"];
+                                        amqpMessage.data.ServiceTag =
+                                            deviceInfoCatalog["Product Serial"];
+                                        amqpMessage.data.SN = deviceInfoCatalog["Board Serial"];
+
+                                        return notificationApiService.
+                                        alertNotification(amqpMessage);
```
yyscamper: Nesting promise has to be returned for async timing control.

spec/lib/api/2.0/notification-spec.js
```diff
@@ -265,4 +271,75 @@ describe('Http.Api.Notification', function () {
...
+                    'Board Serial' : "theDonald"
+                }
+            }
+        };
+
+
+        var alert = {
+            '{"Severity":"Critical"}': "d"};
+        
+        var amqpMessage ={action: "alerts",
```
yyscamper: A lot indent problems for this object declaration.

spec/lib/api/2.0/notification-spec.js
```diff
@@ -265,4 +271,75 @@ describe('Http.Api.Notification', function () {
...
+                    'Board Product' : "bernie",
+                    'Product Serial' : "hillary",
+                    'Board Serial' : "theDonald"
+                }
+            }
+        };
+
+
+        var alert = {
+            '{"Severity":"Critical"}': "d"};
```
yyscamper: Why define a so weird key `'{"Severity":"Critical"}'`?

static/monorail-2.0.yaml
```diff
@@ -4680,6 +4680,38 @@ paths:
       - Write
       - workflowsModify
     x-swagger-router-controller: workflows
+  /notification/alerts:
+      post:
+        operationId: notificationAlertsPost
+        consumes:
+              - application/x-www-form-urlencoded
```
yyscamper: The indent in yaml is 2 spaces

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                                    }else{
+                                        //publish with the additional information
+                                        return notificationApiService.
+                                        alertNotification(amqpMessage);
+                                    }
+                                });
+                        }else{
```
pengz1: Suggested to use guard clause to improve the long if-else and save one indent

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                            //Unexpected decoded "x-www-forum-urlencoded" data
+                            throw new Errors.BadRequestError(
+                                '"Unexpected decoded "x-www-forum-urlencoded" data"'
+                            );
+                        }
+                    }else{
```
yyscamper: I even can't easily find which is the matching `if`, for better readability, please split this function into some smaller ones. Or you could put the error handling in advance to reduce the code nesting level. 

For example:
```javascript
if (cond) {
    doSomething();
} else {
    throw new Error();
}
```

it could be refactored into:
```javascript
if (!cond) {
    throw new Error();
}
doSomething();
``` 

You could see the function `doSomething()` has  been increased the nesting level to  the same with `if`.

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                                            deviceInfoCatalog["Board Product"];
+                                        amqpMessage.data.ServiceTag =
+                                            deviceInfoCatalog["Product Serial"];
+                                        amqpMessage.data.SN = deviceInfoCatalog["Board Serial"];
+
+                                        return notificationApiService.
+                                        alertNotification(amqpMessage);
+                                    }else{
+                                        //publish with the additional information
+                                        return notificationApiService.
```
pengz1: return notificationApiService.alertNotification(amqpMessage) is duplicated, you should remove the else block and end if block before this sentence instead.

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                            return waterline.catalogs.findOne({source: "ipmi-fru"})
+                                .then(function(catalogIpmiFru){
+                                    if(catalogIpmiFru !== undefined &&
+                                        catalogIpmiFru.data
+                                            .hasOwnProperty("Builtin FRU Device (ID 0)")){
+                                        deviceInfoCatalog =
+                                            catalogIpmiFru.data["Builtin FRU Device (ID 0)"];
+
+                                        //Add useful information to the alert message
+                                        amqpMessage.data.rackhdId = bmcCatalog.node;
```
yyscamper: Change `rackhdId` to `nodeId`

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                                        amqpMessage.data.macAddress = bmcCatalog["MAC Address"];
+                                        amqpMessage.severity = alert.Severity.toLowerCase();
```
pengz1: Should above 3 sentences moved out of this if block? Those assignments should still be valid even if there is no ipmi fru catalog.

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                        amqpMessage.type = "node";
+                        amqpMessage.action = "alerts";
+                        amqpMessage.typeId = bmcCatalog.node;
+
+                        var httpReq = _.defaults(req.swagger.query || {},
+                            req.query || {},
+                            req.body || {});
+                        if(Object.keys(httpReq).length === 1 ) {
+                            var key = Object.keys(httpReq)[0];
+                            alert = JSON.parse(key);
```
yyscamper: Why the `key` need be parsed? What kind of key is so complex? please give some example of the input.

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                            amqpMessage.data = alert;
+                            amqpMessage.data.sourceIpAddress = ip;
+                            return waterline.catalogs.findOne({source: "ipmi-fru"})
```
yyscamper: There may be multiple data for the same catalog source, the most recent one should be the most correct one. please reference https://github.com/RackHD/on-tasks/blob/master/lib/jobs/get-catalog-values-job.js#L68

lib/api/2.0/notification.js
```diff
@@ -30,7 +36,124 @@ var notificationProgressPost = controller(function(req, res){
...
+                                    }else{
+                                        //publish with the additional information
+                                        return notificationApiService.
+                                        alertNotification(amqpMessage);
+                                    }
+                                });
+                        }else{
+                            //Unexpected decoded "x-www-forum-urlencoded" data
+                            throw new Errors.BadRequestError(
+                                '"Unexpected decoded "x-www-forum-urlencoded" data"'
```
yyscamper: typo: x-www-forum-urlencoded => x-www-form-urlencoded

lib/api/2.0/notification.js
yyscamper: If this is only to collect redfish alert, then change the `action` to a more redfish specific one will make sense, such as `redfish.alert`

tannoa2: Because we use the `action` property in the routing key, I avoiding the use of dot notation because it will add confusion.
http://rackhd.readthedocs.io/en/latest/rackhd/event_notification.html?highlight=event#events-payloads .
`Routing Key <type>.<action>.<severity>.<typeId>.<nodeId>`
With your suggestion, a person might confuse  `alert` as a `severity`.
Do you have any one word suggestion ?

lib/api/2.0/notification.js
tannoa2: I wasn't aware of that, I will add the hooks to handle this scenario

spec/lib/api/2.0/notification-spec.js
tannoa2: Th new line is just to accommodate the line too long hound warnings

spec/lib/api/2.0/notification-spec.js
tannoa2: This is  how the body-parser is decoding the x-www-form-urlencoded data. 
The content or data is the in the `key` not the `value`. 
We have a story to look into the root cause of this https://rackhd.atlassian.net/browse/RAC-4699

tannoa2:
@yyscamper for your question about `x-forwarded-for`, it is being used to identify the IP of the originating client. Again, this was dictated  by the "community", I will forward you some of the correspondence.

As far as of why we are decoding  `x-www-form-urlencoded` traffic, because we are expecting `x-www-form-urlencoded` `Content-Type`. 

As for the feature itself, we are trying to enable RackHD to receive redfish alerts.
For more info here is the spec for the redfish alerting: http://redfish.dmtf.org/schemas/DSP0266_1.1.html#service-details

lib/api/2.0/notification.js
tannoa2: Here we go: 
`{ '{"Context":"context string","EventId":"8491","EventTimestamp":"2017-03-30T11:27:15-0500","EventType":"Alert","MemberId":"7e675c8e-127a-11e7-9fc8-64006ac35232","Message":"Successfully logged in using root, from 10.252.79.21 and GUI.","MessageArgs":["root","10.252.79.21","GUI"],"MessageArgs@odata.count":3,"MessageId":"USR0030","OriginOfCondition":"iDRAC.Embedded.1","Severity":"Informational"}\n': '' }`

lib/api/2.0/notification.js
tannoa2: Here we go:
`{ '{"Context":"context string","EventId":"8491","EventTimestamp":"2017-03-30T11:27:15-0500","EventType":"Alert","MemberId":"7e675c8e-127a-11e7-9fc8-64006ac35232","Message":"Successfully logged in using root, from 10.252.79.21 and GUI.","MessageArgs":["root","10.252.79.21","GUI"],"MessageArgs@odata.count":3,"MessageId":"USR0030","OriginOfCondition":"iDRAC.Embedded.1","Severity":"Informational"}\n': '' }`

lib/services/notification-api-service.js
```diff
@@ -109,6 +109,155 @@ function NotificationApiServiceFactory(
...
+    function getAlert (req) {
+        var amqpMessage = {};
+        var ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress ||
+            req.socket.remoteAddress || req.connection.socket.remoteAddress;
+        var httpBody = _.defaults(req.swagger.query || {},req.query || {}, req.body || {});
+
+        //Doing some processing for the request body.cd
+        //It looks like the body parser decoder puts the value of the req.body in the key.
+        //example: {'{"Severity":"Critical"}': ""};
+        if(Object.keys(httpBody).length === 1 ) {
```
brianparry: @tannoa2 Do we know why this parser cleanup is needed?  Maybe another alternative is to pre-process the req prior to passing it into the parser.  I'm a little concerned that this behavior might not be consistent so there may be cases where our post-processing fails.

lib/api/2.0/notification.js
```diff
@@ -30,7 +32,45 @@ var notificationProgressPost = controller(function(req, res){
...
+                //This is a work arround where some vendors are sending json data
+                // with http header content-type of “x-www-form-urlencoded”
+                twistedParser(req, res, function (err) {
+                    if (err) {
+                        return reject(err);
+                    }
+                    resolve();
+                });
+            }
+            else{
```
brianparry: @tannoa2 I don't think this else case is needed for `application/json`.  The swagger params parser should handle this case.

lib/services/notification-api-service.js
```diff
@@ -109,6 +109,147 @@ function NotificationApiServiceFactory(
...
+                    .then(function () {
+                        return amqpMessage;
+                    });
+            });
+    };
+
+    function getAlert (req) {
+        var amqpMessage = {};
+        var ip = req.headers['x-forwarded-for'] || req.connection.remoteAddress ||
+            req.socket.remoteAddress || req.connection.socket.remoteAddress;
```
yyscamper: in http-service, there is a function to retrieve the request IP address (see https://github.com/RackHD/on-http/blob/master/lib/services/http-service.js#L25), its implementation is not the same with yours, please keep both aligned.

lib/api/2.0/notification.js
```diff
@@ -30,7 +32,45 @@ var notificationProgressPost = controller(function(req, res){
...
+ * @apiName notification post
+ * @apiGroup notification
+ */
+var twistedParser = parser.json({ type: 'application/x-www-form-urlencoded' });
+var jsonParser = parser.json();
+var notificationAlertsPost = controller({success: 201}, function(req,res) {
+    var format= req.headers["content-type"];
+    return Promise.try(function(){
+        return new Promise(function(resolve, reject) {
+            if (format === 'application/x-www-form-urlencoded'){
```
yyscamper: From @brianparry 's explanation in email,  there is firmware bug that lead you have to do content-type parsing at here. Please add some comment here to let other developer be aware of such known issue.

lib/services/notification-api-service.js
```diff
@@ -109,6 +109,147 @@ function NotificationApiServiceFactory(
...
+                return  updateBmcInfo(alert);
+            })
+            .then(function(data){
+                return updateDeviceInfo(data);
+            })
+            .then(function(data){
+                amqpMessage = data;
+                return eventsProtocol.publishExternalEvent(data)
+                    .then(function () {
+                        return amqpMessage;
```
yyscamper: the variable `amqpMessage` can be removed, just directly return `data` at here.

lib/services/notification-api-service.js
```diff
@@ -109,6 +109,147 @@ function NotificationApiServiceFactory(
...
+
+        return waterline.catalogs.findMostRecent({ "data.IP Addresss": ip, source: "bmc"})
+            .then(function(bmcCatalog) {
+                if (bmcCatalog !== undefined && bmcCatalog.hasOwnProperty("node") && bmcCatalog.node !== "") {// jshint ignore:line
+                    return extractBmcInfo(amqpMessage,bmcCatalog);
+                }
+                if (bmcCatalog === undefined){
+                    //Sometimes the IP in the bmc catalog is 0.0.0.0.
+                    //In this case will try to get the IP fromm the obms collection
+                    return waterline.obms.findOne({"config.host": ip})
```
yyscamper: The `config.host` sometimes is a MacAddress, this is why this util function is defined (see https://github.com/RackHD/on-tasks/blob/master/lib/utils/job-utils/job-helper.js#L33 ). If this is a MacAddress，then you don't need to lookup the catalog.

lib/services/notification-api-service.js
```diff
@@ -109,6 +109,147 @@ function NotificationApiServiceFactory(
...
+            throw new Errors.BadRequestError(
+                'Unexpected alert data'
+            );
+        }
+    }
+
+    function updateBmcInfo(amqpMessage) {
+        var ip = amqpMessage.data.sourceIpAddress;
+
+        return waterline.catalogs.findMostRecent({ "data.IP Addresss": ip, source: "bmc"})
```
yyscamper: I think the obms settings should take precedent because we should first trust user's setting, then turn to lookup in raw catalog. The order to find the BMC  mac address is:
- If the `host` in ipmi-obm-service is mac address, then use it directly.
- Then if the `host` in ipmi-obm-service is IP address, then check the lookup to find corresponding macAddress, if exists, then use it.
- Check the raw catalog.

yyscamper: Meanwhile, I also mentioned before, sometimes the BMC information is stored in `rmm` catalog, not in `bmc`. I know this is terrible, but this is our current situation.

lib/services/notification-api-service.js
```diff
@@ -109,6 +109,147 @@ function NotificationApiServiceFactory(
...
+            .then(function(bmcCatalog) {
+                if (bmcCatalog !== undefined && bmcCatalog.hasOwnProperty("node") && bmcCatalog.node !== "") {// jshint ignore:line
+                    return extractBmcInfo(amqpMessage,bmcCatalog);
+                }
+                if (bmcCatalog === undefined){
+                    //Sometimes the IP in the bmc catalog is 0.0.0.0.
+                    //In this case will try to get the IP fromm the obms collection
+                    return waterline.obms.findOne({"config.host": ip})
+                        .then(function(node) {
```
yyscamper: Change the variable name from `node` to `obm`

lib/services/notification-api-service.js
```diff
@@ -109,6 +109,147 @@ function NotificationApiServiceFactory(
...
+                            if (node !== undefined) {
+                                return waterline.catalogs.findMostRecent({"node": node.node, source: "bmc"})// jshint ignore:line
+                                    .then(function (bmcCatalog) {
+                                        if (bmcCatalog !== undefined && bmcCatalog.hasOwnProperty("node") && bmcCatalog.node !== "") { // jshint ignore:line
+                                            return extractBmcInfo(amqpMessage, bmcCatalog);
+                                        }
+                                    });
+                            }else{
+                                //unrecognized node
+                                throw new Errors.BadRequestError("unrecognized node");
```
yyscamper: The error message is not correct, this is `unrecoginized node`, this just means the ipmi obm service is not existing.

lib/services/notification-api-service.js
yyscamper: Find the obm service should specify the obm service name, because some non-ipmi obm service still has a filed `config.host`.

static/monorail-2.0.yaml
```diff
@@ -4680,6 +4680,39 @@ paths:
       - Write
       - workflowsModify
     x-swagger-router-controller: workflows
+  /notification/alerts:
```
yyscamper: This API belongs to southbound, please move it to `monorail-2.0-sb.yaml`

spec/lib/services/notification-api-service-spec.js
```diff
@@ -152,5 +200,53 @@ describe('Http.Api.Notification', function () {
...
+            return notificationApiService.redfishAlertProcessing(req)
+                .then(function (resp) {
+                    expect(resp).to.deep.equal(amqpMessage);
+                });
+        });
+
+        it('should post an alert notification successfully(from obms collection)', function () {
+            this.sandbox.restore();
+            waterline.catalogs.findMostRecent.onCall(0).resolves(undefined);
+            waterline.obms.findOne.onCall(0).resolves({confid : {host: "ip"}});
```
yyscamper: typo: `confid` ==> `config`

spec/lib/services/notification-api-service-spec.js
```diff
@@ -152,5 +200,53 @@ describe('Http.Api.Notification', function () {
...
+                    expect(resp).to.deep.equal(amqpMessage);
+                });
+        });
+
+        it('should NOT post an alert notification for an unrecognized node', function () {
+            this.sandbox.restore();
+            waterline.catalogs.findMostRecent.onCall(0).resolves(undefined);
+            waterline.obms.findOne.onCall(0).resolves(undefined);
+            return notificationApiService.redfishAlertProcessing(req)
+                .catch(function (e) {
```
yyscamper: To expect a promise to be caught, you should explicitly fail the test in `then` block. Otherwise, if promise is actually fulfilled (out-of exception), the mocha still marks it as success. We usually do this is following way:
```javascript
it('should expect an error', function(done) {
  return notificationApiService.redfishAlertProcessing(req)
  .then(function() {
    done(new Error('should never run here!'));
  })
  .catch(function(err) {
    expect(e).to.have.property('message').that.equals('unrecognized node');
    done();
  });
});
```

spec/lib/api/2.0/notification-spec.js
```diff
@@ -264,5 +264,26 @@ describe('Http.Api.Notification', function () {
...
+            var alert = {
+                '{"Severity":"Critical"}': "d"};
+            before(function(){
+                sinon.stub(notificationApiService, 'redfishAlertProcessing').resolves();
+            });
+
+            it('should post alert notification successfully', function () {
+                return helper.request()
+                    .post('/api/2.0/notification/alerts')
+                    .set('Content-Type', 'application/x-www-form-urlencoded')
```
yyscamper: Since you specify the API will consume `application/json` as well. so please create a test case to make sure the `application/json` works as well.

spec/lib/api/2.0/notification-spec.js
```diff
@@ -264,5 +264,26 @@ describe('Http.Api.Notification', function () {
                 .expect(400);
             });
         });
+
+        describe('POST /notification/alerts', function () {
+            var alert = {
+                '{"Severity":"Critical"}': "d"};
+            before(function(){
+                sinon.stub(notificationApiService, 'redfishAlertProcessing').resolves();
```
yyscamper: Please restore this stub after test suites finished, so as not to bring side effect for other test cases.

static/monorail-2.0.yaml
brianparry: @yyscamper This won't work because the alerts are coming back over https.  The monorail-2.0-sb.yaml needs to be removed anyway.  This is part of the 1.1 API removal story in the Stryker backlog.

lib/services/notification-api-service.js
brianparry: @yyscamper You make a good point about the `rmm` catalog.  In the short term though this is only used for Dell SKUs so I think we can add this later.

lib/services/notification-api-service.js
tannoa2: @yyscamper  it actually means that the node wasn't found neither by looking the obm collection or the bmc catalog.  I will leave the error description as it is because it is accurate.

static/monorail-2.0.yaml
tannoa2: @yyscamper from what I understand, the monorail-2.0-sb.yaml is not being used. I won't be adding the API there. It is actually going away.....https://rackhd.atlassian.net/browse/RAC-1447

lib/api/2.0/notification.js
tannoa2: It is already commented with this:
`//This is a work around where some vendors are sending json data
// with http header content-type of “x-www-form-urlencoded”`

lib/services/notification-api-service.js
tannoa2: @yyscamper The implementation above work. we are keeping it

lib/services/notification-api-service.js
tannoa2: For now the the bmc catalog will do. Will add the rmm in the future.

lib/services/notification-api-service.js
tannoa2: @yyscamper what is needed the opposite to that in the link. We have the IP, we need to find out the mac. I got your point though, and I will fix it

lib/services/notification-api-service.js
tannoa2: actually what is needed is `.tap`

static/monorail-2.0.yaml
yyscamper: @brianparry @tannoa2 : I didn't realize you were planing to remove the southbound API, but as I know the splitting of southbound API and northbound API is an important part for our forthcoming micro services design. The southbound API is only internal API, it should not be restricted by semantic version, and we may even has some anti-RESTful design if the node ability is limited. If both are merged, if southbound API is still visible to user, how could we apply the difference policy for southbound and northbound? Maybe you already have the solution to handle it within same swagger config file, could you please tell me some detail?

anhou: I also have concern for southbound and northbound API merge. some comments are updated to https://rackhd.atlassian.net/browse/RAC-1447

lib/services/notification-api-service.js
yyscamper: If so, please add a TODO comment at here to avoid forgetting.

lib/services/notification-api-service.js
yyscamper: I know this code will work in your system, but in the long run, this kind of code divergence will bring problems, and something even hard to debug.

For example, this function (https://github.com/RackHD/on-http/blob/master/lib/services/http-service.js#L25) will return `X-Real-IP` as source, this is part of API proxy feature, but your code code doesn't use `X-Real-IP`, so as the API proxy feature will not touch this particular API.

static/monorail-2.0.yaml
brianparry: @yyscamper @anhou We've discussed this in the past.  We will use privileges and firewall rules to separate southbound and northbound routes.  In this case the Dell server uses https, not http, so even without further refactoring, it cannot use the southbound router.  It needs to use port 8443.  This is an entirely seperate discussion that is outside the scope of this PR.

tannoa2:
test this please

lib/services/notification-api-service.js
keedya: @yyscamper we don't support X-Real-IP and this should be replaced by x-forward-for instead.
In this case req.ip will have the forwarded IP from the original client
trustedProxy	Enable trust proxy in express. Populate req.ip with left most IP address from the XForwardFor list.
https://github.com/RackHD/on-http/blob/master/lib/services/http-service.js#L277

keedya: http://rackhd.readthedocs.io/en/latest/rackhd/configuration.html#configuration-parameters

geoff-reid:
test this please

static/monorail-2.0.yaml
anhou: @brianparry  firewall and privilege control only cover security problem. southbound and northbound API splitting don't only cover security, but also have other considerations, please see my comments at https://rackhd.atlassian.net/browse/RAC-1447.  I also don't want to this part impact your progress, so I commented  in RAC-1447.

PR #622, opened 2017-03-29
Fix for auto-discovery

dalebremner:
- Add SNMP settings to node schema
- Remove deserializer from 2.0 nodes
- Change autoDiscover property from string to boolean


BillyAbildgaard:
@dalebremner I tried to make this change from string to boolean before and saw the same test failures you are.  

'Content-Type': 'application/json; charset=utf-8'})\nHTTP response body: {"message":"Validation errors","status":"400","UUID":"76a1b12a-d2d5-44b6-b7ce-94806362aa7a","errors":["Expected type boolean but found type string"]}

What happens is even though you are expecting a boolean the value gets passed in with quotes and so the validator sees it as a string

static/schemas/2.0/node.2.0.json
```diff
@@ -25,6 +25,10 @@
                         "type": "object"
                     }
                 },
+                "snmpSettings": {
```
yyscamper: Isn't the snmpSettings has been moved into `ibms` or `obms`? @brianparry @srinia6

yyscamper:
There are still some code need be fixed, such as:
- https://github.com/RackHD/on-http/blob/master/data/views/node.2.0.json#L2
- https://github.com/RackHD/on-http/blob/master/spec/lib/api/redfish-1.0/system-spec.js#L141
- https://github.com/RackHD/on-http/blob/master/spec/lib/api/redfish-1.0/system-spec.js#L156
- https://github.com/RackHD/on-http/blob/master/spec/lib/api/2.0/skus-spec.js#L81
- (on-tasks): https://github.com/RackHD/on-tasks/blob/master/spec/lib/jobs/ucs-service-profile-discovery-spec.js#L131
- maybe more

Please do a "grep" in all repository (not only on-http) repository to fix all.

PR #624, opened 2017-04-06
Add/update schemas, and reference them in response object

srinia6:
This PR:
1. Addresses RAC-4622, to update/correct schema.
2. Reference them in responses. 
    The original issue was seen while generating client libraries in GO/JAVA, where return types are all list, 
    and the user has to manually parse the returned data. This addresses that.
3. Fix few responses return type, array of object VS just objects.

@RackHD/corecommitters @jasl8r

### DIFF
--- a/static/monorail-2.0.yaml
+++ b/static/monorail-2.0.yaml
@@ -138,7 +138,7 @@ definitions:
       relations:
         type: array
         items:
-          $ref: "#/definitions/relations.2.0_Relation"
+          $ref: "#/definitions/relations_obj"
       tags:
         items:
           type: string
@@ -178,6 +178,8 @@ definitions:
     type: object
     properties:
       id:
+        type: string
+      node:
         type: string
       config:
         type: object  
@@ -247,23 +249,19 @@ definitions:
             type: object
         type: object
     type: object
-  relations.2.0_Relation:
-    description: A key value set of relations.
+  relations_obj:
+    description: A key value set of node relations.
     properties:
-      containedBy:
-        description: The rack containing the node.
+      relationType:
+        description: Relation Type with the node.
+        type: string
+      info:
+        type: string
+      targets:
+        description: Array of targets.
         items:
           type: string
-        minItems: 1
         type: array
-        uniqueItems: true
-      contains:
-        description: An array of nodes contained by the rack.
-        items:
-          type: string
-        minItems: 1
-        type: array
-        uniqueItems: true
     type: object
   role_obj:
     properties:
@@ -1674,7 +1672,7 @@ paths:
           description: Successfully retrieved the nodes relations
           schema:
             items:
-              $ref: '#/definitions/relations.2.0_Relation'
+              $ref: '#/definitions/relations_obj'
             type: array
         404:
           description: The specified node was not found.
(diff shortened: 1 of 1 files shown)


static/monorail-2.0.yaml
```diff
@@ -158,6 +174,15 @@ definitions:
         type: string
       value:
         type: boolean
+  obm:
+    type: object
+    properties:
```
yyscamper: missing the `node` property

static/monorail-2.0.yaml
```diff
@@ -221,6 +247,24 @@ definitions:
             type: object
         type: object
     type: object
+  relations.2.0_Relation:
+    description: A key value set of relations.
+    properties:
+      containedBy:
```
yyscamper: Besides `containedBy` and `contains`, there are still other relations (see https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L256 )

static/monorail-2.0.yaml
```diff
@@ -221,6 +247,24 @@ definitions:
...
+        type: array
+        uniqueItems: true
+      contains:
+        description: An array of nodes contained by the rack.
+        items:
+          type: string
+        minItems: 1
+        type: array
+        uniqueItems: true
+    type: object
```
yyscamper: Actually, only one relation type is allowed within a relation instance, so use `oneOf` will clearly describe such kind of restriction. For example:
```json
properties:
  containedBy:
    ...
    type: array
  contains:
    ...
    type: array
required:
  - oneOf
    - containedBy
    - contains
```

static/monorail-2.0.yaml
```diff
@@ -3078,7 +3130,9 @@ paths:
         200:
           description: Successfully retrieved the list of SKUs
           schema:
-            type: object
+            type: array
+            items:
+             $ref: '#/definitions/skus.2.0_SkusUpsert'
```
yyscamper: indent 2 spaces.

static/monorail-2.0.yaml
```diff
@@ -3301,7 +3355,9 @@ paths:
         200:
           description: Successfully retrieved the nodes associated with the specified SKU
           schema:
-            type: object
+            type: array
+            items:
+             $ref: '#/definitions/node.2.0_PartialNode'
```
yyscamper: indent 2 spaces.

static/monorail-2.0.yaml
yyscamper: Seems our code only enables the containedBy/contains relation at present, so I am OK to keep not adding other relations in your schema.

srinia6: Yup, as we discussed, this might not be an issue. Thanks for confirming!

static/monorail-2.0.yaml
srinia6: I understand,  but I use the same schema for GET, I left them out [ and for now, the secondary schema validation will cover this case ]. With this change, we would have to re-visit to remove that one too , which is out of scope of this story . Possibly I can add two. I have to give some thoughts there.

srinia6: @yyscamper I see swagger 2.0 supports only allOf in yaml. I don't see oneOf/anyOf working.

static/monorail-2.0.yaml
srinia6: Has been added.

yyscamper:
+1

PR #626, opened 2017-04-10
Enhance docker build of on-http

panpan0000:
On behalf of @PengTian0 , this PR was aim to address  P1     RAC-4784, but it's resolved by @BillyAbildgaard  in other PR which directlly remove apidoc.
**But this PR is still valid.**

-----------

    **Change Summary**
    1. add --unsafe-perm , to enable '''npm install''' run as root
    2. add  "npm config set -g production false"
    3. remove explict step in Dockerfile to run install-web-ui.sh & install-swagger-ui.sh
    4. remove duplicated step of "npm install apidoc"
-----------

    **Rational**
-----------

    (1) why "--unsafe-perm"
    in Docker-Build, the "RUN" shell commands are run as "root"
    But when running npm with "root", NPM tries to downgrade its privileges when it runs scripts(defaults to nobody)
    And it will fail to write in local folder without enough permission, so
    NPM skips the command line ""./install-web-ui.sh && ./install-swagger-ui.sh", which is the "install" steps under "scripts" in package.json
    Warning Message as below:
    ```
     npm WARN cannot run in wd on-http@2.1.0 ./install-web-ui.sh && ./install-swagger-ui.sh
    ```

    Per NPM Usage, "--unsafe-perm" is required if running npm as root or sudo.

    *unsafe-perm*
    Set to true to suppress the UID/GID switching when running package scripts. If set explicitly to false, then installing as a non-root user will fail.

    reference
    http://stackoverflow.com/questions/18136746/npm-install-failed-with-cannot-run-in-wd
    https://til.codes/npm-install-failed-with-cannot-run-in-wd-2/

--------------

    (2) By default, the "npm config get production" will show "true" in base docker image(nodesource/wheezy:4.4.6).
    So whenever run "npm install", it's equal to running "npm install --production".
    So the DevDependency are all missing. which makes the 'npm run apidoc' failed to find html-pdf...etc packages.

------------
    (3) why not explicit "install-web-ui.sh & install-swagger-ui.sh"

    Previouly, when @Roland initialized the Dockerfile, He met the same problem which " npm WARN cannot run in wd on-http@2.1.0 ./install-web-ui.sh && ./install-swagger-ui.sh"
    He brought the *workaround* that split the other npm-install steps with the  web-ui/swagger-ui.sh.
    That's why he run "npm install --ignore-script" to skip the "./install-web-ui.sh && ./install-swagger-ui.sh" at the beginning,
    then explictly run "./install-web-ui.sh && ./install-swagger-ui.sh" after that (out of npm control)

    with the real fix of "--unsafe-perm" , npm-install will cover the step of  "./install-web-ui.sh && ./install-swagger-ui.sh" .

------------
   
 (4) in package.json, "apidoc" is installed with a specific version( "^0.12.1") in package.json(DevDependencies). so no necessary to do it here again.

@yyscamper  @changev  @brianparry  @BillyAbildgaard

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,9 @@ RUN mkdir -p ./node_modules \
   && ln -s /RackHD/on-core ./node_modules/on-core \
   && ln -s /RackHD/on-core/node_modules/di ./node_modules/di \
   && apt-get install -y unzip curl \
-  && npm install --ignore-scripts \
-  && npm install apidoc \
+  && npm config set -g production false \
+  && npm install --unsafe-perm \
   && npm run taskdoc \
-  && /RackHD/on-http/install-web-ui.sh \
-  && /RackHD/on-http/install-swagger-ui.sh \
   && npm prune --production
 
 EXPOSE 9080 9090


PengTian0:
:+1:

panpan0000:
@BillyAbildgaard , Yes. your PR did resolve it with apidoc removing.
But the "--unsafe-perm" is still needed when run "npm install" as root.
it's better than previous workaround of "splitting" npm-install into 2 steps.

BillyAbildgaard:
@panpan0000 I don't think we should be taking steps to make our product less safe.  I see no benefit to running npm install as root, which can be more dangerous than simply doing the npm install scripts separately.

panpan0000:
@BillyAbildgaard 
as my offline email:

In our docker build,  the only user is “root”, it’s not intended to run npm install as root in that step , but no choice. 
So I believe that’s why @Roland split the single “npm-install” into 3 steps as a workaround.
As I wrote, If that’s the purpose of splitting the steps into 2 , then "--unsafe-perm” is better than splitting into 3 steps.

this PR is not to make code less safety. the original code already did the same thing and same result with a workaround.  But the original  workaround is to run "install-web-ui.sh & install-swagger-ui.sh" outside 'npm-install', which will cause big confusion to people that it seems an redundant step. and it's not a correct way to use 'npm install'

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

PengTian0:
test this please

panpan0000:
test this please

panpan0000:
test this please

changev:
@panpan0000 agree to @keedya

PR #627, opened 2017-04-10
removing apidoc

BillyAbildgaard:
removing apidoc command.  This was used to generate 1.1 documentation and since we have swagger-ui it seems to add no additional value by also generating this documentation.

@RackHD/veyron

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,6 @@ RUN mkdir -p ./node_modules \
   && apt-get install -y unzip curl \
   && npm install --ignore-scripts \
   && npm install apidoc \
-  && npm run apidoc \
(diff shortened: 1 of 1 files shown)


keedya:
+1

Dockerfile
```diff
@@ -12,10 +12,7 @@ RUN mkdir -p ./node_modules \
   && apt-get install -y unzip curl \
   && npm install --ignore-scripts \
   && npm install apidoc \
-  && npm run apidoc \
   && npm run taskdoc \
-  && /RackHD/on-http/install-web-ui.sh \
```
geoff-reid: Why are you also removing the web-ui and swagger-ui?

PR #632, opened 2017-04-20
 Update secure erase to use latest progress notification api

pengz1:
Changes include:
1. Existing secure_erase.py use a progress notification payload like:
{"taskID": xxx, "progress": {"value":0, "maximum"}}
Our latest progress notification payload require payload to be like:
{"taskID": xxx, "value":0, "maximum"}
Update secure_erase.py to adopt latest payload format.
2. Set default secure erase in Progress class, otherwise progress notification may not posted to RackHD
3. Other changes include: 
    Change Progress class to new type;
    Correct Message to message;
    Set default pass_count to 1.

### DIFF
--- a/data/templates/secure_erase.py
+++ b/data/templates/secure_erase.py
@@ -71,7 +71,7 @@ else:
     PATH = os.getcwd()
 
 
-class Progress:
+class Progress(object):
     """
     Secure erase job progress class.
     """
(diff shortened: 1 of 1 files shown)


pengz1:
@yyscamper @lanchongyizu

data/templates/secure_erase.py
```diff
@@ -244,12 +242,12 @@ def __run(self):
                     erase_start_flags[index] = os.path.exists(self.path + value + '.log')
             total_percent = sum(percentage_list)/disk_count
             if total_percent == float('+inf'):
-                payload["progress"]["percentage"] = "Not Available"
+                payload["percentage"] = "Not Available"
             else:
-                payload["progress"]["percentage"] = str("%.2f" % total_percent) + "%"
-                payload["progress"]["value"] = int(total_percent)
+                payload["percentage"] = str("%.2f" % total_percent) + "%"
+                payload["value"] = int(total_percent)
```
yyscamper: In latest progress design, the property `percentage` will be ignored, any progress need be split into `value` and `maximum`. In your case, you could set a constant `maximum` as `100`.

pengz1: maximum is 100, percentage was already there, I just sent in case progress can save calculation.

pengz1:
test this please

data/templates/secure_erase.py
yyscamper: Thanks, make sense

PR #643, opened 2017-05-11
 Decouple-secure-erase-from-RAID-operation

pengz1:
Move RAID related code to be only executed when disk is with RAID.
This change can make erase cases like all disks are JBOD , or SATADOM/NVMe can be executed.


pengz1:
@iceiilin @anhou @sunnyqianzhang @lanchongyizu

panpan0000:
test this please

panpan0000:
test this please

PR #645, opened 2017-05-11
Remove authentication for the /POST notification route

tannoa2:
@RackHD/veyron @RackHD/corecommitters

### DIFF
--- a/static/monorail-2.0.yaml
+++ b/static/monorail-2.0.yaml
@@ -4767,8 +4767,6 @@ paths:
               $ref: '#/definitions/Error'
         tags:
         - /api/2.0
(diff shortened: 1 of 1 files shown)


static/monorail-2.0.yaml
```diff
@@ -4767,8 +4767,6 @@ paths:
               $ref: '#/definitions/Error'
         tags:
         - /api/2.0
-        x-authentication-type:
```
brianparry: @tannoa2 You should probably remove the privileges too since they aren't used without authentication

amymullins:
Closing PR, changes were part of another PR.

PR #647, opened 2017-05-11
Added PUT to config

mtannous:
@danbmckay and I added PUT rest call which essentially behaves as a PATCH. We are using swagger codegen and it doesnt support PATCH unless we have a POST call which /config doesn't have.
 
Shown below is the generated code from swagger codegen:
```
if ("GET".equals(method)) {
      response = invocationBuilder.get();
    } else if ("POST".equals(method)) {
      response = invocationBuilder.post(entity);
    } else if ("PUT".equals(method)) {
      response = invocationBuilder.put(entity);
    } else if ("DELETE".equals(method)) {
      response = invocationBuilder.delete();
    } else if ("PATCH".equals(method)) {
      response = invocationBuilder.header("X-HTTP-Method-Override", "PATCH").post(entity);
    } else {
      throw new ApiException(500, "unknown method type " + method);
    }
```

### DIFF
--- a/static/monorail-2.0.yaml
+++ b/static/monorail-2.0.yaml
@@ -125,7 +125,7 @@ definitions:
     description: Post a node into RackHD
     properties:
       autoDiscover:
-        type: string
+        type: boolean
       bootSettings:
         type: object
       identifiers:
--- a/static/monorail-2.0.yaml
+++ b/static/monorail-2.0.yaml
@@ -530,6 +530,14 @@ paths:
           description: Unexpected error
           schema:
             $ref: '#/definitions/Error'
+      summary: Patch server configuration
+      tags:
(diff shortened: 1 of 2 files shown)


geoff-reid:
test this please

static/monorail-2.0.yaml
```diff
@@ -530,6 +530,27 @@ paths:
...
+      responses:
+        200:
+          description: Successfully modified the configuration
+          schema:
+            type: object
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
       summary: Patch server configuration
```
geoff-reid: You need move summary, tags, x-authentication-type, and x-privileges back to the PATCH and then create these entries for PUT

PR #650, opened 2017-05-17
keep data consistency in enclosure update

changev:
Fix https://rackhd.atlassian.net/browse/RAC-4805

jenkins: depends on https://github.com/RackHD/on-core/pull/278

Enclosure node will be updated when deleting compute node.
Find enclosure -> modify -> write

Used:
* find enclosure, delete the compute id in targets and get a new  enclosure [object], then ```update``` enclosure with this [object].
* this is not atomic

Now:
* find enclosure, use ```findAndModify``` to delete compute id in its target. If target = [], use ```findAndModify``` to delete the relation too
* this is atomic

### DIFF
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -142,6 +142,63 @@ function nodeApiServiceFactory(
     };
 
     /**
+     * Get the targets to be removed in node. If the node is invalid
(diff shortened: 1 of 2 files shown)


changev:
test this please

changev:
@iceiilin 
```_removeRelation()``` is also used in ```relations add```
I'll check it later after this PR merged

PR #655, opened 2017-05-25
Change parsing source for SATA drives ESXi WWID

pengz1:
Issue is encountered to intall ESXi on a Dell R730. It is found ESXi may parse SATA drive WWID differently. So I changed parsing source for SATA drives ESXi WWID:
1. Original source: Ubuntu 'ls -l /dev/disk/by-id' command output string
 2. New source: raw hex data got from 'hdparm --Istdout /dev/sdx' command

Also to aligned with what we found, an ASCII code 0x00 will be decoded as '00' instead of an empty string.
Related RAC:
https://rackhd.atlassian.net/browse/RAC-5081

### DIFF
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -14,11 +14,12 @@ var options = {
 };
 
 /**
- * Get SATA drive's SN string.
+ * Get SATA drive's logic unit name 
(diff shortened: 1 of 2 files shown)


pengz1:
@keedya @anhou @uppalk1 @iceiilin

PR #658, opened 2017-05-26
keep data consistency when add realtions

changev:
jenkins: depends on https://github.com/RackHD/on-core/pull/280

Modified 
```
_addRelation
_removeRelation
EditRelation
```
to keep data consistency

### DIFF
--- a/lib/services/nodes-api-service.js
+++ b/lib/services/nodes-api-service.js
@@ -72,18 +72,19 @@ function nodeApiServiceFactory(
     };
 
     /**
-     * Remove the relations to original node in target node. If the node is invalid
+     * Remove targets from node relations. If the node is invalid
      * or doesn't have required relation, this function doesn't need to update the
-     * node info and ignore silently with Promise.resolve(). If the relation field
-     * is empty after removing, delete this field. If the node to be updated is a component
-     * of a node delete it
+     * node info and ignore silently with Promise.resolve(). 
+     * If the targets list of one relation item is empty after removing, delete this relations. 
      * @param  {Object}     node node whose relation needs to be updated
      * @param  {String}     type relation type that needs to be updated
      * @param  {String[] | Object[]}    targets - nodes or ids in the relation
      * that needs to be deleted
      * @return {Object}    node after removing relation
      */
-    NodeApiService.prototype._removeRelation = function removeRelation(node, type, targets) {
+    NodeApiService.prototype.removeRelation = function removeRelation(node, type, targets) {
+        var self = this;
+
         if (!node || !type || !_.has(node, 'relations')) {
             return;
         }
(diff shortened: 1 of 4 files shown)


changev:
test this please

spec/lib/services/nodes-api-service-spec.js
```diff
@@ -499,13 +506,24 @@ describe("Http.Services.Api.Nodes", function () {
...
         });
 
-        it("_removeRelation should fail if don't have targets", function() {
+        it("_removeRelation should return undefined if don't have relations", function() {
+            var enclNode = {
+                    id: '1234abcd1234abcd1234abcd',
+                    type: 'enclosure'
+                };
+
+            expect(nodeApiService._removeRelation(enclNode, 'encloses', computeNode))
```
yyscamper: The function that begins with `_` is considered as private function, in unit-test, it better to focus on test those public function (or interfaces). The private function is purely the design internal.

This will leave us the flexibility to refactor the code in future (keep interface unmodified, but change implementation)

anhou: @yyscamper There's still debates for unit test for private methods. my opinion is no objection for this as long as the codes are qualified :)

iceiilin:
a little hard to understand all the details here. would prefer an oral explanation when available.

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

lib/services/nodes-api-service.js
```diff
@@ -84,6 +84,8 @@ function nodeApiServiceFactory(
      * @return {Object}    node after removing relation
      */
     NodeApiService.prototype._removeRelation = function removeRelation(node, type, targets) {
+        var self = this;
+
         if (!node || !type || !_.has(node, 'relations')) {
             return;
```
pengz1: If you want to align all return as promise, this return and next one should return Promise.resolve.

lib/services/nodes-api-service.js
```diff
@@ -117,85 +129,62 @@ function nodeApiServiceFactory(
         if (!(node && type && targets)) {
             return;
```
pengz1: Same as above, you'd better return a Promise as below code return a Promise.

lib/services/nodes-api-service.js
```diff
@@ -117,85 +129,62 @@ function nodeApiServiceFactory(
...
-        if (!node || !type || !_.has(node, 'relations')) {
-            return;
-        }
+            return modifiedNode;
+        })
+        .then(function(modifiedNode){
+            var targetsItems = _.map([].concat(targets), function(targetNode) {
+                targetNode = targetNode.id || targetNode;
+                if(targetNode === node.id ) {
+                    throw new Error('Node cannot have relationship '+type+' with itself');
```
iceiilin: may use promise.reject here. Same as below.

lib/services/nodes-api-service.js
```diff
@@ -84,6 +84,8 @@ function nodeApiServiceFactory(
      * @return {Object}    node after removing relation
      */
     NodeApiService.prototype._removeRelation = function removeRelation(node, type, targets) {
```
iceiilin: Since it is used in api/2.0/nodes, it might be better to remove the "_" prefix, which indicates a private function. Same as _addRelation

iceiilin: The function description could be adjusted accordingly to make it more clear.

pengz1:
+1, however my suggestion is that we should refine relation model, like OBM we should create a mongo model for relation for better and easy operations on it.

changev:
test this please

changev:
test this please

changev:
test this please

PR #660, opened 2017-05-26
client libs tags fix

BillyAbildgaard:
updating client library generation to reflect the view used for the tags property of a node

@RackHD/veyron @king-jam


BillyAbildgaard:
jenkins, test this please

johren:
There were two builds going for this PR at the same time so I aborted one of them to free the resources.  The failure above was that abort.  There is still a test running.

PR #672, opened 2017-06-14
RAC-4625: adding username complexity checking on account create

diontje:
Adding bare minimum account username complexity checking on account creation for both APIs.
As it stands now, accounts can be created with a single space, for example.
Change necessitates that usernames begin with a letter, followed by any number of letters, numbers, hyphens, periods, and underscores.


diontje:
test this please

PR #676, opened 2017-06-16
RAC-5241 Can't perform node rediscovery using post /workflows O

manfrednde:
This PR is to remove the graph length( 0 or 1) check( so that we have able to execute a graph that have a graph within it. 
The rediscovery graph length is 2 and it will fail with the check is place.

### DIFF
--- a/lib/services/workflow-api-service.js
+++ b/lib/services/workflow-api-service.js
@@ -110,6 +110,7 @@ function workflowApiServiceFactory(
             }
         })
         .then(function(graphs) {
+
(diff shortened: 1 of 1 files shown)


manfrednde:
@stuart-stanley 
@linlynn 
@anhou

panpan0000:
test this please

manfrednde:
@jimturnquist Please take a look

jimturnquist:
All good.  Coverage decrease is minimal.

PR #680, opened 2017-06-27
bring missed api "files/<name>" and "files/<name>/md5" back to southbound api

iceiilin:
Bring missed api "getting files/<name>" and "files/<name>/md5" back in southbound API. It is used in firmware update feature in skupack, where users could use workflow payload to tell the node fetching image file from files/ API.


lanchongyizu:
:+1:

PR #681, opened 2017-06-28
Fix the yml schema fot Ibms PUT

tannoa2:


tannoa2:
@RackHD/veyron

BillyAbildgaard:
Should `nodeId` be added to the schema as well?

tannoa2:
HI @BillyAbildgaard, 
The bodeId property is already part of the schema, but it was missing from the yml file. 
So when folks from Symphony coded against the codegen, they ran into issues.
https://github.com/RackHD/on-http/blob/master/static/schemas/ibms/ssh-ibm-service.json#L8

PR #682, opened 2017-06-28
Use the credentials of the microkernal as default value of the IBMs i…

tannoa2:
…f the config file is set correctly


tannoa2:
@RackHD/veyron

PR #687, opened 2017-07-07
OBM Settings Object Update

murilm:
The node Id was missing for the appropriate OBM Settings update call to work, this generates the correc object as consumed by the Monorail API

### DIFF
--- a/static/monorail-2.0.yaml
+++ b/static/monorail-2.0.yaml
@@ -211,9 +211,12 @@ definitions:
         type: object
       service:
         type: string
+      nodeId:
+        type: string
     required:
     - service
     - config
+    - nodeId
     type: object
   obm_led:
     properties:


murilm:
@BillyAbildgaard I tried it locally but when I went to create the Java Objects from it the name was "Id" and I needed it to be "nodeId" for the corresponding request... Not sure why would it recognize it as "nodeId" if I change the name to "Id"... There's other objects that make the reference as "nodeId" though... like here:

ssh-ibm-service_Ibm:
    description: SSH settings
    properties:
      nodeId:
        type: string
      config:
        properties:
          host:
            description: IP address
            type: string
          password:
            description: Password
            type: string
          user:
            description: Username
            type: string
        type: object
      service:
        type: string
    required:
    - service
    - config
    type: object

So why would I have to change it to just "Id" to make an integration test pass? O.o

king-jam:
This PR is going to be captured by: https://github.com/RackHD/on-http/pull/703

PR #689, opened 2017-07-11
RAC-4843

BrentHiggins:
* Added Redfish /Systems/{identifier}/Bios.ChangePassword support
* Added spec tests for /Systems/{identifier}/Bios.ChangePassword
* Added mktemp module for use in command
* updated redfish.yaml to support command


BrentHiggins:
@jimturnquist @derrickostertag

PR #694, opened 2017-07-14
RAC-5616 - Add redfish root endpoint for Redfish Spec compliance

hohene:
Adding support for the API /redfish endpoint.
This is to meet the Redfish Specification.


pathakvaidehi2391:
@jimturnquist @BrentHiggins

PR #697, opened 2017-07-17
clean up unit-test errors when it timeouts

lanchongyizu:
**Backgroud**
This is for the story https://rackhd.atlassian.net/browse/RAC-5639.

During work of https://rackhd.atlassian.net/browse/RAC-5593, we found many unit test cases won't do the clean-up work if running into failure (aka exception handler) , so the plan is to add clean-up work in unit test cases.

**DONE**
1) add helper.httpServerBefore and helper.httpServerAfter as common
methods to start and stop mock HTTP server.
2) fix the issue that some unit test file doesn't stop mock HTTP server
when starting mock server timeouts.
3) add global sandbox in beforeEach, and restore it in afterEach
4) remove unnecessary resetStubs and restoreStubs

**TEST**
It's tested in sandbox:
http://10.62.59.175:8080/job/on-core_jamie/

@RackHD/corecommitters

### DIFF
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -21,6 +21,27 @@ global.onHttpContext = index.onHttpContextFactory();
 // Legacy
 global.dihelper = onHttpContext.helper;
 
+helper.httpServerBefore = function(overrides, endpointOpt) {
+    before('helper.httpServer.before', function() {
+        this.timeout(10000);
+        return helper.startServer(overrides, endpointOpt);
+    });
+
+    beforeEach(function () {
+        this.sandbox = sinon.sandbox.create();
+    });
+
+    afterEach(function () {
+        this.sandbox.restore();
+    });
+};
+
+helper.httpServerAfter = function() {
+    after('helper.httpServer.after', function() {
+        return helper.stopServer();
+    });
+};
+
 helper.startServer = function (overrides, endpointOpt) {
     overrides = (overrides || []).concat([
         onHttpContext.helper.simpleWrapper({
@@ -64,7 +85,9 @@ helper.startServer = function (overrides, endpointOpt) {
 };
 
 helper.stopServer = function () {
-    return helper.injector.get('app').stop();
+    if (helper.injector) {
+        return helper.injector.get('app').stop();
+    }
 };
 
 helper.request = function (url, options) {
--- a/spec/lib/api/2.0/catalogs-spec.js
+++ b/spec/lib/api/2.0/catalogs-spec.js
@@ -9,32 +9,23 @@ describe('2.0 Http.Api.Catalogs', function () {
     var Errors;
     var stubNeedByIdentifier;
     var stubFind;
+    var waterline;
 
-    before('start HTTP server', function () {
-        this.timeout(5000);
-        return helper.startServer([]).then(function () {
-            Promise = helper.injector.get('Promise');
-            Errors = helper.injector.get('Errors');
-            var w = helper.injector.get('Services.Waterline');
-            stubNeedByIdentifier = sinon.stub(w.catalogs, "needByIdentifier");
-            stubFind = sinon.stub(w.catalogs, "find");
-        });
-    });
+    helper.httpServerBefore();
 
-    beforeEach('loading views', function(){
+    before(function () {
+        Promise = helper.injector.get('Promise');
+        Errors = helper.injector.get('Errors');
+        waterline = helper.injector.get('Services.Waterline');
         return helper.injector.get('Views').load();
     });
 
-    afterEach("reset stubs", function() {
-        stubFind.reset();
-        stubNeedByIdentifier.reset();
+    beforeEach('set up mocks', function() {
+        stubNeedByIdentifier = this.sandbox.stub(waterline.catalogs, "needByIdentifier");
+        stubFind = this.sandbox.stub(waterline.catalogs, "find");
     });
 
-    after('stop HTTP server', function () {
-        stubFind.restore();
-        stubNeedByIdentifier.restore();
-        return helper.stopServer();
-    });
+    helper.httpServerAfter();
 
     describe("GET /catalogs", function() {
 
--- a/spec/lib/api/2.0/config-spec.js
+++ b/spec/lib/api/2.0/config-spec.js
@@ -5,25 +5,19 @@
 
 describe('Http.Api.Config v2.0', function () {
     var configuration;
-    before('start HTTP server', function () {
-        this.timeout(10000);
-        return helper.startServer().then(function() {
-            configuration = helper.injector.get('Services.Configuration');
-            sinon.stub(configuration, 'set').returns(configuration);
-            sinon.stub(configuration, 'getAll').returns({});
-        });
+
+    helper.httpServerBefore();
+
+    before(function () {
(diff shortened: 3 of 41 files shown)


spec/helper.js
```diff
@@ -21,6 +21,27 @@ global.onHttpContext = index.onHttpContextFactory();
 // Legacy
 global.dihelper = onHttpContext.helper;
 
+helper.httpServerBefore = function(overrides, endpointOpt) {
+    before('helper.httpServer.before', function() {
+        this.timeout(5000);
+        return helper.startServer(overrides, endpointOpt);
+    });
+
+    beforeEach(function () {
```
yyscamper: Since on-core helper.js  has already create the sandbox, my suggestion is directly calling on-core's `before` (https://github.com/RackHD/on-core/blob/master/spec/helper.js#L268) rather than writing some duplicated code.

spec/helper.js
```diff
@@ -21,6 +21,27 @@ global.onHttpContext = index.onHttpContextFactory();
...
+        this.sandbox = sinon.sandbox.create();
+    });
+
+    afterEach(function () {
+        this.sandbox.restore();
+    });
+};
+
+helper.httpServerAfter = function() {
+    after('helper.httpServer.after', function() {
```
yyscamper: How about moving this to `httpServerBefore`, so that the unit-test doesn't need to call the `httpServerAfter`

spec/lib/api/2.0/config-spec.js
```diff
@@ -5,26 +5,20 @@
...
-            configuration = helper.injector.get('Services.Configuration');
-            sinon.stub(configuration, 'set').returns(configuration);
-            sinon.stub(configuration, 'getAll').returns({});
-        });
-    });
 
-    afterEach('tear down mocks', function () {
-        configuration.set.reset();
-        configuration.getAll.reset();
+    helper.httpServerBefore([]);
```
yyscamper: It would be better to remove the argument `[]`, since an empty array seems redundant

spec/helper.js
lanchongyizu: I think your idea is calling
```
helper.before();
helper.httpServerBefore();
```
instead of
```
helper.httpServerBefore();
```
But we don't need `helper.start`  in `on-http`, so none of the unit test files calls `helper.before`.

spec/helper.js
lanchongyizu: It doesn't make sense here if we want to insert an `after` block before `helper.httpServer.after`.
E.g.:
```
after(function() {
//call something
});
helper.httpServerAfter();
```

spec/lib/api/2.0/ibms-spec.js
```diff
@@ -65,26 +65,15 @@ describe('Http.Api.Ibms', function () {
...
+    helper.httpServerBefore();
 
-    afterEach(function () {
-        if (stub) {
-            stub.restore();
-            stub = undefined;
-        }
+    before(function () {
+        waterline = helper.injector.get('Services.Waterline');
+        Errors = helper.injector.get('Errors');helper.injector.get("Services.Configuration");
```
anhou: what's the purpose of 'helper.injector.get("Services.Configuration");'?

anhou:
@lanchongyizu  Could you help post the errors in the RAC or in this PR's message? I see lots of minor changes which is not related to exception handler.

spec/helper.js
```diff
@@ -21,6 +21,27 @@ global.onHttpContext = index.onHttpContextFactory();
 // Legacy
 global.dihelper = onHttpContext.helper;
 
+helper.httpServerBefore = function(overrides, endpointOpt) {
+    before('helper.httpServer.before', function() {
+        this.timeout(5000);
```
anhou: I see that some cases are setting timeout to 10000ms, if all are set to 5000ms, is it sure that it doesn't have random issues?

lanchongyizu: Previously, they might include other tasks, so it was set to 10000ms, but here we just start HTTP server, so it is all set to 5000ms.

spec/lib/api/2.0/ibms-spec.js
lanchongyizu: Good finding, it's not my purpose here. It's from the original code, and I think we should delete it.

lanchongyizu:
The main errors are in SH sandbox Jenkins, such as http://10.62.59.175:8080/blue/organizations/jenkins/on-core_jamie/detail/on-core_jamie/29/tests

spec/helper.js
lanchongyizu: After group discussion, it will be set to 10000ms.

spec/lib/api/login/login-spec.js
```diff
@@ -81,7 +81,7 @@ describe('Http.Api.Login', function () {
     before('setup default admin user', function() {
         waterline = helper.injector.get('Services.Waterline');
         waterline.localusers = {
-            findOne: sinon.stub()
+            findOne: sandbox.stub()
```
iceiilin: Do we need to use "this.sandbox" here to align with other files?

spec/lib/api/redfish-1.0/event-service-spec.js
```diff
@@ -39,62 +39,41 @@ describe('Redfish Event Service', function () {
             }
         }]
     };
-    
-    before('start HTTP server', function () {
-        this.timeout(5000);
+
+    before(function() {
```
iceiilin: Can this before be combined with this one https://github.com/RackHD/on-http/pull/697/files#diff-cf0f2855402ebeda8abc6126044353caR54 ?

spec/lib/api/2.0/catalogs-spec.js
```diff
@@ -9,32 +9,23 @@ describe('2.0 Http.Api.Catalogs', function () {
...
-    });
+    helper.httpServerBefore();
 
-    beforeEach('loading views', function(){
+    before(function () {
+        Promise = helper.injector.get('Promise');
+        Errors = helper.injector.get('Errors');
+        waterline = helper.injector.get('Services.Waterline');
         return helper.injector.get('Views').load();
     });
```
lanchongyizu: No, they are called one by one.

spec/helper.js
```diff
@@ -64,7 +85,9 @@ helper.startServer = function (overrides, endpointOpt) {
 };
 
 helper.stopServer = function () {
-    return helper.injector.get('app').stop();
+    if (helper.injector) {
```
lanchongyizu: We could imagine that some unhandled exceptions are thrown before https://github.com/RackHD/on-core/blob/master/spec/helper.js#L163, then mocha will go on with the `after` section, then helper.injector would be undefined.

spec/lib/api/login/login-spec.js
lanchongyizu: For this specific file, it defines its own sandbox, so we keep it here. I agree with you that it's better to align with other files, but it needs more work.

spec/lib/api/redfish-1.0/event-service-spec.js
lanchongyizu: Originally, this is to stub `redfishTool` before httpServer is started, so we can't combine these two `before` section.

PR #709, opened 2017-07-27
[RAC5316]Travis CI: failed to build debian packages

PengTian0:
Fix RAC-5316: Travis CI: failed to build debian packages
Error message:
`+debuild --no-lintian --no-tgz-check -us -uc
debuild: fatal error at line 948:
problem running fakeroot: either install the fakeroot package,
use a -r option to select another root command program to use or
run me as root!`
The PR: install fakeroot for debuild to fix the issue.

@anhou @iceiilin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
       - dh-make
       - devscripts
       - debhelper
+      - fakeroot
       - git
       - default-jdk
       - maven


pengz1:
+1

anhou:
it looks like on-imagebuilder also has the same problem, maybe it also needs to be updated

PR #712, opened 2017-07-28
waterline validation error when driveIds.length==0

wangzhizheng:
we found there will be an waterline validation error when driveIds.length==0 as below and task will fail
Error (E_VALIDATION) :: 1 attribute is invalid
            at WLValidationError.WLError (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/waterline/lib/waterline/error/WLError.js:26:15)

So we need to add something to driveIds to avoid the validation error, empty json {} makes sense.


iceiilin:
Sorry for more comments. After discussing with other guys, although the variable looks a little strange, we would tend to leave it as an array as it was. so
https://github.com/RackHD/on-http/pull/712/files#diff-ef9ef77b07c59967b91d8afd5e18c386R280
could be
```
driveId=[{}]
```
as you modified originally.
and 
https://github.com/RackHD/on-http/blob/master/spec/data/templates/get_driveid-spec.js#L167
need to be modified as
```
[{}]
```
to pass the unit test.
Thanks for your efforts here!

PR #713, opened 2017-07-28
RAC-5253 delete install web ui from zip

changev:
delete the web ui installation  from zip file.

We support 3 ways to install web-ui

1. apt-get install, work with apt-get install rackhd.
2. docker images, work with docker-compose up rackhd
3. if users installed rackhd with src, they should clone web-ui src and build statics manually.


changev:
test this please

panpan0000:
Can you please elaborate the way #3 ? 
if people run RackHD from src code, and do npm install.
which document can people follow to build/copy on-web-ui.

changev:
@panpan0000 

I'll add some docs in on-http README!

panpan0000:
As what we talk, the concern is 
if people using Node4 to install rackhd from src, but you request them to build on-web-ui by themselves, which requires Node6...

Let's keep -1 👎   until this gets clarified.
@iceiilin

PR #715, opened 2017-07-28
RAC-5581 : Redfish/Managers/NetworkProtocol for rackhd manager,dell a…

pathakvaidehi2391:
…nd non-dell node.


pathakvaidehi2391:
@jimturnquist @patelb10

PR #716, opened 2017-07-28
RAC-5686 - Normalization of system-spec.js

jimturnquist:
* Dell vs non-dell is not handled quite right. The wsman.isDellSystem call
  should be allowed to succeed or fail based on the result of getNodeById.
* Default resolve statements mask various blocks of code.
* Add some consistency to the use of node ids. Reference object ids
  when possible instead of writing duplicate ids.


jimturnquist:
Closing this PR.  Will reconstitute after other redfish-related changes have gone in.

PR #717, opened 2017-07-31
RAC-5752 - Support for TaskCollection compliance changes

larry-dean:
* Remove embedded tasks object


larry-dean:
@jimturnquist @hohene

PR #718, opened 2017-07-31
RAC-5686 Added @odata.id entry in /redfish/v1/Managers response

patelb10:


patelb10:
@pathakvaidehi2391 @jimturnquist @hohene

PR #724, opened 2017-08-03
Add .npmignore file for npm publish

mcgG:
**/static/files** is ignored in **.gitignore** and we didn’t create **.npmignore** file before, so when using npm publish, **.gitignore** will be converted to **.npmignore** automatically , so the **/static/files** folder is ignored. 
Adding this .npmignore can solve the problem


iceiilin:
add to whitelist

iceiilin:
test this please

PR #731, opened 2017-08-10
rac-5884 - Continuation of convertion systems.js to use wsman.isDell system(identifier)

larry-dean:
rac-5926 - increase unit testing = 90%+
* All calls now have test for both DELL and non-Dell devices as appropriate
* Corrected error in logic as found
* Added or increases mock catalogs


larry-dean:
test this please

stuart-stanley:
@gpaulos or @derrickostertag: you were both paired with Larry for this. Can you bless it please?

@larry-dean : please squash the commits

PR #737, opened 2017-08-16
RAC-5799 - Pv4, IPv6 system IPs are not populated

larry-dean:
* Correct issues with dictionaries that use null as a value
* Do not display LinkStatus if value is null
* Add test to validate


larry-dean:
@stuart-stanley @BrentHiggins  @diontje

hohene:
test this please

PR #739, opened 2017-08-17
RAC-5001 Change WWID to SATA name for ESXi SATA drive ID

nortonluo:
## Background
The ESXi use their own method to generate a identity for the hard drives they found in the system. We use this script run microkernel to gernerate the same ID using the same algorithm as vmware. However in some latest server, we found some SSD's ESXi ID is wrong in RackHD.  It is because RackHD will use WWID to overide the SATA name if it is available. But in fact, ESXi use SATA name to identify this kind of SATA drive.
## Details
Revert the overide process. Use SATA name as ESXi  ID if drive have both SATA name and WWID.

## Result
Validated PASS on EMC platforms.

@pengz1 @iceiilin

### DIFF
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -85,7 +85,7 @@ function parseDriveWwid(idList) {
 
     //According to SCSI-3 spec, vendor specified logic unit name string is 60
 	//Only IDE and SCSI disk will be retrieved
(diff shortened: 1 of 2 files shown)


iceiilin:
Will this change impact secure erase or other features (though I doesn't find others) @pengz1 ?

iceiilin:
Also @nortonluo I suggest to squash your commit into one if you don't view it necessary to keep them.

PR #741, opened 2017-08-18
 Add redfish /Systems/identifier/Processors API for UCS

pengz1:
Background
To enable redfish API data access with UCS.

Details
1. Implemented details to identify Dell/Cisco and other hardware platforms
2. Implement /Systems/identifier/Processors API for UCS

Paired with: @bbcyyb

@anhou @iceiilin @yaolingling @mcgG @larry-dean

### DIFF
--- a/data/views/redfish-1.0/ucs.1.0.0.processorcollection.json
+++ b/data/views/redfish-1.0/ucs.1.0.0.processorcollection.json
@@ -1,0 +1,16 @@
+{
+    "@odata.context" : "<%= basepath %>/$metadata#Systems/Links/Members/<%= identifier %>/Processors/#entity",
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#Procssors.1.0.0.ProcessorsCollection",
+    "Oem" : {},
+    "Name": "Processors Collection",
+    "Members@odata.count": <%= processors.data.num_of_cpus %>,
+    "Members": [
+        <% for(var i = 0; i < processors.data.num_of_cpus; i += 1) { %>
+            {
+                "@odata.id": "<%= basepath %>/Systems/<%=identifier%>/Processors/<%=i%>"
+            }
+        <%= ( i + 1 < processors.data.num_of_cpus ) ? ',': '' %>
+        <% }; %>
+    ]
+}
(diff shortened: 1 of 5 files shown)


data/views/redfish-1.0/ucs.1.0.0.processorcollection.json
```diff
@@ -0,0 +1,16 @@
+{
+    "@odata.context" : "<%= basepath %>/$metadata#Systems/Links/Members/<%= identifier %>/Processors/#entity",
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#Procssors.1.0.0.ProcessorsCollection",
```
iceiilin: "Procssors" typo. 
One question: why are these fields different from both https://github.com/RackHD/on-http/blob/master/data/views/redfish-1.0/wsman.1.0.0.processorcollection.json and https://github.com/RackHD/on-http/blob/master/data/views/redfish-1.0/redfish.1.0.0.processorcollection.json ?

lib/services/redfish-validator-service.js
```diff
@@ -208,5 +212,40 @@ function redfishValidatorFactory(
...
+        _.forEach(_identifiers, function(_identifier){
+            var items = _identifier.split(':');
+            if (dellFormat.test(_identifier)){
+                vendor = 'Dell';
+                return false;
+            } else if (validator.isIP(items[0]) && ucsDnFormat.test(items[1])) {
+                vendor = 'Cisco';
+                return false;
+            }
+        });
```
iceiilin: What about making a structure like
```
{
    "Dell": /^[0-9|A-Z]{7}$/,
    "Cisco": /^(sys|org-root)(\/[\w-]+)+$/
}
```
thus you just go through identifies loop without any "if... else" case and it would be convenient to add more vendor in the future.

lib/services/redfish-validator-service.js
```diff
@@ -208,5 +212,40 @@ function redfishValidatorFactory(
...
+        return vendor;
+    }
+
```
iceiilin: It might be a little confuse to add these common functions in redfish validator. What about changing the name of current service, or move these newly-added into a new service?

pengz1: @iceiilin 
We need another PR to fix this, either to change this redfish-validator-service.js to redfish-api-service.js, or we need split this document into two parts.

pengz1:
test this please

lib/services/redfish-validator-service.js
iceiilin: agree

PR #742, opened 2017-08-21
[RAC-5094] Fix Redfish Api : /Systems/identifier

yaolingling:
**Background**
   Users cannot get the specific identifier systems information. So, the problem need to be fixed.

**Details**
1. To solve this problem, on-http repo code needs  to be changed.
2. Related files: 
    on-http/lib/api/redfish-1.0/systems.js
    on-http/spec/lib/api/redfish-1.0/systems-spec.js
    on-http/lib/services/redfish-validator-service.js
3. Validate whether the api is fixed successfully.
Execute the command:
     curl localhost:8080/redfish/v1/Systems/5994f71bd0e1e80257af9ff9 | jq '.'

**Results**
1) After the command is executed, you can get the result.
image
image

2) Unit test result
execute the command in on-http directory: "npm test"
image

3) Coverage:
image

**Reviewers**
@iceiilin  @anhou @pengz1 @mcgG

### DIFF
--- a/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
+++ b/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
@@ -1,0 +1,74 @@
+{
+    "@odata.context" : "<%= basepath %>/$metadata#ComputerSystem.ComputerSystem",
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#ComputerSystem.v1_3_0.ComputerSystem",
+    "Oem" : {},
+    "Id": "<%= identifier %>",
+    "Description": "<%= data.descr %>",
+    "Name": "Computer System",
+    "SystemType": "<%= systemType %>",
+    "AssetTag": "",
+    "Manufacturer": "<%= data.vendor %>",
+    "Model": "<%= data.model %>",
+    "SKU": "<%= skuInfo %>",
+    "SerialNumber": "<%= data.serial %>",
+    "PartNumber": "<%= data.part_number %>",
+    "UUID": "<%= data.uuid %>",
+    "HostName": "",
+    "IndicatorLED": null,
+    "PowerState": null,
+    "Boot": {},
+    "BiosVersion": "",
+    "ProcessorSummary": {
+        "Count": <%= data.num_of_cpus %>,
+        "Model": "",
+        "Status": {}
+    },
+    "MemorySummary": {
+        "TotalSystemMemoryGiB": <%= data.total_memory %>,
+        "Status": {}
+    },
+    "Actions": {
+        "Oem": {
+            "RackHD": {
+                "#RackHD.BootImage": {
+                    "target": "<%=basepath%>/Systems/<%=identifier%>/Actions/RackHD.BootImage"
+                }
+            }
+        },
+        "#ComputerSystem.Reset": {
+            "target": "<%=basepath%>/Systems/<%=identifier%>/Actions/ComputerSystem.Reset"
+        }
+    },
+    "Status": {},    
+    "Processors": {
+        "@odata.id": "<%= basepath %>/Systems/<%= identifier %>/Processors"
+    },
+    "EthernetInterfaces": {
+        "@odata.id": "<%= basepath %>/Systems/<%= identifier %>/EthernetInterfaces"
+    },
+    "SimpleStorage": {
+        "@odata.id": "<%= basepath %>/Systems/<%= identifier %>/SimpleStorage"
+    },
+    "LogServices": {
+        "@odata.id": "<%= basepath %>/Systems/<%= identifier %>/LogServices"
+    },
+    "Links": {
+        "Oem": {},
+        "ManagedBy@odata.count": 0,
+        "ManagedBy": [
+            {}
+        ],
+        "PoweredBy@odata.count": 0,
+        "PoweredBy": [],
+        "CooledBy@odata.count": 0,
+        "CooledBy": [],
+        "Chassis@odata.count": 0,
+        "Chassis": [
+            {}
+        ]
+    },
+    "Storage":{
+        "@odata.id": "<%= basepath %>/Systems/<%= identifier %>/Storage"
+    }
+}
--- a/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
+++ b/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
@@ -15,7 +15,7 @@
     "PartNumber": "<%= hardware.data.part_number %>",
     "UUID": "<%= hardware.data.uuid %>",
     "HostName": "",
-    "IndicatorLED": "<%= led.data.oper_state === 'on'?'On':'Off' %>",
+    "IndicatorLED": "<%= led.data.oper_state === 'on'?'Lit':'Off' %>",
     "PowerState": "<%= hardware.data.oper_power.charAt(0).toUpperCase() + hardware.data.oper_power.slice(1) %>",
     "Boot": {},
     "BiosVersion": "<%= bios.data.revision %>",
--- a/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
+++ b/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
@@ -4,28 +4,28 @@
     "@odata.type": "#ComputerSystem.v1_3_0.ComputerSystem",
     "Oem" : {},
     "Id": "<%= identifier %>",
-    "Description": "<%= data.descr %>",
+    "Description": "<%= hardware.data.descr %>",
     "Name": "Computer System",
     "SystemType": "<%= systemType %>",
     "AssetTag": "",
-    "Manufacturer": "<%= data.vendor %>",
-    "Model": "<%= data.model %>",
+    "Manufacturer": "<%= hardware.data.vendor %>",
+    "Model": "<%= hardware.data.model %>",
     "SKU": "<%= skuInfo %>",
-    "SerialNumber": "<%= data.serial %>",
-    "PartNumber": "<%= data.part_number %>",
-    "UUID": "<%= data.uuid %>",
+    "SerialNumber": "<%= hardware.data.serial %>",
+    "PartNumber": "<%= hardware.data.part_number %>",
+    "UUID": "<%= hardware.data.uuid %>",
     "HostName": "",
-    "IndicatorLED": null,
-    "PowerState": null,
+    "IndicatorLED": "<%= led.data.oper_state.charAt(0).toUpperCase() +  led.data.oper_state.slice(1) %>",
+    "PowerState": "<%= hardware.data.oper_power.charAt(0).toUpperCase() + hardware.data.oper_power.slice(1) %>",
     "Boot": {},
-    "BiosVersion": "",
+    "BiosVersion": "<%= bios.data.revision %>",
     "ProcessorSummary": {
-        "Count": <%= data.num_of_cpus %>,
+        "Count": <%= hardware.data.num_of_cpus %>,
         "Model": "",
         "Status": {}
     },
     "MemorySummary": {
-        "TotalSystemMemoryGiB": <%= data.total_memory %>,
+        "TotalSystemMemoryGiB": <%= hardware.data.total_memory/1024 %>,
         "Status": {}
     },
     "Actions": {
@@ -55,17 +55,27 @@
     },
     "Links": {
         "Oem": {},
-        "ManagedBy@odata.count": 0,
+        "ManagedBy@odata.count": <%= obm.length %>,
         "ManagedBy": [
-            {}
+            <% obm.forEach(function(obm, i, arr) { %>
+                {
+                    "@odata.id": "<%= basepath %>/Managers/<%= obm %>"
+                }
+                <%= ( arr.length > 0 && i < arr.length-1 ) ? ',': '' %>
+            <%  }); %>
         ],
         "PoweredBy@odata.count": 0,
         "PoweredBy": [],
         "CooledBy@odata.count": 0,
         "CooledBy": [],
-        "Chassis@odata.count": 0,
+        "Chassis@odata.count": <%= chassis.length %>,
         "Chassis": [
-            {}
+            <% chassis.forEach(function(chassis, i, arr) { %>
+                {
+                    "@odata.id": "<%= basepath %>/Chassis/<%= chassis %>"
+                }
+                <%= ( arr.length > 0 && i < arr.length-1 ) ? ',': '' %>    
+            <% }); %>
         ]
     },
     "Storage":{
--- a/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
+++ b/data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
@@ -15,7 +15,7 @@
     "PartNumber": "<%= hardware.data.part_number %>",
     "UUID": "<%= hardware.data.uuid %>",
     "HostName": "",
-    "IndicatorLED": "<%= led.data.oper_state.charAt(0).toUpperCase() +  led.data.oper_state.slice(1) %>",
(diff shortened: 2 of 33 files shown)


data/views/redfish-1.0/redfish.2016.3.computersystem.1.4.0.json
```diff
@@ -0,0 +1,74 @@
...
+        "CooledBy": [],
+        "Chassis@odata.count": 0,
+        "Chassis": [
+            {}
+        ]
+    },
+    "Storage":{
+        "@odata.id": "<%= basepath %>/Systems/<%= identifier %>/Storage"
+    }
+}
```
pengz1: Don't change the version, it is no controlled by us.
Use name like:
ucs.1.0.0.computersystem.x.x.x.json

lib/api/redfish-1.0/systems.js
```diff
@@ -47,6 +47,8 @@ var dataFactory = function(identifier, dataName) {
             });
         case 'Redfish':
             return waterline.catalogs.find({"node": identifier, "source":dataName });
+        case 'UCS':
+            return waterline.catalogs.findOne({"node": identifier, "source":dataName });
```
pengz1: You should use nodeApi.getNodeCatalogSourceById(identifier, dataName);

lib/api/redfish-1.0/systems.js
```diff
@@ -280,7 +282,13 @@ var getSystem = controller(function(req, res) {
             }).catch(function(error) {
                 return redfish.handleError(error, res);
             });
-        } else {
+        } else if(result.vendor === 'Cisco'){
+	     return dataFactory(identifier,'UCS').then(function(data){
```
pengz1: You should get data from UCS:blade-x for blade servers and UCS for rack mount servers.

lib/api/redfish-1.0/systems.js
```diff
@@ -280,7 +282,13 @@ var getSystem = controller(function(req, res) {
...
+		 return redfish.render('redfish.2016.3.computersystem.1.4.0.json',
```
pengz1: Update name redfish.2016.3.computersystem.1.4.0.json

spec/lib/api/redfish-1.0/system-spec.js
```diff
@@ -771,6 +783,10 @@ describe('Redfish Systems Root', function () {
     });
 
     it('should return a valid system', function() {
```
pengz1: A suggestion for you, use describe to group all test for your API.
We have two many test in this file and should divide them into small groups.

data/views/redfish-1.0/redfish.2016.3.computersystem.1.4.0.json
yaolingling: ok. I rename the file to "ucs.1.0.0.computersystem.1.0.0.json"

lib/api/redfish-1.0/systems.js
yaolingling: ok

data/views/redfish-1.0/redfish.2016.3.computersystem.1.4.0.json
```diff
@@ -0,0 +1,74 @@
...
+    "@odata.type": "#ComputerSystem.v1_3_0.ComputerSystem",
+    "Oem" : {},
+    "Id": "<%= identifier %>",
+    "Description": "<%= data.descr %>",
+    "Name": "Computer System",
+    "SystemType": "<%= systemType %>",
+    "AssetTag": "",
+    "Manufacturer": "",
+    "Model": "<%= data.model %>",
+    "SKU": "",
```
yaolingling: Node does not have the attribute:sku

lib/services/redfish-validator-service.js
```diff
@@ -207,6 +211,34 @@ function redfishValidatorFactory(
         return Promise.reject(new Errors.NotFoundError ('Message registry not found'));
 
     };
+    RedfishValidator.prototype.getVendorNameById = function(identifier) {
```
yyscamper: It is not suitable to put this function in `RedfishValidator`. `RedfishValidator` is a module which designed manages the redfish json schema validation. Actually this function has nothing relationship with schema and validation.  So please move it out.

lib/services/redfish-validator-service.js
```diff
@@ -207,6 +211,34 @@ function redfishValidatorFactory(
...
+        var result = {node: undefined, vendor: undefined};
+        return waterline.nodes.getNodeById(identifier)
+        .then(function(node){
+	    if(!node){
```
yyscamper: Please fix all the indent problems

lib/services/redfish-validator-service.js
```diff
@@ -207,6 +211,34 @@ function redfishValidatorFactory(
...
+	        throw new Errors.NotFoundError('invalid node id.');
+	    }
+	    result.node = node;
+	    result.vendor = _getSystemVendorByIdentifiers(node.identifiers);
```
yyscamper: You don't have to declare a variable in advance, and also use `needOne` rather than `getNodeById` will save a throw statement. For simplicity, you could just do as following:
```javascript
RedfishValidator.prototype.getVendorNameById = function(nodeId) {
    return waterline.nodes.needOne({id: nodeId})
        .then(function(node){
	        return {
                node: node;
	            vendor: _getSystemVendorByIdentifiers(node.identifiers);
            }
	});
};
```

spec/lib/services/redfish-validator-service-spec.js
```diff
@@ -63,5 +69,79 @@ describe("Redfish Validator Service", function() {
...
+                ]
+            };
+
+            waterline.nodes.getNodeById.resolves(sampleDataInfo);
+            return redfish.getVendorNameById(id)
+                .then(function(result){
+                    expect(result.vendor).to.equal("Cisco");
+                    expect(result.node).to.deep.equal(sampleDataInfo);
+                    expect(waterline.nodes.getNodeById).to.be.calledOnce;
+                    expect(waterline.nodes.getNodeById).to.be
```
yyscamper: I would suggest you don't make assertion on the `waterline.nodes.getNodeById` as the function `getVendorNameById` is designed to get a vendor name by parsing the node id, but it doesn't need to depend on the waterline to do such parsing. For example, in future, we may can get the vendor name directly from the SKU, after then, we don't need to change unit-test

yyscamper: And then, after remove the assertion on `getNodeById`,  this unit-test can be change to be as simple as following:
```javascript
it('should get Cisco vendor name by identifier', function() {
            var id = "599337d6ff99ed24305bc58a";
            var sampleDataInfo = {
                "id": id,
                "identifiers": [
                    "1.1.1.0:sys/rack-unit-2"
                ]
            };

            waterline.nodes.getNodeById.WithArgs(id).resolves(sampleDataInfo);

            return expect(redfish.getVendorNameById(id)).to.become({
                vendor: "Cisco",
                node: sampleDataInfo
             });
        });
```

lib/services/redfish-validator-service.js
yaolingling: Thanks Felix. I will move the function to a new defined js file.

lib/services/redfish-validator-service.js
yaolingling: ok

iceiilin:
add to whitelist

spec/lib/api/redfish-1.0/system-spec.js
```diff
@@ -771,6 +783,10 @@ describe('Redfish Systems Root', function () {
...
+        redfish.getVendorNameById.resolves({
+            node: node,
+            vendor: 'notDellAndCisco'
```
yaolingling: It's purpose is to get the results of the third system information which is not dell and Cisco

iceiilin:
test this please

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
```diff
@@ -0,0 +1,74 @@
...
+    "SystemType": "<%= systemType %>",
+    "AssetTag": "",
+    "Manufacturer": "<%= data.vendor %>",
+    "Model": "<%= data.model %>",
+    "SKU": "<%= skuInfo %>",
+    "SerialNumber": "<%= data.serial %>",
+    "PartNumber": "<%= data.part_number %>",
+    "UUID": "<%= data.uuid %>",
+    "HostName": "",
+    "IndicatorLED": null,
```
pengz1: You can use UCS:locator-led.oper_state

spec/lib/api/redfish-1.0/system-spec.js
iceiilin: Suggest to change it to undefined (as the mechanism in https://github.com/RackHD/on-http/pull/742/files#diff-3ef377bee23e7df190a0c8a2a9be21f6R215), or "other" is better than "notDellAndCisco", in case other vendor will be added into this function, when "notDellAndCisco" could not cover all "other" cases.

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
```diff
@@ -0,0 +1,74 @@
+{
+    "@odata.context" : "<%= basepath %>/$metadata#ComputerSystem.ComputerSystem",
+    "@odata.id": "<%= url %>",
+    "@odata.type": "#ComputerSystem.v1_3_0.ComputerSystem",
```
yaolingling: ok

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
```diff
@@ -0,0 +1,74 @@
...
+    "AssetTag": "",
+    "Manufacturer": "<%= data.vendor %>",
+    "Model": "<%= data.model %>",
+    "SKU": "<%= skuInfo %>",
+    "SerialNumber": "<%= data.serial %>",
+    "PartNumber": "<%= data.part_number %>",
+    "UUID": "<%= data.uuid %>",
+    "HostName": "",
+    "IndicatorLED": null,
+    "PowerState": null,
```
yaolingling: ok

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
yaolingling: ok

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
```diff
@@ -0,0 +1,74 @@
...
+    "Name": "Computer System",
+    "SystemType": "<%= systemType %>",
+    "AssetTag": "",
+    "Manufacturer": "<%= data.vendor %>",
+    "Model": "<%= data.model %>",
+    "SKU": "<%= skuInfo %>",
+    "SerialNumber": "<%= data.serial %>",
+    "PartNumber": "<%= data.part_number %>",
+    "UUID": "<%= data.uuid %>",
+    "HostName": "",
```
yaolingling: ???

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
```diff
@@ -0,0 +1,74 @@
...
+    "IndicatorLED": null,
+    "PowerState": null,
+    "Boot": {},
+    "BiosVersion": "",
```
yaolingling: ok

spec/lib/api/redfish-1.0/system-spec.js
yaolingling: ok

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
```diff
@@ -0,0 +1,74 @@
...
+    },
+    "LogServices": {
+        "@odata.id": "<%= basepath %>/Systems/<%= identifier %>/LogServices"
+    },
+    "Links": {
+        "Oem": {},
+        "ManagedBy@odata.count": 0,
+        "ManagedBy": [
+            {}
+        ],
```
yaolingling: ok

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.0.0.json
```diff
@@ -0,0 +1,74 @@
...
+        "PoweredBy@odata.count": 0,
+        "PoweredBy": [],
+        "CooledBy@odata.count": 0,
+        "CooledBy": [],
+        "Chassis@odata.count": 0,
+        "Chassis": [
+            {}
```
yaolingling: ok

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
```diff
@@ -0,0 +1,84 @@
...
+    "SystemType": "<%= systemType %>",
+    "AssetTag": "",
+    "Manufacturer": "<%= hardware.data.vendor %>",
+    "Model": "<%= hardware.data.model %>",
+    "SKU": "<%= skuInfo %>",
+    "SerialNumber": "<%= hardware.data.serial %>",
+    "PartNumber": "<%= hardware.data.part_number %>",
+    "UUID": "<%= hardware.data.uuid %>",
+    "HostName": "",
+    "IndicatorLED": "<%= led.data.oper_state.charAt(0).toUpperCase() +  led.data.oper_state.slice(1) %>",
```
pengz1: On should not work here. You should change On to "Lit"

data/views/redfish-1.0/ucs.1.0.0.computersystem.1.3.0.json
```diff
@@ -0,0 +1,84 @@
...
+    "PowerState": "<%= hardware.data.oper_power.charAt(0).toUpperCase() + hardware.data.oper_power.slice(1) %>",
+    "Boot": {},
+    "BiosVersion": "<%= bios.data.revision %>",
+    "ProcessorSummary": {
+        "Count": <%= hardware.data.num_of_cpus %>,
+        "Model": "",
+        "Status": {}
+    },
+    "MemorySummary": {
+        "TotalSystemMemoryGiB": <%= hardware.data.total_memory/1024 %>,
```
pengz1: Minor suggestion:
use >> 10 instead of /1024

yaolingling:
test this please

lib/api/redfish-1.0/systems.js
```diff
@@ -283,12 +285,32 @@ var getSystem = controller(function(req, res) {
...
+                bios: dataFactory(identifier,'UCS:bios'),
+                led: dataFactory(identifier,'UCS:locator-led'),
+                skuInfo: Promise.try(function(){
+                            return node.sku === undefined ? "": node.sku;
+                        }),
+                chassis: Promise.map(node.relations,function(val){
+                            if(val.relationType === 'enclosedBy'){
+                                return val.targets;
+                            }
+                        }),
```
iceiilin: If node has relations other than enclosedBy, what the Promise.map() will return?  I think the "chassis" item will be something like [nodeid, undefined,...]. Will it impact the following render process?

spec/lib/api/redfish-1.0/system-spec.js
```diff
@@ -866,140 +869,159 @@ describe('Redfish Systems Root', function () {
...
-            node: node.id,
-            source: 'dummysource',
-            data: catalogData
-        }));
+        it('should return a valid system', function() {
+            redfish.getVendorNameById.resolves({
+                node: node,
+                vendor: 'undefined'
+            });
+            waterline.catalogs.findLatestCatalogOfSource.resolves(Promise.resolve({
```
iceiilin: Using the object directly as the parameter for "resolves" is enough. Promise.resolve is not necessary here. Same as the following case. Just a suggestion for further work.

lib/api/redfish-1.0/systems.js
yaolingling: I have tested it. It will not impact rendor process

PR #745, opened 2017-08-22
Fix: failed to run unit test as user root

PengTian0:
Failed to run script "HWIMO-TEST"  as user root.
Cause: 
```npm install``` does not run scripts specified in package.json files as user root.
That causes the failure of unit test.
Because the file may be run in docker, we need to fix the issue.

Solution:
1. Run ```npm install --unsate-perm```
2. Run the scripts specified in package.json files after ```npm install```. That brings duplicate between package.json and HWIMO-TEST.

@geoff-reid @anhou @iceiilin @keedya @panpan0000 Looking forward to your comments!

### DIFF
--- a/HWIMO-TEST
+++ b/HWIMO-TEST
@@ -48,6 +48,7 @@ restoreDependencies(){
 
 build(){
     rm -rf node_modules
+    ./install-swagger-ui.sh
     npm install
 }
 


PengTian0:
test this please

panpan0000:
+@keedya
in the concourse demo, @keedya mentioned unit-test has been successfully run in workers' container.
So ping @keedya about his solution.
if it happens to be the same as @PengTian0's solution, then this PR can be +1

panpan0000:
👍 
since @keedya  also did the same thing in Concourse alpha

PengTian0:
test this please

PR #748, opened 2017-08-23
 Change redfish-validator-service.js  filename to align with injector name

pengz1:
Background
redfish-validator-service.js uses an injectable name Http.Api.Services.Redfish, this is a little confusing. We should align filename with injectable name

Details:
Change redfish-validator-service.js  filename to redfish-validator-service.js
Change redfish-validator-service-spec.js  filename to redfish-validator-service-spec.js
Modified related variables and description in the two documents


panpan0000:
test this please

PR #753, opened 2017-08-28
RAC-4910 Added systems/{identifier}/Bios.ResetBios

patelb10:
* Added ResetBios Redfish Api Endpoint
* Added spec tests for ResetBios


patelb10:
@larry-dean @jimturnquist @manfrednde @pathakvaidehi2391

PR #764, opened 2017-09-01
remove lookup for macAddress when call /profiles?macs&&ips

lanchongyizu:
Fix https://rackhd.atlassian.net/browse/RAC-5768
The solution is remove lookup for macAddress when call /profiles?macs&&ips.

@RackHD/corecommitters

### DIFF
--- a/lib/services/profiles-api-service.js
+++ b/lib/services/profiles-api-service.js
@@ -23,7 +23,8 @@ di.annotate(profileApiServiceFactory,
         'Profiles',
         'Services.Environment',
         'Http.Services.Swagger',
-        'Constants'
+        'Constants',
+        'Assert'
     )
 );
 function profileApiServiceFactory(
@@ -40,7 +41,8 @@ function profileApiServiceFactory(
     profiles,
     Env,
     swaggerService,
-    Constants
+    Constants,
+    assert
 ) {
 
     var logger = Logger.initialize(profileApiServiceFactory);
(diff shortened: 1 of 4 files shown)


lib/services/profiles-api-service.js
```diff
@@ -75,32 +75,29 @@ function profileApiServiceFactory(
...
-        var index = _.findIndex(ipAddresses, function(ip) {
-            return (ip && (ip === res.locals.ipAddress));
-        });
+            var index = _.findIndex(ipAddresses, function(ip) {
+                return (ip && (ip === ipAddress));
+            });
 
-        if (index < 0 || !macAddresses[index]) {
-            return Promise.resolve();
+            if(index >= 0) {
```
anhou: missing !macAddresses[index] check

lib/services/profiles-api-service.js
```diff
@@ -75,32 +75,29 @@ function profileApiServiceFactory(
         return _.flattenDeep([macs]);
     };
 
-    ProfileApiService.prototype.setLookup = function(req, res) {
-        var query = req.swagger ? req.swagger.query : req.query;
+    ProfileApiService.prototype.getRequestMacAddress = function(query, ipAddress) {
```
anhou: appreciate extracting this function, but it's better to refine the function name and it's parameter, maybe getMacAddressInRequest or other name is more clear. and the parameter ipAddress here is specific from 'res'

lib/services/profiles-api-service.js
```diff
@@ -75,32 +75,29 @@ function profileApiServiceFactory(
...
+        if (query.macs && query.ips) {
```
anhou: it looks like you're changing guard sentences, it reduce 'return' numbers but increase the nesting level, maybe there are debates for this

lib/services/profiles-api-service.js
lanchongyizu: The function is designed to return `undefined/null` when we can't get macAddress.

lib/services/profiles-api-service.js
lanchongyizu: You are right, it's only my preference here, since we only have one nesting level here.

lib/services/profiles-api-service.js
```diff
@@ -375,33 +372,44 @@ function profileApiServiceFactory(
...
-                        .then(function (node) {
-                            return self.getProfileFromTaskOrNode(node, 'compute')
-                                .then(function (render) {
-                                    return(render);
+                return self.getNode(macAddresses, options)
+                    .then(function (node) {
+                        return self.getProfileFromTaskOrNode(node, 'compute')
+                            .then(function (render) {
+                                return _.defaults(render, {
+                                    ignoreLookup: res.locals.macAddress ? true : false
```
anhou: it will leads to macAddress render missing,

lanchongyizu: Here I use res.locals.macAddress to render macAddress, https://github.com/RackHD/on-http/pull/764/files#diff-65e9fcdeffbfaa4a0d0b2de5bd283da6R294

PR #767, opened 2017-09-06
RAC-6084: Remove .0, .1 indexes from system linkages. (ie ManagedBy)

jimturnquist:
No need to iterate over (and return) obm settings when the
/redfish/v1/managers query ignores the indexes anyway.

Redfish ServiceValidator found index issues where the ManagedBy
resources were not valid query urls.


jimturnquist:
Test this please.

PR #773, opened 2017-09-13
waterline validation error when driveIds.length==0

wangzhizheng:
we found there will be an waterline validation error when driveIds.length==0 as below and task will fail
Error (E_VALIDATION) :: 1 attribute is invalid
at WLValidationError.WLError (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/waterline/lib/waterline/error/WLError.js:26:15)

So we need to add something to driveIds to avoid the validation error, empty json {} makes sense.

### DIFF
--- a/data/templates/get_driveid.js
+++ b/data/templates/get_driveid.js
@@ -14,7 +14,7 @@ var options = {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


iceiilin:
ok to test

PR #790, opened 2017-09-26
RAC-6012 - address CentOS os install 360000ms timeouts on physical hardware

hohene:
removed the network option to address issues being seen in baremetal nodes.
The network port is not always coming up in time to allow selection of which interface to start the kickstart on.

Paired with Manfred on this update.  Tested against physical bare-metal stack with Quantas.
@manfrednde 
@keedya


tannoa2:
@hohene just so you know, we changed on-builconfig as part of a task we are working and it will make this PR fail

tannoa2:
test this please

hohene:
test this please

hohene:
@jimturnquist

PR #812, opened 2017-11-10
Update the reponse in redfish API for setSecureBoot

lanchongyizu:
**Backgroud**

The response for `POST /redfish/v1/Systems/{identifier}/SecureBoot` should be
```
{
"@odata.id": "string"
}
```
instead of
```
{
"Message": "Successfully Completed Request"
}
```
**Done**
1) Update the reponse in redfish API for setSecureBoot
2) refine the code

**Reviewers**
@RackHD/corecommitters @nortonluo

### DIFF
--- a/lib/api/redfish-1.0/systems.js
+++ b/lib/api/redfish-1.0/systems.js
@@ -4,7 +4,6 @@
 
 var injector = require('../../../index.js').injector;
 var redfish = injector.get('Http.Api.Services.Redfish');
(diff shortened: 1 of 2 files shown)


anhou:
@lanchongyizu So this will change the interface? is there any impacts?

lanchongyizu:
@anhou This won't change the interface. The swagger document has already defined the response to be 
```
{
"@odata.id": "string"
}
```
So this PR would be a bug fix.

PR #816, opened 2017-11-20
Add retry in swagger and web ui script

mcgG:
**Background**
----
Because of occasionally network fail, curl may fail in install-swagger-ui.sh and install-web-ui.sh
[https://rackhd.atlassian.net/browse/RAC-4894](https://rackhd.atlassian.net/browse/RAC-4894)

**Detail**
----
Add curl retry.

@iceiilin @anhou @pengz1 @bbcyyb @PengTian0

### DIFF
--- a/install-web-ui.sh
+++ b/install-web-ui.sh
@@ -1,4 +1,4 @@
-# Copyright 2017, Dell EMC, Inc.
+# Copyright 2015-2017, Dell EMC, Inc.
 
 cd ./static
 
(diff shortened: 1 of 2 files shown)


install-web-ui.sh
```diff
@@ -1,9 +1,15 @@
-# Copyright 2015, EMC, Inc.
+# Copyright 2017, Dell EMC, Inc.
```
pengz1: Copyright 2015-2017, Dell EMC, Inc

PR #818, opened 2017-11-28
add ucsCallback to north bound

PengTian0:
@anhou @pengz1


pengz1:
You can keep this callback API for both southbound and northbound API.

PR #823, opened 2017-12-13
RAC-6553 Port add & delete volume redfish api

yaolingling:
**Background**
      To port add & delete volume redfish api, the graph need to be replaced with new graphs. Details in https://rackhd.atlassian.net/browse/RAC-6553.

**Reviewers**
@nortonluo @lanchongyizu @AlaricChan

### DIFF
--- a/lib/api/redfish-1.0/systems.js
+++ b/lib/api/redfish-1.0/systems.js
@@ -71,7 +71,7 @@ var dataFactory = function(identifier, dataName) {
             }).then(function(data) {
                 return data[0].sel;
             });
-        default: 
+        default:
             return nodeApi.getNodeCatalogSourceById(identifier, dataName);
     }
 };
@@ -1042,7 +1042,6 @@ var getStorage = controller(function (req, res) {
                                             options.devices.push(ele.SMART);
                                         }
                                     });
-
                                     return redfish.render('redfish.1.0.0.storage.1.0.0.json',
                                         'Storage.v1_1_1.json#/definitions/Storage',
                                         options);
@@ -1157,7 +1156,6 @@ var listVolume = controller(function(req, res)  {
             return Promise.resolve(dataFactory(identifier, 'hardware'))
             .then(function(hardware) {
                 options.hardware = hardware;
-
                 options.virtualDisks = [];
                 _.forEach(hardware.data.storage.virtualDisks, function(ele) {
                     options.virtualDisks.push(ele);
@@ -1841,7 +1839,7 @@ var deleteVolume = controller(function(req,res) {
     var options = redfish.makeOptions(req, res, identifier);
     var volumeIndex = req.swagger.params.volumeIndex.value;
     var payload = req.swagger.params.payload.value;
-    var graphName = 'Graph.Delete.Volume';
+    var graphName = 'Graph.Dell.Wsman.Delete.Volume';
     var graphOptions = {
         defaults: payload
     };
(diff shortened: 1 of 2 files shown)


lib/api/redfish-1.0/systems.js
```diff
@@ -240,9 +240,7 @@ var getSystem = controller(function(req, res) {
         if(result.vendor === 'Dell'){
             return Promise.props({
                 hardware: dataFactory(identifier, 'hardware'),
-                boot: dataFactory(identifier, 'boot'),
                 chassis: dataFactory(identifier, 'chassis'),
-                chassisData: dataFactory(identifier, 'chassisData'),
```
lanchongyizu: Why delete these two sources?

yaolingling: It was deleted carelessly.

lib/api/redfish-1.0/systems.js
```diff
@@ -1918,19 +1916,22 @@ var addVolume = controller(function(req,res) {
...
                     case "Mirrored":
-                        graphOptions.defaults.raidLevel = 'r1';
+                        graphOptions.defaults.raidLevel = 1;
                         break;
                     case "StripedWithParity":
-                        graphOptions.defaults.raidLevel = 'r5';
+                        graphOptions.defaults.raidLevel = 5;
+                        break;
+                    case "RawDevice":
+                        graphOptions.defaults.raidLevel = 6;
```
lanchongyizu: RawDevice means that `The volume is a raw physical device without any RAID or other virtualization applied.`  https://github.com/RackHD/on-http/blob/master/static/DSP8010_2016.3/metadata/Volume_v1.xml#L180-L182
Please remove RawDevice here.

yaolingling: ok

spec/lib/api/redfish-1.0/system-spec.js
```diff
@@ -3084,6 +3084,74 @@ describe('Redfish Systems Root', function () {
...
+                    {
+                        "@odata.id": "/redfish/v1/Systems/SomeNodeId/Storage/RAID.Integrated.1-1/Drives/1"
+                    },
+                    {
+                        "@odata.id": "/redfish/v1/Systems/SomeNodeId/Storage/RAID.Integrated.1-1/Drives/2"
+                    }]
+                }
+            }
+        };
+        waterline.catalogs.findLatestCatalogOfSource.resolves(Promise.resolve({
```
pengz1: Promise.resolve() is not necessary, any value returned from promise is a promise already.

spec/lib/api/redfish-1.0/system-spec.js
```diff
@@ -3084,6 +3084,74 @@ describe('Redfish Systems Root', function () {
...
+                "Name": "name 123",
+                "CapacityBytes": 1234567,
+                "VolumeType": "NonRedundant",
+                "Links": {
+                    "Drives@odata.count": 3,
+                    "Drives": []
+                }
+            }
+        };
+        waterline.catalogs.findLatestCatalogOfSource.resolves(Promise.resolve({
```
pengz1: Promise.resolve() is not necessary, any value returned from promise is a promise already.

pengz1: Generally, it is not a good idea to throw 500 for user input error, 400 is more reasonable.

spec/lib/api/redfish-1.0/system-spec.js
yaolingling: ok

spec/lib/api/redfish-1.0/system-spec.js
yaolingling: ok

PR #831, opened 2018-03-05
update package "moment" due to security vulnerability

keedya:


keedya:
@geoff-reid @dalebremner

PR #832, opened 2018-03-06
 Expose timestamp for getAllCatalogs API 

pengz1:
Expose timestamp for getAllCatalogs API. Otherwise user can't tell difference between catalogs with the same source and nodeId but created at different time.


pengz1:
@iceiilin @lanchongyizu @HQebupt

